PR #4, opened 2021-06-02
Handle wallet account change

sebastinez:
Hello @xla and @cloudhead!

I created this PR to handle the eventual wallet account change by the user.
Please let me know any edits you need or improvements.

Cheers!

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -165,7 +165,7 @@ export const session = derived(state, s => {
 });
 
 // Updates state when user changes accounts
-window.ethereum?.on("accountsChanged", state.setChangedAccount)
+window.ethereum?.on("accountsChanged", state.setChangedAccount);
 
 state.subscribe(s => {
   console.log("session.state", s);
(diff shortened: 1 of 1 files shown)


src/session.ts
```diff
@@ -154,6 +164,9 @@ export const session = derived(state, s => {
   return null;
 });
 
+// Updates state when user changes accounts
+window.ethereum?.on("accountsChanged", state.setChangedAccount)
```
cloudhead, line 168:
```suggestion
window.ethereum?.on("accountsChanged", state.setChangedAccount);
```

cloudhead:
Very nice! Have you confirmed locally that it works as expected?

sebastinez:
> Very nice! Have you confirmed locally that it works as expected?

Yes, I ran the build in Firefox and Chrome with Metamask and did use different scenarios:
- with Metamask disabled
- with Metamask disabled but enabling it with the page open and refreshing, to have the `window.ethereum` provider.
- changing accounts inside Metamask

cloudhead:
Awesome, thanks!

cloudhead (approved):

PR #7, opened 2021-06-03
Extract contract abis to config.json

sebastinez:
Hello @cloudhead,

I was looking into the registration workflow and thought it would be a good idea to check the users RAD balance to be higher than the`registrar.registrationFeeRad` before sending the tx, looking into this I started to extract all the contract ABIs into the `config.json` file.

IMHO the code gets cleaner and it reduces code duplication.
Please let me know what you think.

Cheers

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -26,18 +26,6 @@ const GetOrgs = `
   }
 `;
 
-const orgFactoryAbi = [
-  "function createOrg(address) returns (address)",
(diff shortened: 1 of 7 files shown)


cloudhead:
Hey, yes I like this change. Could you rebase on current master? Thanks!

cloudhead (approved):

sebastinez:
> Hey, yes I like this change. Could you rebase on current master? Thanks!

No problem, there I rebased it onto master

PR #8, opened 2021-06-03
Validate RAD balance before commit tx

sebastinez:
Hi @cloudhead,

Sorry for the consecutive PRs but I wanted to build upon the ABIs refactoring with this PR, and I don't like to propose changes of different kind in the same PR.

Trying to register my name yesterday I spend fee for a tx without having enough RAD tokens, so the idea is to stop the user that he has not enough RAD to register a name before he spends ETH in the tx.

I'm also thinking about writing some kind of util function for the construction of the `ethers.Contract` instance

Cheers

### DIFF
--- a/src/base/registrations/registrar.ts
+++ b/src/base/registrations/registrar.ts
@@ -67,6 +67,10 @@ export function registrar(config: Config) {
   return new ethers.Contract(config.registrar.address, config.abi.registrar, config.provider);
 }
 
+export function radToken(config: Config) {
+  return new ethers.Contract(config.radToken.address, config.abi.token, config.provider);
+}
+
 export async function registrationFee(config: Config) {
   return await registrar(config).registrationFeeRad();
 }
@@ -87,7 +91,7 @@ export async function registerName(name: string, owner: string, config: Config) 
       await commitAndRegister(name, owner, config);
     }
   } catch (e) {
-    throw { type: Failure.TransactionFailed, message: e.message, txHash: e.txHash };
+    throw { type: e.type || Failure.TransactionFailed, message: e.message, txHash: e.txHash };
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/base/registrations/registrar.ts
```diff
@@ -95,6 +99,9 @@ async function commitAndRegister(name: string, owner: string, config: Config) {
   let salt = ethers.utils.randomBytes(32);
   let minAge = (await registrar(config).minCommitmentAge()).toNumber();
   let fee = await registrationFee(config);
+  // Avoids gas spent by the owner, trying to commit to a name and not having
+  // enough RAD balance
+  assertBalance((await radToken(config).balanceOf(owner)).toNumber(), fee);
```
cloudhead, line 104: I think this would be better handled by throwing a custom error that we can catch in the view, see line 94 for eg. Perhaps adding a `Failure.InsufficientBalance` variant to the enum is a good idea.

cloudhead:
> Sorry for the consecutive PRs but I wanted to build upon the ABIs refactoring with this PR, and I don't like to propose changes of different kind in the same PR.

I prefer this too, thanks.

src/base/registrations/registrar.ts
sebastinez: I agree, eventually the `commitAndRegister` function could throw a custom `InsufficientBalance` error and the `registerName` function could re throw this custom error with a custom message, something like `Not enough RAD funds` or in its absence throw a generic `TransactionFailed` error.

sebastinez:
I rebased and forced pushed on the latest master commit 01585eaa39d5b9771ac383a5b4ca2e74dc0457e9

src/base/registrations/registrar.ts
```diff
@@ -87,14 +91,19 @@ export async function registerName(name: string, owner: string, config: Config)
...
   }
 }
 
 async function commitAndRegister(name: string, owner: string, config: Config) {
   let salt = ethers.utils.randomBytes(32);
   let minAge = (await registrar(config).minCommitmentAge()).toNumber();
   let fee = await registrationFee(config);
+  // Avoids gas spent by the owner, trying to commit to a name and not having
+  // enough RAD balance
+  if ((await radToken(config).balanceOf(owner)).toNumber() < fee) {
```
cloudhead, line 104: This can overflow if the balance is not representable as a javascript number. It's best to use the `.lt()` method on `BigNumber` to make the comparison.

sebastinez:
> > I rebased and forced pushed on the latest master commit [01585ea](https://github.com/radicle-dev/radicle-interface/commit/01585eaa39d5b9771ac383a5b4ca2e74dc0457e9)
> 
> Thanks, I had introduced a breaking build, hence the force push sweat_smile

It was a nice git challenge to clean my branches up 😄

cloudhead (approved):

cloudhead:
So if we actually want to preserve your GPG signature, you have to squash + force push, then I can do the merge.

I should probably write a little contribution guide.

sebastinez:
> So if we actually want to preserve your GPG signature, you have to squash + force push, then I can do the merge.

Done

PR #13, opened 2021-06-08
Add session persistence

### LINKED ISSUE #5: Store session state to avoid reconnection after page refresh
Hi @cloudhead,

I noticed that on doing a page refresh the page looses the connection to the web3 provider because of a fresh state and one needs to connect again to Metamask.

If one would like to keep the session open, the session could be stored in `LocalStorage` and on a new visit or page refresh, could be checked if there is already something stored.
Or eventually to avoid storing app state in `LocalStorage` one could store a preference to keep the session open and this could execute Metamask to request the currently available accounts, without needing the interaction of the user to reconnect.

Please let me know what you think, would provide a PR for it.

Cheers

sebastinez:
This PR fixes #5.

- It saves the session storage to `LocalStorage`
- Updates the address when it changes
- And removes the session from `LocalStorage` when the user disconnects or Metamask gets closed.

There are some edge cases that Metamask gets locked by the user and does not emit a `accountsChanged` event..
Eventually there will be some kind of lock / unlock event, in the future by Metamask.

The implementation is quite generic and can also work with WalletConnect without great modifications.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -25,12 +25,15 @@ export type State =
 
 export interface Session {
   address: string
-  tokenBalance: any
+  tokenBalance: BigNumber
   tx: TxState
 }
 
-export const createState = (initial: State) => {
+export const loadState = (initial: State) => {
   const store = writable<State>(initial)
+
+  const session = window.localStorage.getItem("session");
+  if (session) store.set({ connection: Connection.Connected, session: JSON.parse(session) })
 
   return {
     subscribe: store.subscribe,
@@ -52,11 +55,12 @@ export const createState = (initial: State) => {
       const address = await signer.getAddress();
 
       try {
-        const tokenBalance = await token.balanceOf(address);
+        const tokenBalance: BigNumber = await token.balanceOf(address);
         store.set({
           connection: Connection.Connected,
           session: { address, tokenBalance, tx: null }
         });
+        window.localStorage.setItem("session", JSON.stringify({ address, tokenBalance, tx: null }));
       } catch (e) {
         console.error(e);
       }
@@ -77,7 +81,7 @@ export const createState = (initial: State) => {
 
       try {
         const token = new ethers.Contract(config.radToken.address, config.abi.token, config.provider);
-        const tokenBalance = await token.balanceOf(addr);
+        const tokenBalance: BigNumber = await token.balanceOf(addr);
 
         state.session.tokenBalance = tokenBalance;
         store.set(state);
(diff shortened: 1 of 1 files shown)


src/session.ts
```diff
@@ -57,6 +60,7 @@ export const createState = (initial: State) => {
           connection: Connection.Connected,
           session: { address, tokenBalance, tx: null }
         });
+        window.localStorage.setItem("session", JSON.stringify({ address, tokenBalance }));
```
cloudhead, line 63: Might be worth setting `tx: null` here, otherwise it'll be `undefined` when read back out, which is not well typed.

src/session.ts
```diff
@@ -32,6 +32,9 @@ export interface Session {
 export const createState = (initial: State) => {
```
cloudhead, line 32: Perhaps this should now be renamed to `loadState`, the param now acts as a default.

src/session.ts
```diff
@@ -146,7 +150,13 @@ export const createState = (initial: State) => {
       store.update(s => {
         switch (s.connection) {
           case Connection.Connected:
-            s.session.address = address; 
+            // In case of locking Metamask the accountsChanged event returns undefined.
+            // To prevent out of sync state, the wallet gets disconnected.
+            if(address === undefined) disconnectWallet();
```
cloudhead, line 155: Little style thing: we usually have whitespace between keywords and the `()`, eg. `if (...)`, `function (...)`, etc., same on line 36.

src/session.ts
```diff
@@ -146,7 +150,13 @@ export const createState = (initial: State) => {
...
+            else {
+              s.session.address = address; 
+              window.localStorage.setItem("session", JSON.stringify({ address, tokenBalance: s.session.tokenBalance }))
```
cloudhead, line 158: Should we also save the `tx` state?

src/session.ts
```diff
@@ -25,13 +25,16 @@ export type State =
 
 export interface Session {
   address: string
-  tokenBalance: any
+  tokenBalance: BigNumber
```
sebastinez, line 28: I added the `BigNumber` type for `tokenBalance`, since `token.balanceOf` doesn't define its return type.
Here on line 58 and 84

src/session.ts
sebastinez: For persisting the session it has no direct value, but to comply with the interface type we should.
Eventually in the future, it'll be of use.

cloudhead (approved):

PR #16, opened 2021-06-09
Implement code formatter

### LINKED ISSUE #6: Implement Linter
Hi @cloudhead,

Wanted to know if you thought about implementing some kind of linter like ESLint to enforce code standards and style?
If you agree I could propose some standard rules that apply to the current code and enforce it for future contributions.

Cheers

sebastinez:
This PR closes #6 

I used `prettier` and `prettier-plugin-svelte` to implement a code formatter.
The configuration is pretty simple and originally from `radicle-upstream` with some minor changes for this repo.

To be able to use GitHub CI I created a very bare GitHub Action, and the core is in the shell script `scripts/format`

Let us discuss here which format options are the correct ones for this repo
https://prettier.io/docs/en/options.html

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+node_modules
--- a/scripts/check-format
+++ b/scripts/check-format
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+# Checks local /node_modules,
+# if not found will download that package and execute it.
+# Does not modify files, informs bad code format
+npx eslint .
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,32 @@
+module.exports = {
+  env: {
+    browser: true,
+    node: true,
+  },
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended"
+  ],
+  parser: "@typescript-eslint/parser",
+  parserOptions: {
+    ecmaVersion: 12,
+    sourceType: "module",
+  },
+  plugins: [
+    "svelte3",
+    "@typescript-eslint"
+  ],
+  overrides: [
+    {
+      files: ["*.svelte"],
+      processor: "svelte3/svelte3",
+    }
+  ],
+  rules: {
+    "@typescript-eslint/no-explicit-any": "off",
+    indent: ["error", 2, { "SwitchCase": 1 }]
+  },
+  settings: {
+    "svelte3/typescript": true, 
+  },
+};
--- a/.github/workflows/check-format.yml
+++ b/.github/workflows/check-format.yml
@@ -1,0 +1,12 @@
+name: check-format
+on: [push]
+
+jobs:
+  check-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - run: npm ci
+      - run: ./scripts/check-format
+        shell: bash


cloudhead:
I've gone through the formatted code, and I think it does a good job on the `.ts` files, but not the HTML in the .svelte files, as it wraps or unwraps lines seemingly randomly, hurting readability.

So as a first step I'd propose these settings:

```diff
diff --git a/prettier.config.js b/prettier.config.js
index e08b9ff..bee6d04 100644
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -2,10 +2,10 @@ module.exports = {
   "arrowParens": "avoid",
   "tabWidth": 2,
   "semi": true,
-  "singleQuote": true,
+  "singleQuote": false,
   "quoteProps": "preserve",
-  "trailingComma": "es5",
+  "trailingComma": "all",
```

and a `.prettierignore` with

```
*.html
*.svelte
*.css
```

sebastinez:
@cloudhead force pushed the latest changes like we talked about

scripts/check-format
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# Checks local /node_modules,
+# if not found will download that package and execute it.
+# Does not modify files, informs bad code format
+npx eslint .
```
cloudhead, line 6: Until we've linted the codebase, we should keep this commented out.

.eslintignore
```diff
@@ -0,0 +1 @@
+node_modules
```
cloudhead, line 1: Missing a new-line here and in the other file too

scripts/check-format
sebastinez: Eventually we could keep it like this and disable the CI in the .github/workflows folder, so it doesn't fail on every push

cloudhead: Yeah that's fine too.

sebastinez: [We can disable the workflow from the GitHub UI until everything is linted.](https://docs.github.com/es/actions/managing-workflow-runs/disabling-and-enabling-a-workflow#disabling-and-enabling-workflows-with-the-github-ui)

What do you think?

cloudhead:
Hmm weird, build is failing for platform reasons :thinking:

sebastinez:
> Hmm weird, build is failing for platform reasons thinking

It seems like an issue to how the fsevents package got compiled, it wants macOS darwin instead of linux, 
```sh
5:10:17 PM: npm ERR! notsup Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
```
Eventually I could remove the package-lock.json file I commited, and see if this helps

sebastinez:
I think to finish this PR you should commit a new package.lock.json based on the new packages used.
Since I can't create one on macOS.

cloudhead:
> @cloudhead I think to finish this PR you should commit a new package.lock.json based on the new packages used.
> Since I can't create one on macOS.

You mean on `master`?

sebastinez:
Yes this should do the trick. Or I could remove the packages from packages.json and you could install them

cloudhead:
Ok done, I think. Or does this require a completely fresh lock file? Trying to avoid updating all dependencies though.

sebastinez:
I'll rebase this branch without updating the lock file but including the new dependencies, and then you could do a `npm install` to update the lock file

cloudhead:
> I'll rebase this branch without updating the lock file but including the new dependencies, and then you could do a `npm install` to update the lock file

This should be done already - I added the dependencies to master.

sebastinez:
Okay great! Then I'll remove both package.json and lock file from this PR

cloudhead (approved):

Issue #17, opened 2021-06-09
User Identities


sebastinez:
## Objective
User and Orgs Identities require the display of their identities if registered. Using ENS every change to ones metadata requires a transaction to the corresponding contract.
We will work with the IDX protocol from ceramic to implement a way to more efficiently update the information.

### Lookup Priority
1. ENS Registrar
2. DID IDX on Ceramic
3. Ethereum Blockies

### Home
- [x] Search input should be able to also search for user identities and ETH addresses. (3114db12c1a7687ce722b5823da04815b36dcb82)
### User profile
- [x] Replace user addresses with resolved ENS names (if registered) (a9eebfa033bc4d53e6692e86e8caf46f30e8b5cf)
- [x] Tooltip with eth address for `Address` component when not showing address. (5c63c1da1b0d376b68af1b3830b6923fc866b1d7)
- [x] Instead of the Ethereum blockies a url avatar should be loaded if available (5c63c1da1b0d376b68af1b3830b6923fc866b1d7)
- [x] /users/:address page should show avatar, name, websites if available as shortcuts (b74943609d78a5b21fef02993e25b38d03b9b5ed)
- [x] EOA addresses should link to /users/:address page (5c63c1da1b0d376b68af1b3830b6923fc866b1d7)
- [x] The owner in the registration page should be resolved. (805ac0ddbbe9bc0e8476dd1b6af972ea8a7c4290)
- [x] Orgs listing should show avatars of orgs (552755a3b9e1bfd1e1d3ecb8d893a351a2423914)
- [x] Member list of orgs should resolve to available metadata (cb493801dcad5c6ec92dbd25c0ad4e2fef5eaccf)
- [x] Org details should show resolved owner and participants avatar and registered name (cb493801dcad5c6ec92dbd25c0ad4e2fef5eaccf)
- [x] Connect/Disconnect button in header should show avatar (or blockies) and ENS name if available (805ac0ddbbe9bc0e8476dd1b6af972ea8a7c4290)

### Improvements
- [x] Allow linking of CAIP10 Accounts to DIDs (dcdec8c8e1ba98b5a6473d21f8698c1a411fb4de)

## PR Implementations
1. Creation of user route with UI that shows user address and metadata. (#20)
2. Implementation of the IDX protocol #28 
3. Usage of resolved addresses and avatars in other parts of the app #44

cloudhead:
Thoughts on where ENS can be integrated:

[...]

PR #20, opened 2021-06-12
Add User View Route

sebastinez:
This PR works on #17

- It implements a new User route `/users/:address` which shows a user profile, with avatar or blockies as a fallback and the profile links.
- Adds a loading flag to show only the final state of the profile
- If one access the route with a contract address it also shows the stored information on ENS.
- Removes the Blockies from the Address component, so it can be decided by each page if it wants to show.

To test the implementation you could use the address `0xef1B2D659F4662f6a1e4a1B4791A1B4cEFF8FbA1` for sebastinez.rad.eth

### DIFF
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -1,12 +1,10 @@
 <script lang="typescript">
   import { onMount } from 'svelte';
-  import { navigate } from 'svelte-routing';
   import { link } from 'svelte-routing';
   import { ethers } from 'ethers';
   import { safeLink, identifyAddress, formatAddress, AddressType } from '@app/utils';
   import Loading from '@app/Loading.svelte';
   import Avatar from "@app/Avatar.svelte";
-  import Error from '@app/Error.svelte';
   import type { Config } from '@app/config';
   import type { Registration } from '@app/base/registrations/registrar';
   import { getRegistration } from '@app/base/registrations/registrar';
@@ -23,13 +21,13 @@
     : ethers.utils.getAddress(address);
   let addressType: AddressType | null = null;
   let addressName: string | null = null;
-  let getInfo: Promise<Registration | null>;
+  let info: Registration | null;
 
   onMount(async () => {
     identifyAddress(address, config).then((t: AddressType) => addressType = t);
     if (resolve) {
       addressName = await config.provider.lookupAddress(address);
-      getInfo = getRegistration(addressName, config);
+      info = await getRegistration(addressName, config);
     }
   });
   $: addressLabel = addressName ?? checksumAddress;
@@ -56,28 +54,22 @@
   }
 </style>
 
-{#await getInfo}
-  <Loading fadeIn />
-{:then info}
-  <div class="address" title={address} class:no-badge={noBadge}>
-    {#if !noAvatar}
-      <Avatar inline source={info?.avatar ?? address} />
-    {/if}
-    {#if addressType === AddressType.Org}
-      <a use:link href={`/orgs/${address}`}>{addressLabel}</a>
-      <span class="badge">org</span>
-    {:else if addressType === AddressType.Safe}
-      <a href={safeLink(address, config)} target="_blank">{addressLabel}</a>
-      <span class="badge safe">safe</span>
-    {:else if addressType === AddressType.Contract}
-      <a href={`/orgs/${address}`} target="_blank">{addressLabel}</a>
-      <span class="badge">contract</span>
-    {:else if addressType === AddressType.EOA}
-      <a href={`/users/${address}`} target="_blank">{addressLabel}</a>
-    {:else if !noBadge}
-      <div class="loading"><Loading small /></div>
-    {/if}
-  </div>
-{:catch error}
-  <Error {error} title="Address could not be loaded" on:close={() => navigate('/')} />
-{/await}
+<div class="address" title={address} class:no-badge={noBadge}>
+  {#if !noAvatar}
+    <Avatar inline source={info?.avatar ?? address} />
+  {/if}
+  {#if addressType === AddressType.Org}
+    <a use:link href={`/orgs/${address}`}>{addressLabel}</a>
+    <span class="badge">org</span>
+  {:else if addressType === AddressType.Safe}
+    <a href={safeLink(address, config)} target="_blank">{addressLabel}</a>
+    <span class="badge safe">safe</span>
+  {:else if addressType === AddressType.Contract}
+    <a href={`/orgs/${address}`} target="_blank">{addressLabel}</a>
+    <span class="badge">contract</span>
+  {:else if addressType === AddressType.EOA}
+    <a href={`/users/${address}`} target="_blank">{addressLabel}</a>
+  {:else if !noBadge}
+    <div class="loading"><Loading small /></div>
+  {/if}
+</div>
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -59,26 +59,25 @@
 {#await getInfo}
   <Loading fadeIn />
 {:then info}
-<div class="address" title={address} class:no-badge={noBadge}>
-  {#if !noAvatar}
+  <div class="address" title={address} class:no-badge={noBadge}>
+    {#if !noAvatar}
       <Avatar inline source={info?.avatar ?? address} />
-  {/if}
-  {#if addressType === AddressType.Org}
-    <a use:link href={`/orgs/${address}`}>{addressLabel}</a>
-    <span class="badge">org</span>
-  {:else if addressType === AddressType.Safe}
-    <a href={safeLink(address, config)} target="_blank">{addressLabel}</a>
-    <span class="badge safe">safe</span>
-  {:else}
-    <a href={`/users/${address}`} target="_blank">{addressLabel}</a>
-    {#if addressType === AddressType.Contract}
+    {/if}
+    {#if addressType === AddressType.Org}
+      <a use:link href={`/orgs/${address}`}>{addressLabel}</a>
+      <span class="badge">org</span>
+    {:else if addressType === AddressType.Safe}
+      <a href={safeLink(address, config)} target="_blank">{addressLabel}</a>
+      <span class="badge safe">safe</span>
+    {:else if addressType === AddressType.Contract}
+      <a href={`/orgs/${address}`} target="_blank">{addressLabel}</a>
       <span class="badge">contract</span>
     {:else if addressType === AddressType.EOA}
-      <!-- Don't show anything for EOAs -->
+      <a href={`/users/${address}`} target="_blank">{addressLabel}</a>
     {:else if !noBadge}
       <div class="loading"><Loading small /></div>
     {/if}
-</div>
+  </div>
 {:catch error}
   <Error {error} title="Address could not be loaded" on:close={() => navigate('/')} />
 {/await}
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -6,6 +6,7 @@
   import Loading from '@app/Loading.svelte';
   import type { Config } from '@app/config';
   import type { Registration } from '@app/base/registrations/registrar';
+  import Avatar from "@app/Avatar.svelte";
   import { getRegistration } from '@app/base/registrations/registrar';
   import { identifyAddress, formatAddress, AddressType } from '@app/utils';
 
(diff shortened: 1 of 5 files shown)


cloudhead:
> Removes the Blockies from the Address component, so it can be decided by each page if it wants to show.

I don't think this is the right way -- we are using this component in a bunch of places which rely on it showing the avatar. This PR removes the avatar from all those instances.

I'm curious where we need addresses without the avatar?

sebastinez:
> I'm curious where we need addresses without the avatar?

The issue where I see that a avatar in the Address component could be a problem, is if we replace the label part of the org and user profile with the address component we would receive the avatar twice.. Since we also want to show shortcuts to github, twitter, and url there.

I see two ways here, keep the address component for the registration view for example with the avatar and in the headers of orgs and users just show the address or label part, not using the address component there.

Or eventually we could do a disableAvatar flag in the address component for the latter?

cloudhead:
> I see two ways here, keep the address component for the registration view, showing the avatar, and in the headers of orgs and users pages just show the address or label part, not using the `Address` component there.

I think this is how we currently do it.. which is one solution.

Otherwise yeah there's already a `noBadge` property, you could add a `noAvatar` property? Though we do show an avatar on the orgs view, it's just that the layout makes it hard to have a single component rendering that.

Perhaps it could also be helpful to create a `Avatar` component, that takes an optional image url or address and falls back on blockies.

sebastinez:
> Otherwise yeah there's already a `noBadge` property, you could add a `noAvatar` property? Though we do show an avatar on the orgs view, it's just that the layout makes it hard to have a single component rendering that.
> 
> 
> 
> Perhaps it could also be helpful to create a `Avatar` component, that takes an optional image url or address and falls back on blockies.

Yeah I think that would be the best way.
I'll do the `noAvatar` property and a new `Avatar` component for the places where we have the `Address` component with `noAvatar` and want to show eventually the avatar in a different way.

sebastinez:
@cloudhead with the new Avatar component I implemented the Blockies component logic in the former, to avoid issues with CSS classes, and reduce unnecessary components growth.
Also I'll eventually reduce `getRegistration` calls, since we are calling it for every address and it takes some seconds to complete.

src/Address.svelte
```diff
@@ -2,29 +2,36 @@
...
-  onMount(() => {
-    identifyAddress(address, config).then(typ => addressType = typ);
+  onMount(async () => {
     if (resolve) {
-      config.provider.lookupAddress(address).then(name => addressName = name);
+      addressName = await config.provider.lookupAddress(address);
+      if (addressName) registration = await getRegistration(addressName, config);
     }
+    addressType = await identifyAddress(address, config);
+    loading = false;
```
cloudhead, line 34: A cleaner pattern is to assign the contents of `onMount` to a promise, eg.

```
let getInfo: Promise<AddressType> = identifyAddress(...);
```

and then in the HTML, you do

```svelte
{#await getInfo}
<!-- Loading -->
{:then info}
<!-- Loaded! -->
{:catch err}
<!-- Err -->
{/await}
```

cloudhead: One other thing to consider is that the address type doesn't depend on the name or registration, so it shouldn't have to wait for the others to return, it can be loaded in parallel, for eg. by using the `.then(...)` syntax.

src/Avatar.svelte
```diff
@@ -0,0 +1,50 @@
...
+    height: 1rem;
+    margin-right: 0.5rem;
+  }
+</style>
+
+{#if isAddress(source)}
+  <div class="avatar" class:icon bind:this={container} class:glowOnHover title={source}/>
+{:else}
+  <img class="avatar" class:icon src={source} class:glowOnHover alt="avatar"/>
+{/if}
```
cloudhead, line 50: You should get your editor to add new-lines at the end of files, I'm not sure why it doesn't do that by default (almost all editors do). Files without a newline at the end cause problems with unix tooling, as the last line is not properly terminated.

src/Address.svelte
```diff
@@ -58,16 +58,21 @@
...
+  {#if !noAvatar}
+    <Avatar source={registration?.avatar ?? address} />
+  {/if}
   {#if addressType === AddressType.Org}
     <a use:link href={`/orgs/${address}`}>{addressLabel}</a>
     <span class="badge">org</span>
   {:else if addressType === AddressType.Safe}
     <a href={safeLink(address, config)} target="_blank">{addressLabel}</a>
     <span class="badge safe">safe</span>
   {:else}
```
cloudhead, line 74: I'd rather explicitly check for `EOA` here in case we add another variant later.

src/Avatar.svelte
```diff
@@ -0,0 +1,50 @@
...
+  import { isAddress } from './utils';
+  
+  export let source: string;
+  export let icon: boolean = true;
+  export let glowOnHover = false;
+
+  let container: HTMLElement;
+
+  onMount(() => {
+    if(isAddress(source)) {
```
cloudhead, line 13:
```suggestion
    if (isAddress(source)) {
```

src/Avatar.svelte
```diff
@@ -0,0 +1,50 @@
+<script lang="typescript">
+  import { onMount } from 'svelte';
+  import { createIcon } from '@app/blockies';
+  import { isAddress } from './utils';
+  
+  export let source: string;
+  export let icon: boolean = true;
```
cloudhead, line 7: What does this do?

src/Avatar.svelte
```diff
@@ -0,0 +1,50 @@
+<script lang="typescript">
+  import { onMount } from 'svelte';
+  import { createIcon } from '@app/blockies';
+  import { isAddress } from './utils';
```
cloudhead, line 4: `@app/utils` -- breaks less often if we move this file.

src/Address.svelte
```diff
@@ -2,29 +2,36 @@
...
   import { link } from 'svelte-routing';
   import { ethers } from 'ethers';
-  import { explorerLink, safeLink } from '@app/utils';
-  import Blockies from '@app/Blockies.svelte';
+  import { safeLink } from '@app/utils';
   import Loading from '@app/Loading.svelte';
   import type { Config } from '@app/config';
+  import type { Registration } from '@app/base/registrations/registrar';
+  import Avatar from "@app/Avatar.svelte";
+  import { getRegistration } from '@app/base/registrations/registrar';
```
cloudhead, line 10: Maybe group the imports that come from the same file?

src/Avatar.svelte
sebastinez: I didn't want my editor to enforce styling since it didn't comply with the format of the code base.
With ESLint I'm sure this will be a thing of the past.

src/Avatar.svelte
sebastinez: To be able to define which style of avatar the component should show I introduced this icon property that reduces the avatar to small icon size by default.
If one doesn't want a icon size avatar one should pass `icon={false}`.

Eventually there is a way to make this happen with CSS without hardcoding it, else we could also do a `style` prop with some options like `icon`, `header`, etc`.

src/Address.svelte
sebastinez: > A cleaner pattern than the `loading: boolean` one is to assign the contents of `onMount` to a promise, eg.

Yes I agree I started using it in the users/View page, will apply it also here. 👍

src/Avatar.svelte
sebastinez: For example if using an avatar as inline icon we want to have `margin-right: 0.5rem`, but as part of the users or orgs header it shouldn't have margins.

And the blockies div, since they are not images but divs with a background image they behave very differently to images, regarding trying to fill their width and height as part of an inline element..
If we give them as inline icons a width and height of 1rem they behave as expected..
When they form part of a fixed size div it's not needed to define their size since they adapt to the parent div.

src/Avatar.svelte
cloudhead: I've added this lint to eslint, but it's not really a styling thing - it causes problems when combining files for eg.

src/Avatar.svelte
cloudhead: Added this to eslint also!

src/Avatar.svelte
cloudhead: Hmmm can we call this `inline` then and have it default to false? I think it's nicer if components have params that default to `false`, as the default behavior can change with just `myParam` instead of `myParam={false}`.

sebastinez: Yes that should be no problem 👍

src/Address.svelte
sebastinez: The issue is that we cannot group a type import with a function import.

cloudhead: Oh I just meant having the import lines next to each other :grimacing:

src/Avatar.svelte
sebastinez: It seems that eslint is not checking this in .svelte files.. I'll check all those files manually

src/Address.svelte
sebastinez: To avoid misunderstandings, you say to heck for `EOA` before doing the label check?
In this case we should check after or before `EOA` for `Contract` also

sebastinez: In the case of a contract, we should potentially create a link to an organisation I understand, or could there be eg. multisig users?

sebastinez: 3dda56ee20ec6874230e28215b18dca7a3860a05 fixes this, I'm not quite sure about the final `noBadge` else if, if you have a suggestion please let me know

src/Avatar.svelte
sebastinez: ea01cf541e06988ff6a41785e9076616531958d0 fixes this

src/Address.svelte
sebastinez: 3dda56ee20ec6874230e28215b18dca7a3860a05 fixes this

src/Address.svelte
```diff
@@ -1,51 +1,49 @@
...
+  let getInfo: Promise<Registration | null>;
 
-  onMount(() => {
-    identifyAddress(address, config).then(typ => addressType = typ);
+  onMount(async () => {
+    identifyAddress(address, config).then((t: AddressType) => addressType = t);
     if (resolve) {
-      config.provider.lookupAddress(address).then(name => addressName = name);
+      addressName = await config.provider.lookupAddress(address);
+      getInfo = getRegistration(addressName, config);
```
cloudhead, line 32: Note that `getInfo` is only set here if `resolve == true`, but in the HTML below, we don't render anything unless getInfo resolves!

cloudhead: I think it's preferrable to leave that logic the way it currently is, where the address rendering doesn't depend on anything async from resolving.

src/base/users/View.svelte
```diff
@@ -0,0 +1,92 @@
...
+    align-items: center;
+    justify-content: left;
+  }
+  .url {
+    display: flex; /* Ensures correct vertical positioning of icons */
+    margin-right: 1rem;
+  }
+</style>
+
+{#await getInfo}
```
cloudhead, line 62: Here too I would try to move `getInfo` down, so that we can progressively render the page, otherwise if getInfo fails, nothing renders at all.

cloudhead: I think you can have multiple `await` blocks on the same promise, though I've never tried.

sebastinez: > I think you can have multiple `await` blocks on the same promise, though I've never tried.

I'll give that a try since I had to compromise on what I wanted to be awaited on..

src/Address.svelte
sebastinez: I was looking for someway to show the page only if finished loading so we don't see rapid changes of information.. Will see if I come up with a better way

cloudhead: Yeah I guess if it falls back to the address in case of error, that would be fine. Otherwise I'd say don't worry about it for now.

src/base/users/View.svelte
sebastinez: I removed the waiting on a result, there is some flickering while the information loads, but it's acceptable.

src/Address.svelte
sebastinez: I removed the waiting on a result, there is some flickering while the information loads, but it's acceptable.

cloudhead (approved):
Looks good - maybe rebase on master and then we can merge.

sebastinez:
> Looks good - maybe rebase on master and then we can merge.

Done

PR #21, opened 2021-06-12
Fix address comparison in ens/SetName

sebastinez:
Hi @cloudhead 

Found this little bug that prevented one to set a name to a orgs, we were comparing checksummed address against lowercase addresses.
I used the `isAdressEqual` util function to fix it.


sebastinez:
This is a way to solve with. The other one would be to checksum `org.address`

cloudhead (approved):

PR #23, opened 2021-06-13
Refresh balance after a network change

sebastinez:
This PR resolves the issue with the balance that does not update itself when a network change happens.
Eventually there is a cleaner way to obtain the `config` needed by `state.refreshBalance`

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -2,7 +2,7 @@ import { get, writable, derived, Readable } from "svelte/store";
 import { ethers } from "ethers";
 import type { BigNumber } from 'ethers';
 import type { TransactionReceipt, TransactionResponse } from '@ethersproject/providers';
-import type { Config } from "@app/config";
+import { Config, getConfig } from "@app/config";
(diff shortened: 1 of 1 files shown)


src/session.ts
```diff
@@ -188,7 +188,11 @@ export const session = derived(state, s => {
 window.ethereum?.on('chainChanged', () => location.reload());
 
 // Updates state when user changes accounts
-window.ethereum?.on("accountsChanged", state.setChangedAccount);
+window.ethereum?.on("accountsChanged", async (address: string[]) => {
+  const config = await getConfig()
```
cloudhead, line 192:
```suggestion
  const config = await getConfig();
```

cloudhead (approved):
Yeah I think this is a good enough solution for now.

cloudhead:
Might need a rebase, since it includes some unrelated changes.

src/session.ts
sebastinez: Done

Issue #25, opened 2021-06-16
Change Snowpack in favor of Vite


sebastinez:
We are having troubles with bundling packages like ceramic-js, will implement Vite to see if this brings a solution.

cloudhead:
Closed by #26

PR #26, opened 2021-06-17
Replace Snowpack with Vite

sebastinez:
This PR completes #25.

It replaces Snowpack with Vite, which lets us use IDX and ceramic-js for some reason better than Snowpack.

For it to work I had to:
- Replace all the lang attributes in the svelte script tags from "typescript" to "ts".
- Create new config files with the necessary configurations to work.

Since I can't compile package-lock.json without getting compile errors in the CI, I commit only package.json.

I tried all the eventual functionality and it is very similar to Snowpack in speed and configuration, to be sure I also tried different operations with IDX and it works like a charm.


sebastinez:
It seems that there is a problem with `vite-plugin-svelte` on deployment.
I created this issue https://github.com/sveltejs/vite-plugin-svelte/issues/61#issuecomment-862883818
If I hear anything or get a solution I'll post it here

svelte.config.js
```diff
@@ -1,9 +1,7 @@
...
 
-module.exports = {
-  preprocess: autoPreprocess({
-    defaults: {
-      script: 'typescript',
-    },
-  }),
+export default {
+  // Consult https://github.com/sveltejs/svelte-preprocess
+  // for more information about preprocessors
```
cloudhead, line 5: Let's remove this comment.

snowpack.config.js
```diff
@@ -1,38 +0,0 @@
...
-  },
-  packageOptions: {
-    polyfillNode: true,
-  },
-  devOptions: {
-    open: "none",
-    output: "dashboard"
-  },
-  buildOptions: {
-    sourcemap: true,
```
cloudhead, line 36: Make sure we have sourcemaps as well

snowpack.config.js
```diff
@@ -1,38 +0,0 @@
-/** @type {import("snowpack").SnowpackUserConfig } */
-module.exports = {
-  env: {
-    RADICLE_ALCHEMY_API_KEY: process.env.RADICLE_ALCHEMY_API_KEY,
-    RADICLE_HTTP_API: process.env.RADICLE_HTTP_API,
```
cloudhead, line 5: This works out of the box with vite?

snowpack.config.js
```diff
@@ -1,38 +0,0 @@
...
-  },
-  plugins: [
-    '@snowpack/plugin-svelte',
-    '@snowpack/plugin-typescript',
-  ],
-  routes: [
-    /* Enable an SPA Fallback in development: */
-    {"match": "routes", "src": ".*", "dest": "/index.html"},
-    {"match": "all", "src": "/projects/.*", "dest": "/index.html"},
-    {"match": "all", "src": "/orgs/.*/projects/.*", "dest": "/index.html"},
```
cloudhead, line 22: This works out of the box? Try a path with a filename in it, eg. `/foobar/file.png` or `/orgs/asd/bar.html` and make sure it routes to the app (should just show a blank screen I think)

.github/workflows/check-build.yml
```diff
@@ -0,0 +1,11 @@
+name: check-build
+on: [push]
```
cloudhead, line 2: Add `pull_request` here also.

.github/workflows/svelte-check.yml
```diff
@@ -0,0 +1,11 @@
+name: svelte-check 
+on: [push]
```
cloudhead, line 2: Add `pull_request` here also.

snowpack.config.js
sebastinez: Yeah I think I have to add this to the env variables in vite.config

sebastinez: Vite works with dotenv loading all the env vars that are located in a .env file in the root directory.
https://vitejs.dev/guide/env-and-mode.html#env-files

snowpack.config.js
sebastinez: Yep works out of the box

sebastinez:
The CI errors come from the missing updated package-lock.json.. I did an eslint linting and no errors where thrown

cloudhead (approved):
:+1:

sebastinez:
It seems like the `check-build` CI we could have pass installing as a devDependency `@types/node`

sebastinez:
I uploaded the `package-lock.json` with the `@types/node` package installed and it seems that this time there were no errors. So I'll leave it this way.
Rebased it on master, keen on going back to idx and less devops 😄

src/config.json
```diff
@@ -62,6 +62,7 @@
     },
     "tokens": []
   },
+  "RADICLE_HTTP_API":"",
```
cloudhead, line 65: Let's call this `radicle: { api: string }`, so it matches the format of the safe config.

sebastinez: Okay yeah no problem, I kept it that way to maintain the `process.env` style for a eventual return to injected env vars

cloudhead: Ah I see, we can just do `config.radicle.api = import.meta.env.RADICLE_HTTP_API` when that happens

cloudhead (approved):

PR #28, opened 2021-06-19
Add IDX functionality

sebastinez:
This PR works on #17 

It implements the [idx protocol by Ceramic](https://developers.idx.xyz/learn/overview/)
To use the functionality in all the different components it adds new properties to and `config` and some `utils` functions that allow read from the idx network.

Since it adds a new protocol and is a bigger feature, I would like to discuss this PR in depth with you @cloudhead to see were you see improvements or need changes.
I'm already working on the usage of these functions in all the components and pages in particular the `getLinkedDid,` and `lookupAddress` but feel it's important to lay down the base first.

## Docs
To create a did to be used use the idx CLI, we use the 3boxlabs endpoint which can be configured with `idx config:set ceramic-url https://ceramic-clay.3boxlabs.com`
https://developers.idx.xyz/build/quick-start/#step-3-create-a-did

To write a record on idx with the newly created did one should use the `radicleDefinition` which is `kjzl6cwe1jw14b4i43fpebvxonfyu8ei1f1a3mg32xhlmeugddnsa943dvgkzl` use it with `idx index:set <did:key:myDID>  <radicleDefinitionHash> '{"name":"YourName", "avatar":"avatarUrl","url":"url","github":"GitHubUrl","twitter":"TwitterUrl"}'`

I'm creating a page to enable the user link their address with a did but for now, one has to do it on its own
https://developers.ceramic.network/streamtypes/caip-10-link/api/#set-did-to-caip10link

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,8 @@
 import { ethers } from "ethers";
 import type { TypedDataSigner } from '@ethersproject/abstract-signer';
 import SafeServiceClient from "@gnosis.pm/safe-service-client";
+import CeramicClient from "@ceramicnetwork/http-client";
+import { IDX } from "@ceramicstudio/idx";
 import config from "@app/config.json";
 
 declare global {
@@ -27,6 +29,8 @@ export class Config {
   };
   abi: { [contract: string]: string[] };
   seed: { api?: string };
+  idx: { client: IDX };
+  ceramic: { client: CeramicClient };
   tokens: string[];
   token: ethers.Contract;
 
@@ -42,6 +46,9 @@ export class Config {
       throw `Network ${network.name} is not supported`;
     }
 
+    const ceramic = new CeramicClient(config.ceramic.api);
+    const idx = new IDX({ ceramic });
+
     this.network = network;
     this.seed = { api };
     this.registrar = cfg.registrar;
@@ -56,6 +63,8 @@ export class Config {
     this.signer = signer;
     this.gasLimits = gasLimits;
     this.abi = config.abi;
+    this.idx = { client: idx };
+    this.ceramic = { client: ceramic };
     this.tokens = cfg.tokens;
     this.token = new ethers.Contract(
       this.radToken.address,
--- a/src/session.ts
+++ b/src/session.ts
@@ -36,7 +36,7 @@ export interface Store extends Readable<State> {
   setTxSigning(): void;
   setTxPending(tx: TransactionResponse): void;
   setTxConfirmed(tx: TransactionReceipt): void;
-  setChangedAccount([address]: string[]): void;
+  setChangedAccount(address: string): void;
 }
 
 export const loadState = (initial: State): Store => {
@@ -154,7 +154,8 @@ export const loadState = (initial: State): Store => {
         }
       });
     },
-    setChangedAccount: ([address]: string[]) => {
+
+    setChangedAccount: (address: string) => {
       store.update(s => {
         switch (s.connection) {
           case Connection.Connected:
@@ -164,7 +165,7 @@ export const loadState = (initial: State): Store => {
               disconnectWallet();
             } else {
               s.session.address = address;
-              window.localStorage.setItem("session", JSON.stringify({ address, tokenBalance: s.session.tokenBalance, tx: s.session.tx }));
+              window.localStorage.setItem("session", JSON.stringify({ ...s.session }));
             }
             return s;
           default:
@@ -183,10 +184,15 @@ export const session = derived(state, s => {
   return null;
 });
 
-window.ethereum?.on('chainChanged', () => location.reload());
+window.ethereum?.on('chainChanged', () => {
+  // We disconnect the wallet to avoid out of sync state
+  // between the account address and IDX DIDs
+  disconnectWallet();
+  location.reload();
+});
 
 // Updates state when user changes accounts
-window.ethereum?.on("accountsChanged", async (address: string[]) => {
+window.ethereum?.on("accountsChanged", async ([address]: string) => {
   const config = await getConfig();
   state.setChangedAccount(address);
   state.refreshBalance(config);
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,6 +5,14 @@ import multihashes from 'multihashes';
 import EthersSafe from "@gnosis.pm/safe-core-sdk";
 import type { Config } from '@app/config';
 import { assert } from '@app/error';
+import type { Registration } from "@app/base/registrations/registrar";
+import { getRegistration } from '@app/base/registrations/registrar';
+import type { BasicProfile } from "@ceramicstudio/idx-constants";
+
+export interface Profile {
+  ens: Registration | null;
+  idx: BasicProfile | null;
+}
 
 export enum AddressType {
   Contract,
@@ -32,6 +40,10 @@ export function isAddressEqual(left: string, right: string): boolean {
 
 export function formatBalance(n: BigNumber): string {
   return ethers.utils.commify(parseFloat(ethers.utils.formatUnits(n)).toFixed(2));
+}
+
+export function formatCAIP10Address(address: string, protocol: string, impl: number): string {
+  return `${address.toLowerCase()}@${protocol}:${impl.toString()}`;
 }
 
 export function formatAddress(addr: string): string {
@@ -71,6 +83,11 @@ export function isRadicleId(input: string): boolean {
 // Check whether the input is a URL.
 export function isUrl(input: string): boolean {
   return /^https?:\/\//.test(input);
+}
+
+// Check whether the input is a DID
+export function isDid(input: string): boolean {
+  return /^did:[a-zA-Z0-9]+:[a-zA-Z0-9]+$/.test(input);
 }
 
 // Check whether the input is an Ethereum address.
@@ -180,6 +197,38 @@ export async function identifyAddress(address: string, config: Config): Promise<
 // Resolve a label under the radicle domain.
 export async function resolveLabel(label: string, config: Config): Promise<string | null> {
   return config.provider.resolveName(`${label}.${config.registrar.domain}`);
+}
+
+export async function lookupAddress(address: string, config: Config): Promise<Profile>  {
+  const profile: Profile = { ens: null, idx: null };
+
+  try {
+    const [ens, idx] = await Promise.allSettled([
+      resolveEnsProfile(address, config),
+      resolveIdxProfile(formatCAIP10Address(address, "eip155", config.network.chainId), config)
+    ]);
+
+    if (ens.status == "fulfilled") profile.ens = ens.value;
+    if (idx.status == "fulfilled") profile.idx = idx.value;
+  } catch (error) {
+    console.error(error);
+  }
+
+  return profile;
+}
+
+// Resolves an IDX profile or return null
+export async function resolveIdxProfile(caip10: string, config: Config): Promise<BasicProfile | null> {
+  return config.idx.client.get<BasicProfile>("basicProfile", caip10);
+}
+
+// Resolves an ENS profile or return null
+export async function resolveEnsProfile(address: string, config: Config): Promise<Registration | null> {
+  const label = await config.provider.lookupAddress(address);
+  if (label && await resolveLabel(parseEnsLabel(label, config), config)) {
+    return await getRegistration(label, config);
+  }
+  return null;
 }
 
 // Check whether a Gnosis Safe exists at an address.
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,11 @@
   "packages": {
     "": {
       "dependencies": {
+        "@ceramicnetwork/blockchain-utils-linking": "^1.0.2",
+        "@ceramicnetwork/http-client": "^1.0.6",
+        "@ceramicnetwork/stream-caip10-link": "^1.0.6",
+        "@ceramicstudio/idx": "^0.12.1",
+        "@ceramicstudio/idx-constants": "^1.0.0",
         "@gnosis.pm/safe-core-sdk": "^0.2.0",
         "@gnosis.pm/safe-core-sdk-types": "^0.1.1",
         "@gnosis.pm/safe-service-client": "^0.1.0-alpha.0",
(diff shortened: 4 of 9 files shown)


cloudhead:
This looks like a good start. Let's discuss it Monday.

src/utils.ts
```diff
@@ -173,6 +193,35 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
...
+    const content = await resolveIdxProfile(did, config);
+    if (content) return { type: "idx", content };
+  }
+  // If everything fails return address
+  return { type: "address", content: { name: address }};
+}
+
+// Resolves an existant idx profile
+export async function resolveIdxProfile(did: string, config: Config): Promise<radProfile | null> {
+  return config.idx.api.get<radProfile>("radProfile", did);
```
cloudhead, line 218: Is "radProfile" a string we choose when we create the document type? Are there not existing document types for people who already have an IDX?

src/utils.ts
```diff
@@ -173,6 +193,35 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
...
+  // First get eventual ENS Registration
+  let label: string | null = null;
+  try {
+    label = await config.provider.lookupAddress(address);
+  } catch (error) {
+    console.log("LookupAddress: Not able to get ENS Name, trying with IDX");
+  }
+  if (label && await config.provider.resolveName(label)) {
+    return {type: "ens", content: await getRegistration(label, config) as Registration };
+  }
```
cloudhead, line 206: I think we can move this block into the `try {}` block, and the `label` as well.

src/utils.ts
```diff
@@ -173,6 +193,35 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
...
+  // Else lookup IDX profile
+  if (did && isDid(did)) {
+    const content = await resolveIdxProfile(did, config);
+    if (content) return { type: "idx", content };
+  }
+  // If everything fails return address
+  return { type: "address", content: { name: address }};
+}
```
cloudhead, line 214: I'm not sure how this function will be used, so can't really comment on whether it makes sense. But one issue I can see is that it returns either an IDX profile, or an ENS name, where it might be more useful to return both, ie. anything we can find, and let the frontend component decide what to display.

src/utils.ts
```diff
@@ -64,6 +79,11 @@ export function isUrl(input: string): boolean {
   return /^https?:\/\//.test(input);
 }
 
+// Check whether the input is a did
```
cloudhead, line 82: In comment strings, we can be precise with the acronyms, eg. `DID` here.

src/utils.ts
```diff
@@ -3,6 +3,21 @@ import type { BigNumber } from "ethers";
...
+import type { Registration } from "@app/base/registrations/registrar";
+import { getRegistration } from '@app/base/registrations/registrar';
+
+export interface radProfile {
+  name: string;
+  avatar: string;
+  url: string;
+  github: string;
+  twitter: string;
+}
```
cloudhead, line 15: Probably this is something I'd move to a `profile.ts` file, if that makes sense, and it should just be called `Profile`.

Is eslint okay with non-capitalized type names?

src/utils.ts
```diff
@@ -173,6 +193,35 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
...
+
+// Resolves an existant idx profile
+export async function resolveIdxProfile(did: string, config: Config): Promise<radProfile | null> {
+  return config.idx.api.get<radProfile>("radProfile", did);
+}
+
+export function getCAIP10Address(address:string, protocol: string, impl: number): string {
+  return `${address}@${protocol}:${impl.toString()}`;
+}
```
cloudhead, line 223: Functions like this are called `formatXXX` in the codebase, as it's clearer that no I/O is performed and no state is fetched.

src/utils.ts
cloudhead: Seems like it's okay by default, so I added a lint for it on master, with a couple of other things, so our variable/type names are consistent!

src/utils.ts
sebastinez: The `radProfile` string we chose when we define it in `config.json` as the definition, which we use in the initialization of the IDX instance.
```js
// config.json
"idx": {
  "definition": "kjzl6cwe1jw14b4i43fpebvxonfyu8ei1f1a3mg32xhlmeugddnsa943dvgkzly"
}
// config.ts
const idx = new IDX({ ceramic,
  aliases: {
    radProfile: config.idx.definition
  }
});
```
We could eventually use the [`basicProfile` definition](https://developers.idx.xyz/guides/definitions/default/#basic-profile) but it does not correlate with the information that gets saved on the ENS record.
Eventually when we allow users to save their profiles through the app, we could allow users to port their `basicProfile` definition to a `radProfile`

src/utils.ts
sebastinez: Okay yeah that would be fine too.
To avoid overfetching I was thinking of not fetching everything all the time, just the necessary according to the defined heuristic.

src/utils.ts
cloudhead: Hmm I see. Perhaps it's worth supporting basicProfile first though, and seeing if there's a way of extending it?

sebastinez: 008cd890ab5a0b7c2b668509bec1fbbc098af819 solves this

src/utils.ts
sebastinez: 77762ea49c0d5ea4e8816b77921861ab52941991 solves this

src/utils.ts
sebastinez: 27e9a28f819fb09a9c4ea3796d96362b28f89658 solves this

src/utils.ts
sebastinez: 27e9a28f819fb09a9c4ea3796d96362b28f89658 solves this

src/utils.ts
sebastinez: 27e9a28f819fb09a9c4ea3796d96362b28f89658 solves this

src/utils.ts
sebastinez: This commit 008cd890ab5a0b7c2b668509bec1fbbc098af819 uses BasicProfile as type of `@ceramicstudio/idx-constants`

src/base/idx/View.svelte
```diff
@@ -0,0 +1,34 @@
...
+
+      const accountLink = await Caip10Link.fromAccount(ceramicClient, caip10.toLowerCase());
+      await accountLink.setDid(
+        did,
+        ethAuthProvider
+      );
+
+    }
+  });
+</script>
```
cloudhead, line 34: Based on what this does, it should probably called `Link`, not View.

src/base/idx/View.svelte
```diff
@@ -0,0 +1,34 @@
...
+    if (config.signer) {
+      const address = await config.signer.getAddress();
+      console.log(address);
+      console.log(did);
+      const ethAuthProvider = new EthereumAuthProvider(
+        // TODO: Using the window.ethereum provider here since the config provider is throwing error.
+        config.provider, address
+      );
+
+      const ceramicClient = new CeramicClient(ceramic.api);
```
cloudhead, line 23: Should we add this to Config? Eg. `config.ceramic.client`

cloudhead: I see it's already in Config actually, but under idx

src/utils.ts
```diff
@@ -173,6 +195,37 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
   return config.provider.resolveName(`${label}.${config.registrar.domain}`);
 }
 
+export async function lookupAddress(address: string, config: Config): Promise<Profile | null>  {
+  const profile: Profile = {} as Profile;
+  try {
+
+    const ens = await resolveEnsProfile(address, config);
+    if (ens) profile.ens = ens;
+    const idx = await resolveIdxProfile(formatCAIP10Address(address, "eip155", 3), config);
```
cloudhead, line 204: We don't want to execute these serially, as they don't depend on each other. You can use `Promise.allSettled`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

src/utils.ts
```diff
@@ -3,6 +3,19 @@ import type { BigNumber } from "ethers";
...
+
+export interface Profile {
+  ens?: Registration;
+  idx?: BasicProfile;
+}
+
+export type Lookup =
+    { type: "address"; content: { name: string } }
+  | { type: "idx"; content: BasicProfile }
+  | { type: "ens"; content: Registration };
```
cloudhead, line 18: I think we can remove this?

src/utils.ts
```diff
@@ -173,6 +195,37 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
...
+    if (ens) profile.ens = ens;
+    const idx = await resolveIdxProfile(formatCAIP10Address(address, "eip155", 3), config);
+    if (idx) profile.idx = idx;
+  } catch (error) {
+    console.error(error);
+  }
+
+  console.log("Profiles:", profile);
+
+  return profile || null;
```
cloudhead, line 212: Not sure what you expect here, but isn't `{} || null` going to return `{}`?

I think we should just return `Profile` from this function anyway. The ui code can check if the ens or idx keys are set.

src/base/idx/View.svelte
cloudhead: "View" would be used for `/idx/:did`

src/base/idx/View.svelte
cloudhead: "View" would be used for `/idx/:did`

src/utils.ts
sebastinez: Yeah it's a leftover

src/base/idx/View.svelte
sebastinez: Yeah, we could eventually do two separate properties

src/utils.ts
sebastinez: Yeah you're right, wanted to be explicit first but, even if nothing gets resolved it returns an empty object which is not null

sebastinez:
@cloudhead I think this PR require some squashing, would squash it down to the essential commits, what do you think.

cloudhead:
Yes!

src/utils.ts
```diff
@@ -208,9 +208,12 @@ export async function lookupAddress(address: string, config: Config): Promise<Pr
...
 
     const ens = await resolveEnsProfile(address, config);
-    if (ens) profile.ens = ens;
-    const idx = await resolveIdxProfile(formatCAIP10Address(address, "eip155", 3), config);
-    if (idx) profile.idx = idx;
+    const idx = await resolveIdxProfile(formatCAIP10Address(address, "eip155", config.network.chainId), config);
+    Promise.allSettled([ens, idx]).then(([ens, idx]) => {
+      profile.ens = ens.status == "fulfilled" ? ens.value : null;
+      profile.idx = idx.status == "fulfilled" ? idx.value : null;
+    });
```
cloudhead, line 215: You have to await this promise, otherwise you're returning before it completes?

Eg. `let [ens, idx] = await Promise.allSettled(...)`

cloudhead: I also believe you can simplify this, as `ens.value` will be `undefined` if the promise is rejected, so you don't need the coniditional and you can remove `null` from the type.

src/config.ts
```diff
@@ -62,7 +63,8 @@ export class Config {
     this.signer = signer;
     this.gasLimits = gasLimits;
     this.abi = config.abi;
-    this.idx = { api: idx, ceramic };
+    this.idx = { api: idx };
+    this.ceramic = { api: ceramic };
```
cloudhead, line 67: The use of "api" here is confusing as in the other cases it's used to store the api url string. I'd suggest we use `client` if we're storing a client, like for `safe.client` on line 26.

src/utils.ts
sebastinez: > I also believe you can simplify this, as `ens.value` will be `undefined` if the promise is rejected, so you don't need the conditional and you can remove `null`s from the Profile type.

The issue is that when a `PromiseRejectedResult` is returned there is no `ens.value` and this doesn't work for the Typescript compiler.
However we could store the result from the `Promise.allSettled` function and let the UI look if the promise has been resolved or not.

sebastinez: > I also believe you can simplify this, as `ens.value` will be `undefined` if the promise is rejected, so you don't need the conditional and you can remove `null`s from the Profile type.

I see a problem here since if the promise is rejected it has no property `value`, so the type checking fails.
However we could return the result of Promise.allSettled and let the components decide how they read the settled or rejected promise.

cloudhead: Ah true, that's unfortunate. I think your solution could work though.

sebastinez: 8397b7876769a943e9df4d8653014a0c511b2b2a fixes this

sebastinez:
I tried to test if everything was working accordingly, but there is a issue with the `gnosis-safe-core-sdk` package that throws an error and does not allow the page to be loaded..

src/session.ts
```diff
@@ -183,10 +184,15 @@ export const session = derived(state, s => {
   return null;
 });
 
-window.ethereum?.on('chainChanged', () => location.reload());
+window.ethereum?.on('chainChanged', () => {
+  // We disconnect the wallet to avoid out of sync state
+  // between the account address and idx dids
```
cloudhead, line 189:
```suggestion
  // between the account address and IDX DIDs
```

src/utils.ts
```diff
@@ -182,6 +203,42 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
   return config.provider.resolveName(`${label}.${config.registrar.domain}`);
 }
 
+export async function lookupAddress(address: string, config: Config): Promise<Profile>  {
+  const profile: Profile = { ens: null, idx: null };
+
+  try {
+
```
cloudhead, line 210:
```suggestion
```

src/utils.ts
```diff
@@ -182,6 +203,42 @@ export async function resolveLabel(label: string, config: Config): Promise<strin
...
+    ]);
+
+    if (ens.status == "fulfilled") profile.ens = ens.value;
+    if (idx.status == "fulfilled") profile.idx = idx.value;
+
+  } catch (error) {
+    console.error(error);
+  }
+
+  console.log("Profiles:", profile);
```
cloudhead, line 223:
```suggestion
```

cloudhead (commented):
Just a few minor things left and this can be merged!

src/utils.ts
```diff
@@ -73,6 +85,15 @@ export function isUrl(input: string): boolean {
...
 }
 
+// Check whether the input is a DID
+export function isDid(input: string): boolean {
+  return /^did:[a-zA-Z0-9]+:[a-zA-Z0-9]+$/.test(input);
+}
+
+export function isIPFSCID(input: string): boolean {
+  return /^ipfs:\/\/[a-zA-Z0.9]+$/.test(input);
+}
```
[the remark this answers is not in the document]
sebastinez: Hmm yeah you're right, this util function was meant for a later commit.. I was looking for a way to validate the ipfs CID, but since the IDX basicProfile already validates it, eventually it is not even necessary.
Will eventually remove this fn.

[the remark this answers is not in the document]
cloudhead: Ah well up to you.

cloudhead (approved):

cloudhead:
Good stuff!

Issue #33, opened 2021-06-24
Commitment wait time loading bar


cloudhead:
When waiting for the commitment `minAge`, show a loading bar that advances on every block.

sebastinez:
This sounds very interesting! Will look into it and provide something!

cloudhead:
Nice!

PR #41, opened 2021-06-26
Add commitment progress bar

sebastinez:
This PR solves #33

It implements a new component called `BlockTimer` that could eventually be used in other use cases where one has to wait for a specific quantity of blocks to be mined.
It takes a starting block number and the number of blocks it needs to fill up to 100%.

To implement it in the registration process I did some changes to the `Connection` type to implement the `commitmentBlock` and `minAge`, and avoid refetching it later.

P.S. For some reason the `tx` object in the `commit` fn in `registrar.ts` did not include blockNumber or hash, thats why I'm feching the tx receipt to obtain the `commitmentBlock`

### DIFF
--- a/src/BlockTimer.svelte
+++ b/src/BlockTimer.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+  import type { Config } from "@app/config";
+
+  export let config: Config;
+  export let startBlock: number;
+  export let duration: number;
+
+  let currentBlock: number = startBlock;
+
+  config.provider.on("block", (latestBlock: number) => {
+    if (startBlock + duration > currentBlock) currentBlock = latestBlock;
+  });
+</script>
+
+<style>
+  .parent {
+    text-align: center;
+    height: 20px;
+    width: 290px;
+    border-radius: 5px;
+    background-color: var(--color-secondary-background);
+  }
+  .loader {
+    color: #fff;
+    height: 20px;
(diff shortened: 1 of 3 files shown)


src/base/registrations/registrar.ts
```diff
@@ -30,12 +30,21 @@ export enum State {
...
-export const state = writable(State.Connecting);
+export type Connection =
+    { connection: State.Failed }
+  | { connection: State.Connecting }
+  | { connection: State.Committing }
+  | { connection: State.WaitingToRegister; commitmentBlock: number; minAge: number }
+  | { connection: State.Registering }
+  | { connection: State.Registered };
+
+
```
cloudhead, line 41:
```suggestion
```

src/BlockTimer.svelte
```diff
@@ -0,0 +1,34 @@
...
+    border-radius: 5px;
+    background-color: var(--color-secondary-background);
+  }
+  #loader {
+    color: #fff;
+    height: 20px;
+    width: 0px;
+    border-radius: 5px;
+    background-color: var(--color-secondary);
+  }
```
cloudhead, line 29: We generally use classes instead of ids for our components, if nothing but for consitency.

cloudhead (approved):
Awesome, it works great!

cloudhead (approved):

PR #44, opened 2021-06-29
User Identity Components

sebastinez:
This PR works on #17 

It brings IDX and ENS identities to all the components where some identity is shown.
For that it implements some new util functions to relieve the components load to create the information to be displayed.

The implementation of the Radicle URN, requires development in collaboration with ceramic-js and the radicle-link team so it will be for another PR.

I welcome feedback since there are some new functions and ways to get profile data into the components.

### DIFF
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -2,12 +2,11 @@
   import { onMount } from 'svelte';
   import { link } from 'svelte-routing';
   import { ethers } from 'ethers';
-  import { safeLink, explorerLink, identifyAddress, formatAddress, AddressType } from '@app/utils';
+  import { safeLink, explorerLink, identifyAddress, formatAddress, AddressType, parseEnsLabel } from '@app/utils';
+  import { Profile } from '@app/profile';
   import Loading from '@app/Loading.svelte';
   import Avatar from "@app/Avatar.svelte";
   import type { Config } from '@app/config';
-  import type { Registration } from '@app/base/registrations/registrar';
-  import { getRegistration } from '@app/base/registrations/registrar';
 
   export let address: string;
   export let config: Config;
@@ -15,24 +14,21 @@
   export let noBadge = false;
   export let noAvatar = false;
   export let compact = false;
+  // This property allows components eg. Header.svelte to pass a resolved profile object.
+  export let profile: Profile | null = null;
 
   let checksumAddress = compact
     ? formatAddress(address)
     : ethers.utils.getAddress(address);
   let addressType: AddressType | null = null;
-  let addressName: string | null = null;
-  let info: Registration | null;
 
   onMount(async () => {
     identifyAddress(address, config).then((t: AddressType) => addressType = t);
-    if (resolve) {
-      addressName = await config.provider.lookupAddress(address);
-      if (addressName) {
-        info = await getRegistration(addressName, config);
-      }
+    if (resolve && !profile) {
+      Profile.get(address, config).then(p => profile = p);
     }
   });
-  $: addressLabel = addressName ?? checksumAddress;
+  $: addressLabel = profile?.name ? compact ? parseEnsLabel(profile.name, config) : profile.name : checksumAddress;
 </script>
 
 <style>
@@ -59,7 +55,7 @@
 
 <div class="address" title={address} class:no-badge={noBadge}>
   {#if !noAvatar}
-    <Avatar inline source={info?.avatar ?? address} />
+    <Avatar inline source={profile?.avatar ?? address} />
   {/if}
   {#if addressType === AddressType.Org}
     <a use:link href={`/orgs/${address}`}>{addressLabel}</a>
--- a/src/Form.svelte
+++ b/src/Form.svelte
@@ -4,6 +4,7 @@
     value: string | null;
     label?: string;
     placeholder?: string;
+    resolve: boolean;
     editable: boolean;
   }
 </script>
@@ -105,7 +106,7 @@
                 <a class="link" href="{field.value}" target="_blank">{field.value}</a>
               </span>
             {:else if isAddress(field.value)}
-              <Address address={field.value} {config} />
+              <Address resolve={field.resolve} address={field.value} {config} />
             {:else}
               {field.value}
             {/if}
--- a/src/base/orgs/Create.svelte
+++ b/src/base/orgs/Create.svelte
@@ -130,7 +130,7 @@
         <span><Address address={org.address} {config} /></span>
 
         <span class="label">Owner</span>
-        <span><Address address={org.owner} {config} /></span>
+        <span><Address resolve address={org.owner} {config} /></span>
       </div>
     </span>
 
--- a/src/base/orgs/List.svelte
+++ b/src/base/orgs/List.svelte
@@ -1,9 +1,15 @@
 <script lang="ts">
   import { Link } from 'svelte-routing';
   import type { Org } from '@app/base/orgs/Org';
-  import Blockies from '@app/Blockies.svelte';
+  import Avatar from '@app/Avatar.svelte';
+  import { Profile } from '@app/profile';
+  import type { Config } from '@app/config';
+  import Loading from '@app/Loading.svelte';
 
+  export let config: Config;
   export let orgs: Org[];
+
+  const orgsAddresses = orgs.map(org => org.address);
 </script>
 
 <style>
@@ -13,14 +19,30 @@
     margin: 3rem;
     display: inline-block;
   }
+  .list {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: wrap; 
+  }
+  .loading {
+    padding: 3rem 0;
+  }
 </style>
 
-{#each orgs as org}
-  <div class="org">
-    <Link to={`/orgs/${org.address}`}>
-      <Blockies glowOnHover address={org.address} />
-    </Link>
+{#await Profile.getMulti(orgsAddresses, config)}
+  <div class="loading">
+    <Loading center /> 
   </div>
-{:else}
-  <slot />
-{/each}
+{:then profiles}
+  <div class="list">
+    {#each profiles as profile}
+      <div class="org">
+        <Link to={`/orgs/${profile.address}`}>
+          <Avatar glowOnHover source={profile.avatar ?? profile.address} />
+        </Link>
+      </div>
+    {:else}
+      <slot />
+    {/each}
+  </div>
+{/await}
--- a/src/base/orgs/TransferOwnership.svelte
+++ b/src/base/orgs/TransferOwnership.svelte
@@ -2,10 +2,9 @@
   import { onMount, createEventDispatcher } from 'svelte';
   import Modal from '@app/Modal.svelte';
   import type { Config } from '@app/config';
-  import { formatAddress } from '@app/utils';
+  import { formatAddress, isAddress } from "@app/utils";
   import Loading from '@app/Loading.svelte';
   import { assert } from '@app/error';
-  import * as utils from '@app/utils';
 
   import type { Org } from './Org';
 
@@ -38,7 +37,7 @@
   const onSubmit = async () => {
     assert(newOwner);
 
-    if (! utils.isAddress(newOwner)) {
+    if (! isAddress(newOwner)) {
       state = State.Failed;
       error = `"${newOwner}" is not a valid Ethereum address.`;
       return;
@@ -58,7 +57,7 @@
   };
 </script>
 
-{#if state === State.Success}
+{#if state === State.Success && newOwner}
   <Modal floating small>
     <div slot="title">
       ✅
--- a/src/base/orgs/View.svelte
+++ b/src/base/orgs/View.svelte
@@ -1,39 +1,28 @@
 <script lang="ts">
-  import * as ethers from 'ethers';
-  import { onMount } from 'svelte';
   import type { SvelteComponent } from 'svelte';
   import Link from '@app/Link.svelte';
   import type { Config } from '@app/config';
-  import type { Registration } from '@app/base/registrations/registrar';
-  import { getRegistration } from '@app/base/registrations/registrar';
   import { parseEnsLabel, explorerLink } from '@app/utils';
   import { session } from '@app/session';
   import Loading from '@app/Loading.svelte';
   import Modal from '@app/Modal.svelte';
   import Error from '@app/Error.svelte';
   import Icon from '@app/Icon.svelte';
-  import Blockies from '@app/Blockies.svelte';
   import SetName from '@app/ens/SetName.svelte';
   import Project from '@app/base/projects/Widget.svelte';
   import Address from '@app/Address.svelte';
+  import Avatar from '@app/Avatar.svelte';
   import Message from '@app/Message.svelte';
   import * as utils from '@app/utils';
 
-  import { Org } from './Org';
-  import TransferOwnership from './TransferOwnership.svelte';
+  import { Org } from '@app/base/orgs/Org';
+  import TransferOwnership from '@app/base/orgs/TransferOwnership.svelte';
+  import { Profile } from '@app/profile';
 
   export let address: string;
   export let config: Config;
 
-  let registration: Registration | null = null;
-  let name: string | null = null;
-
   const back = () => window.history.back();
-
-  onMount(async () => {
-    name = await config.provider.lookupAddress(address);
-    if (name) registration = await getRegistration(name, config);
-  });
 
   let setNameForm: typeof SvelteComponent | null = null;
   const setName = () => {
@@ -44,8 +33,6 @@
   const transferOwnership = () => {
     transferOwnerForm = TransferOwnership;
   };
-
-  $: label = name && parseEnsLabel(name, config);
   $: isOwner = (org: Org): boolean => $session
     ? utils.isAddressEqual(org.owner, $session.address)
     : false;
@@ -97,11 +84,6 @@
     width: 64px;
     height: 64px;
   }
-  .avatar img {
-    width: 100%;
-    height: 100%;
-    border-radius: 50%;
-  }
   .links {
     display: flex;
     align-items: center;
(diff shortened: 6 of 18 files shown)


cloudhead:
This looks directionally good from a quick look!

The only thing I'd say is that we could replace the `getInfoFromProfile` with a `Profile` class, so that you'd get something more like: `profile.avatar()` for eg., and that would figure out what to return.

sebastinez:
Okay @cloudhead let me know what you think, I took the liberty to implement user identities in more places than just the view pages for example:
- Transfer Org Ownership a ENS subdomain can be used.
- The Orgs creation modals show resolved user identities if available
- All the `formatAddress` occurrences have been changed to `<Address>` components, in general without avatar and badge, please let me know if you think somewhere where it should be shown or hidden.
- The `isRadicleId` check in the avatar component allows the blockies to be shown if the input isn't an ETH address
- In `Org.ts` I removed the lookupAddress method, to enforce the usage of the Profile class
- Resolving of the Radicle URNs will be kept open for later when we can implement a RadicleAuthProvider required by Ceramic

src/Form.svelte
```diff
@@ -104,8 +104,10 @@
               <span>
                 <a class="link" href="{field.value}" target="_blank">{field.value}</a>
               </span>
+            {:else if field.name === "owner"}
+              <Address address={field.value} resolve {config} />
```
cloudhead, line 108: The 'Form' component shouldn't change its behavior based on field names, this breaks separation of concerns -- rather the parent component should configure the form to display the "owner" field a certain way.

We can add a `resolve` attribute to fields, for example.

src/Form.svelte
```diff
@@ -104,8 +104,10 @@
...
             {:else if isAddress(field.value)}
-              <Address address={field.value} {config} />
+              <Address noAvatar address={field.value} {config} />
```
cloudhead, line 110: Any reason we're not showing avatars anymore here?

src/base/orgs/Create.svelte
```diff
@@ -126,10 +126,10 @@
     <span slot="body">
       <div class="fields">
         <span class="label">Address</span>
-        <span><Address address={org.address} {config} /></span>
+        <span><Address noAvatar address={org.address} {config} /></span>
```
cloudhead, line 129: I also wonder here why we are changing this?

src/base/orgs/View.svelte
```diff
@@ -191,74 +169,73 @@
...
-        <p class="highlight"><strong>{address}</strong></p>
+        <p class="highlight"><Address compact {address} {config} /></p>
         <p>Sorry, there is no Org at this address.</p>
         {#if utils.isAddress(address)}
           <p>
             <a href={explorerLink(address, config)} class="link" target="_blank">View in explorer</a>
             <span class="faded">↗</span>
           </p>
-        {/if}
+          {/if}
```
cloudhead, line 238: Typo?

src/utils.ts
```diff
@@ -90,6 +92,12 @@ export function isDid(input: string): boolean {
   return /^did:[a-zA-Z0-9]+:[a-zA-Z0-9]+$/.test(input);
 }
 
+export function isENSLabel(input: string, config: Config): boolean {
+  const domain = config.registrar.domain.replace(".", "\\.");
+  const regEx = new RegExp(`^[a-zA-Z0-9]+.${domain}$`);
+  return regEx.test(input);
+}
```
cloudhead, line 99: This would be `isEnsName`, the label is just the "foo" in "foo.radicle.eth".

src/utils.ts
```diff
@@ -101,6 +109,12 @@ export function getSearchParam(key: string, location: RouteLocation): string | n
   return params.get(key);
 }
 
+// Checks an array for a pattern, if found returns element else returns undefined
+export function hasArrayElement(array: string[], pattern: string): string | undefined {
+  const reg = new RegExp(pattern);
+  return array.find(a => reg.test(a));
+}
```
cloudhead, line 116: I find this util function a little dubious, based on the two places it's used. I think it's perfectly fine to use `Array::find()` instead, we don't really save much!

src/Address.svelte
```diff
@@ -20,19 +19,20 @@
     ? formatAddress(address)
     : ethers.utils.getAddress(address);
   let addressType: AddressType | null = null;
-  let addressName: string | null = null;
-  let info: Registration | null;
+  let addressName: string | undefined;
+  let avatarSrc: string | undefined;
```
cloudhead, line 23: I don't see this used anywhere, perhaps we can remove it?

src/Form.svelte
sebastinez: I removed the avatars, since the place we wanted to show avatars was in the owner field.
We can eventually add a `noAvatar` attribute to fields were we don't want to show avatar eg: in the address field in the `Registrations::View`.

src/base/orgs/Create.svelte
sebastinez: Since we are showing only the address there without reference to owners or users I added the `noAvatar` attribute.
But it's also the address of the orgs, which is kind of a edge case imo.

src/utils.ts
sebastinez: Yeah it's a reminiscent of the need to return `null` I think we can remove it safely

src/base/orgs/View.svelte
sebastinez: Yeah it seems a wild indentation appeared :man_shrugging:

src/base/orgs/View.svelte
```diff
@@ -141,33 +120,32 @@
...
-            <img src={registration.avatar} alt="avatar" />
-          {:else}
-            <Blockies address={org.address} />
-          {/if}
+          <Avatar source={profile.avatar() ?? address} />
         </div>
         <div class="info">
           <span class="title bold">
-            {registration ? label : ethers.utils.getAddress(address)}
+            <Address resolve noAvatar {address} {config} />
```
cloudhead, line 132: We want `noBadge` here also

src/base/orgs/List.svelte
```diff
@@ -13,14 +16,26 @@
...
+  {#await Profile.get(org.address, config)}
+    <!-- Avoid using <Loading /> here since it fills the page with loading elements -->
+    <div /> 
+  {:then profile}
+    <div class="org">
+      <Link to={`/orgs/${org.address}`}>
+        <Avatar glowOnHover source={profile.avatar() ?? org.address} />
+      </Link>
+    </div>
+  {/await}
```
cloudhead, line 37: I think we will eventually want a more efficient way of doing this, as right now it does a bunch of requests for each org :thinking:

src/base/orgs/TransferOwnership.svelte
```diff
@@ -38,10 +38,20 @@
   const onSubmit = async () => {
     assert(newOwner);
 
+    // Check if newOwner is valid ENS label or Ethereum address
```
cloudhead, line 41:
```suggestion
    // Check if newOwner is valid ENS name or Ethereum address
```

src/base/orgs/TransferOwnership.svelte
```diff
@@ -38,10 +38,20 @@
...
     if (! utils.isAddress(newOwner)) {
-      state = State.Failed;
-      error = `"${newOwner}" is not a valid Ethereum address.`;
-      return;
+      if (! utils.isENSName(newOwner, config)) {
+        newOwner = `${newOwner}.${config.registrar.domain}`;
```
cloudhead, line 44: This behavior is strange: if it's not an ENS name, it adds "radicle.eth" to the input field.

I would instead go to a failed state here, if we have neither an address nor an ENS name.

src/base/orgs/TransferOwnership.svelte
```diff
@@ -98,7 +108,7 @@
 
     <div slot="body">
       {#if state == State.Idle}
-        <input type="text" size="40" disabled={state !== State.Idle} bind:this={input} bind:value={newOwner} />
+        <input type="text" size="40" disabled={state !== State.Idle} placeholder="Enter Ethereum address or ENS subdomain" bind:this={input} bind:value={newOwner} />
```
cloudhead, line 111: TBH I'm not a fan of this change as it's not generalized across address inputs and not really thought through, eg. it only supports .radicle.eth domains. I would revert this to the previous behavior and we can look at the problem more comprehensively.

src/base/orgs/View.svelte
```diff
@@ -191,67 +169,66 @@
...
-            <div class="loading"><Loading small /></div>
-          {:then name}
-            {#if name}
-              <Link to={`/registrations/${label}`}>{name}</Link>
-            {:else}
-              <span class="subtle">Not set</span>
-            {/if}
-          {/await}
+          {#if profile.name()}
+            <Link to={`/registrations/${parseEnsLabel(profile.name() ?? "", config)}`}>{profile.name()}</Link>
```
cloudhead, line 173: Not sure about this: in what cases exactly are we returning `""`?

src/base/orgs/View.svelte
cloudhead, line 132: Are we resolving twice here? Since we already have the profile but the Address is having to do the lookup again? I think this is a case where we should use a conditional, or pass an optional `label` to Address. Not sure which is a better idea..

src/base/orgs/TransferOwnership.svelte
```diff
@@ -58,15 +68,15 @@
...
   <Modal floating small>
     <div slot="title">
       ✅
     </div>
 
     <div slot="subtitle">
-      The ownership of <strong>{formatAddress(org.address)}</strong> was
-      successfully transfered to <strong>{newOwner}</strong>.
+      The ownership of <Address address={org.address} {config} resolve noAvatar noBadge compact /> was
+      successfully transfered to <Address address={newOwner} {config} resolve compact noAvatar noBadge />.
```
cloudhead, line 79: I would prefer to leave these as they were for now. It exhibits strange behavior sometimes and when it fails to load there is just a gap in the text.

src/base/orgs/TransferOwnership.svelte
```diff
@@ -88,7 +98,7 @@
       {:else if state == State.Pending}
         Transaction is being processed by the network...
       {:else if state == State.Idle}
-        Transfer the ownership of Org <strong>{formatAddress(org.address)}</strong> to a new address.
+        Transfer the ownership of Org <Address {config} address={org.address} resolve compact noAvatar noBadge /> to a new address.
```
cloudhead, line 101: Same here, I think inline addresses should remain pure text.

src/base/orgs/List.svelte
```diff
@@ -13,14 +16,26 @@
...
 
+<div class="list">
 {#each orgs as org}
-  <div class="org">
-    <Link to={`/orgs/${org.address}`}>
-      <Blockies glowOnHover address={org.address} />
-    </Link>
-  </div>
+  {#await Profile.get(org.address, config)}
+    <!-- Avoid using <Loading /> here since it fills the page with loading elements -->
```
cloudhead, line 29: I agree, but at the same time, now we have an intermediate state where there's no loading indicator at all, and just an empty area. Perhaps the ideal would be to have a `Profile.getMulti([string], config)` that sends the requests in parallel and only returns when all have resolved. This way we can show a single loader and also load faster.

cloudhead: image

src/base/orgs/TransferOwnership.svelte
cloudhead: EDIT: See comments below, I think it's best we just remove this behavior for now.

src/base/orgs/View.svelte
sebastinez: I think to avoid introduce noise into the `<Address>` component we should check i.e.: `profile.name() ?? address` and remove the `<Address>` component here.

src/base/orgs/View.svelte
sebastinez: Yeah the issue here is that `profile.name()` returns `string | undefined` and `parseEnsLabel` demands a string.
We could eventually update `parseEnsLabel` and the functions that use said function to be able to handle a `undefined` value.
Or the other option would be to resolve the Profile in an onMount function and pass it to the components already resolved.

src/base/orgs/List.svelte
sebastinez: I think this can be resolved with `Profile.getMulti([string], config)`

src/base/orgs/View.svelte
sebastinez: Ok I pushed a commit with a posible solution, it's not the nicest because it starts replacing `null` with `undefined` and does it only on two individual functions.. And it also changes the behaviour of these two functions. b4dfa25963be687eff8d208543abd74da10b4212
Let me know what you think, and as another option we could do the resolving in an onMount function

src/profile.ts
```diff
@@ -10,9 +10,11 @@ export interface IProfile {
 
 export class Profile {
   profile: IProfile;
+  address: string;
 
-  constructor(profile: IProfile) {
+  constructor(profile: IProfile, address: string) {
```
sebastinez, line 15: This is a solution to the issue that when creating multiple Orgs Profiles I don't have access to their initial address only their Profile.
So know the address forms part of the profile class instance

src/base/orgs/View.svelte
cloudhead: Yeah I don't like the solution in b4dfa25 because we lose type safety: the compiler will not error if we pass an undefined variable in the function..

cloudhead:
Couple of other things I noticed:

* Here we shouldn't show `radicle.eth`, unless it's a different domain eg. `lol.eth`.
image

* The Connect button, if hovered slightly to the left will flicker like crazy, as the size keeps changing. Is there a way we can maybe show "Disconnect" without changing the button size? Perhaps by setting `visibility: hidden` on the content and making the "Disconnect" text float?

src/base/orgs/View.svelte
```diff
@@ -191,67 +169,66 @@
...
-            {/if}
-          {/await}
+            {:then authorized}
+              {#if authorized}
+                <button class="tiny secondary" on:click={setName}>
+                  Set
+                </button>
+              {/if}
+            {/await}
+          </div>
```
cloudhead, line 188: Indentation around here is messed up :)

src/base/orgs/View.svelte
cloudhead: It's annoying because the `{#if profile.name()}` should provide the type assertion that the value is not `undefined`, but it seems like svelte-check is not smart enough here..

cloudhead: It seems like switching to accessors for the Profile class gets around this issue:

```diff
diff --git a/src/profile.ts b/src/profile.ts
index 9c6049a..6cf8bd1 100644
--- a/src/profile.ts
+++ b/src/profile.ts
@@ -39,7 +39,7 @@ export class Profile {
   }

   // Using undefined as return type if nothing to be returned since it works better with <a href> links
-  name(): string | undefined {
+  get name(): string | undefined {
     if (this.profile?.ens?.name) return this.profile.ens.name;
     else if (this.profile?.idx?.name) return this.profile.idx.name;
     else return undefined;
@@ -169,8 +169,8 @@
         <!-- Name -->
         <div class="label">Name</div>
         <div>
-          {#if profile.name()}
-            <Link to={`/registrations/${parseEnsLabel(profile.name(), config)}`}>{profile.name()}</Link>
+          {#if profile.name}
+            <Link to={`/registrations/${parseEnsLabel(profile.name, config)}`}>{profile.name}</Link>
           {:else}
```

sebastinez: > It seems like switching to accessors for the Profile class gets around this issue:
> 
> ```diff
> diff --git a/src/profile.ts b/src/profile.ts
> index 9c6049a..6cf8bd1 100644
> --- a/src/profile.ts
> +++ b/src/profile.ts
> @@ -39,7 +39,7 @@ export class Profile {
>    }
> 
>    // Using undefined as return type if nothing to be returned since it works better with <a href> links
> -  name(): string | undefined {
> +  get name(): string | undefined {
>      if (this.profile?.ens?.name) return this.profile.ens.name;
>      else if (this.profile?.idx?.name) return this.profile.idx.name;
>      else return undefined;
> @@ -169,8 +169,8 @@
>          
>          <div class="label">Name</div>
>          <div>
> -          {#if profile.name()}
> -            <Link to={`/registrations/${parseEnsLabel(profile.name(), config)}`}>{profile.name()}</Link>
> +          {#if profile.name}
> +            <Link to={`/registrations/${parseEnsLabel(profile.name, config)}`}>{profile.name}</Link>
>            {:else}
> ```

Ooh wow that's great! Ty will do the refactor right away

sebastinez:
> * The Connect button, if hovered slightly to the left will flicker like crazy, as the size keeps changing. Is there a way we can maybe show "Disconnect" without changing the button size? Perhaps by setting `visibility: hidden` on the content and making the "Disconnect" text float?

I think I found a way to improve upon, getting the final width after loading the button and on mouseover this width is being applied, it has one downside that when changing accounts it keeps this width.. Will check if I find an improvement for that.

> * The loading on the /orgs page is a bit janky still, but this is something I can look into after we merge

I think with an additional div with the `loading` class I added this is looking much better

cloudhead:
One minor issue I'm seeing is that sometimes the loading of the org members looks misaligned:

image

I wonder if we can avoid one of the loading stages, as right now I think it loads in at least 3 stages:
1. load list
2. load each member individually
3. load member labels

In terms of the Connect button, I guess you're still working on it, because I see this atm:

image

cloudhead:
Cool, do you want to resolve the conflicts before I review?

sebastinez:
> Cool, do you want to resolve the conflicts before I review?

After your review I'll squash it down to one commit.

src/Address.svelte
```diff
@@ -2,37 +2,33 @@
...
-  import { getRegistration } from '@app/base/registrations/registrar';
 
   export let address: string;
   export let config: Config;
   export let resolve = false;
   export let noBadge = false;
   export let noAvatar = false;
   export let compact = false;
+  // This property allows components eg. Header.svelte to pass a resolved profile object.
+  export let profile: Profile | null = null;
```
cloudhead, line 18: I actually like this change generally!

sebastinez: Yeah I think it could be used to avoid building a user profile when a component already did all the heavy lifting.

cloudhead (approved):
:partying_face:

PR #49, opened 2021-07-10
Feature: Implement wallet connect

Khay-EMMA:


package.json
```diff
@@ -32,11 +33,22 @@
     "@stardazed/streams": "^3.1.0",
     "@types/marked": "^2.0.3",
     "buffer": "^6.0.3",
+    "big.js": "^6.1.1",
```
cloudhead, line 36: Do we need this? Ethers.js already comes with a BigNumber library.

src/components/ModalOverlay.svelte
```diff
@@ -0,0 +1,58 @@
+<!--
```
cloudhead, line 1: We shouldn't need this component, as we have our own modal system.

package.json
Khay-EMMA: no we dont. I need to remove unnecessary packages also

src/modal.ts
```diff
@@ -0,0 +1,48 @@
...
+): void => {
+  const stored = get(store);
+
+  if (stored && stored.modalComponent === modalComponent) {
+    hide();
+    return;
+  }
+
+  show(modalComponent, onHide, modalComponentProps);
+};
```
cloudhead, line 48: Same here, we shouldn't need this as we already have modals.

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,76 @@
...
+      background: "white",
+      ecl: "M"
+    });
+
+    const onClickConnect = () => {
+
+      console.log(config, "from svelete");
+    
+      state.connectMetamask(config);
+};
```
cloudhead, line 29: Indentation here is wrong -- but why doesn't the eslint "indent" rule catch this :roll_eyes:

Khay-EMMA: The new commit you made to package.json fixed this

src/session.ts
```diff
@@ -3,6 +3,11 @@ import type { BigNumber } from 'ethers';
 import type { TransactionReceipt, TransactionResponse } from '@ethersproject/providers';
 import { Config, getConfig } from "@app/config";
 import { Unreachable, assert, assertEq } from "@app/error";
+import WalletConnect from "@walletconnect/client";
+import * as ethers from "ethers";
+import ModalWalletQRCode from "@app/Components/Modal/QRCode.svelte";
```
cloudhead, line 8: We don't want to be importing UI components in a non-ui typescript module. These should be separate. This shouldn't be needed once you switch to our existing modal system anyway.

src/components/ModalOverlay.svelte
```diff
@@ -0,0 +1,58 @@
...
+    <div class="overlay" on:click={clickOutside} />
+    <div class="content">
+      {#if store !== null}
+        <svelte:component
+          this={store.modalComponent}
+          {...store.modalComponentProps} />
+      {/if}
+    </div>
+  </div>
+
```
cloudhead, line 58: Files should have a newline at the end to be unix-compatible, check https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

src/session.ts
Khay-EMMA: Yeah. noted

src/components/ModalOverlay.svelte
Khay-EMMA: Yeah. this file will become useless tho. I'm already refactoring the modal approach for the QRCode

src/components/ModalOverlay.svelte
Khay-EMMA: Done!

package.json
```diff
@@ -32,11 +33,21 @@
...
     "buffer": "^6.0.3",
+    "@walletconnect/client": "^1.4.2-rc.2",
+    "@walletconnect/qrcode-modal": "^1.4.1",
+    "@ethersproject/abstract-provider": "^5.4.0",
+    "@walletconnect/signer-connection": "^1.4.2-rc.2",
+    "@walletconnect/types": "^1.4.2-rc.2",
+    "@walletconnect/utils": "^1.4.2-rc.2",
     "ethereum-blockies": "^0.1.1",
+    "eventemitter3": "4.0.7",
+    "events": "^3.3.0",
```
cloudhead, line 44: Do we need both event modules? Surely we should be able to get away with only one?

src/App.svelte
```diff
@@ -45,7 +44,6 @@
     height: 100%;
   }
 </style>
-
```
cloudhead, line 48: Please leave these new lines :)

src/Connect.svelte
```diff
@@ -10,19 +10,26 @@
...
-  on:click={() => state.connect(config)}
+  on:click={onClickConnect}
   {style}
   class="connect {className}"
   disabled={connecting || walletUnavailable}
   data-waiting={connecting || null}
 >
+{#if modalConnected}
+<ModalWalletQRCode {config} uri={$store.modalProps?.uri || ''} />
+{/if}
```
cloudhead, line 32: Indentation

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,72 @@
...
+      <div class="qrcode-wrapper">
+      <div data-cy="qr-code">
+        {@html svgString}
+      </div>
+    </div>
+    <p>Or Connect your metamask wallet</p>
+    <button on:click={onClickConnect} style="background-color: white; border-radius: 1rem; height:20;">
+      <div style="display: flex; align-items: center; justify-content: center;">
+     
+      <Metamask/> <p style="margin-left: 0.5rem; color: var(--color-secondary); font-weight: bold;">{caption}</p>
```
cloudhead, line 63: Let's just have a `<a class="link">` here that says `Connect with Metamask` instead of the button/logo

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,72 @@
...
+    
+
```
cloudhead, line 65:
```suggestion
```

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,72 @@
...
+    <div slot="body">
+      <div class="qrcode-wrapper">
+      <div data-cy="qr-code">
+        {@html svgString}
+      </div>
+    </div>
+    <p>Or Connect your metamask wallet</p>
+    <button on:click={onClickConnect} style="background-color: white; border-radius: 1rem; height:20;">
+      <div style="display: flex; align-items: center; justify-content: center;">
+
```
cloudhead, line 62:
```suggestion
```

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,72 @@
...
+      <Metamask/> <p style="margin-left: 0.5rem; color: var(--color-secondary); font-weight: bold;">{caption}</p>
+    
+    
+    </div>
+    </button>
+    </div>
+  </Modal>
+  </div>
+  
+
```
cloudhead, line 72:
```suggestion
```

src/config.ts
```diff
@@ -89,31 +85,89 @@ export class Config {
...
+  let cfg: Config;
+
+  if (isWalletConnectConnected()) {
+    const signer = new WalletConnectSigner(walletConnect, provider, disconnect);
+
+    cfg = new Config(network, provider, signer);
+  } else if (isMetamaskInstalled()) {
+    // If we have Metamask, use it as the signer, but try to use Alchemy
+    // as the provider.
+
```
cloudhead, line 163:
```suggestion
```

src/config.ts
```diff
@@ -89,31 +85,89 @@ export class Config {
...
-    const network = await metamask.ready;
-    config = new Config(network, metamask, metamask.getSigner());
+  let network = { name: "homestead", chainId: 1 };
+
+  const metamask = isMetamaskInstalled() ? new ethers.providers.Web3Provider(window.ethereum) : null;
+
+  if (metamask) {
+    network = await metamask.ready;
+  }
+
```
cloudhead, line 121:
```suggestion
```

src/session.ts
```diff
@@ -28,29 +36,136 @@ export interface Session {
...
+
+      config = new Config(network, provider, signer);
+
+      window.localStorage.setItem("signer", signer);
+
+      store.set({ connection: Connection.Connected, session });
+
+      saveSession({ ...session, tokenBalance: null });
+
+      window.location.reload();
```
cloudhead, line 151: The vertical spacing in this block seems random

src/session.ts
```diff
@@ -28,29 +36,136 @@ export interface Session {
...
+
+    } catch (e) {
+      console.log(e);
+      assertEq(state.connection, Connection.Disconnected);
+      store.set({ connection: Connection.Disconnected });
+      assert(e, "Could not connect to wallet connect");
+    }
+  // store.set({ connection: Connection.Connecting });
```
cloudhead, line 159:
```suggestion
```

src/components/Modal/QRCode.svelte
cloudhead: This file still has no newline at the end..

src/config.ts
```diff
@@ -3,15 +3,22 @@ import type { TypedDataSigner } from '@ethersproject/abstract-signer';
...
     ethereum: any;
     registrarState: any;
   }
 }
+/// Gas limits for various transactions.
+const gasLimits = {
+  createOrg: 1_200_000,
+};
 
+let walletConnect: WalletConnect;
```
cloudhead, line 21: I would avoid global mutable state. This should go inside the `Config` class

src/config.ts
```diff
@@ -89,31 +85,89 @@ export class Config {
...
+    let provider;
+
+    if (network.name === "homestead" && import.meta.env.PROD) {
+      provider = new ethers.providers.AlchemyProvider(network.name, config.alchemy.key);
+    } else if (isMetamaskInstalled()) {
+      provider = new ethers.providers.Web3Provider(window.ethereum);
+    }
+    if (! provider) {
+      throw `No Web3 provider available.`;
+    }
```
cloudhead, line 139: This function can be cleaned up by removing `let provider` and just having an `else` clause which throws.

src/session.ts
```diff
@@ -28,29 +36,136 @@ export interface Session {
   tx: TxState;
 }
 
+export type ModalState = { status: ModalStateType.Open; modalProps: ModalProps } | { status: ModalStateType.Close; modalProps: null };
+
+export interface ModalProps {
+  config: Config;
+  uri: string;
+}
```
cloudhead, line 44: These types and vars are specifically used by the connection modal, so they should be named as such, eg. `ConnectionModal`, `ConnectionModalProps` etc.

src/session.ts
```diff
@@ -28,29 +36,136 @@ export interface Session {
...
+          modalStore.set({ status: ModalStateType.Open, modalProps: { uri, config } });
+          console.log(_opts);
+          onClose();
+        },
+        close: () => {
+          modalStore.set({ status: ModalStateType.Close, modalProps: null });
+        },
+      },
+    });
+  }
```
cloudhead, line 85: This function is duplicated -- it's also in config.ts

src/session.ts
```diff
@@ -28,29 +36,136 @@ export interface Session {
...
+
 export const loadState = (initial: State): Store => {
+
+
   const store = writable<State>(initial);
   const session = window.localStorage.getItem("session");
+  function newWalletConnect(config: Config): WalletConnect {
+
+    return new WalletConnect({
+      bridge: "https://radicle.bridge.walletconnect.org",
```
cloudhead, line 73: This should be in config.json

cloudhead (changes requested):
Good progress, still some work to do.

src/config.ts
Khay-EMMA: throws an error if i put it inside the config class

cloudhead: As a class attribute I don't see why it would cause problems..

cloudhead:
I'm also noticing:

* some files use 4-space indent instead of 2, eg. the QR modal
* lots of trailing whitespace, you should have your editor delete trailing whitespace on save

src/config.ts
```diff
@@ -119,48 +118,33 @@ export async function getConfig(): Promise<Config> {
...
 
-
-  const disconnect = async () => {
-    await walletConnect.killSession().catch(() => {
-      // When the user disconnects wallet-side, calling `killSession`
-      // app-side trows an error because the wallet has already closed
-      // its socket. Therefore, we simply ignore it.
-    });
-  };
-
```
cloudhead: How come we don't need this anymore?

Khay-EMMA: I don't get?

src/Metamask.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts"></script>
```
cloudhead, line 1: This file can be removed now

Khay-EMMA: alright

src/WalletConnectSigner.ts
```diff
@@ -0,0 +1,142 @@
...
+export class WalletConnectSigner extends ethers.Signer {
+  public walletConnect: WalletConnect;
+
+  private _provider: ethers.providers.JsonRpcProvider;
+  constructor(
+    walletConnect: WalletConnect,
+    provider: ethers.providers.JsonRpcProvider,
+  ) {
+    super();
+    defineReadOnly(this, "provider", provider);
```
cloudhead, line 27: You can do `readonly provider` in typescript. We can avoid having the variable set twice.

Khay-EMMA: yeah true!

src/Connect.svelte
```diff
@@ -10,19 +10,26 @@
...
 <button
-  on:click={() => state.connect(config)}
+  on:click={onClickConnect}
   {style}
   class="connect {className}"
   disabled={connecting || walletUnavailable}
   data-waiting={connecting || null}
 >
+  {#if modalConnected}
+    <ModalWalletQRCode {config} uri={$store.modalProps?.uri || ''} />
```
cloudhead, line 31: In what cases can `modalProps` be null? What does it mean for `ModalWalletQRCode` to be initialized without a uri?

src/WalletConnectSigner.ts
```diff
@@ -0,0 +1,142 @@
...
+    const accountAddress = this.walletConnect.accounts[0];
+    if (!accountAddress) {
+      throw new Error(
+        "The connected wallet has no accounts or there is a connection problem"
+      );
+    }
+    return accountAddress;
+  }
+
+
```
cloudhead, line 42:
```suggestion
```

src/WalletConnectSigner.ts
```diff
@@ -0,0 +1,142 @@
...
+  async _signTypedData(domain: TypedDataDomain, types: Record<string, Array<TypedDataField>>, value: Record<string, any>): Promise<string> {
+    // Populate any ENS names (in-place)
+    const populated = await _TypedDataEncoder.resolveNames(domain, types, value, (name: string) => {
+      return this._provider.resolveName(name);
+    });
+
+    const address = await this.getAddress();
+
```
cloudhead, line 50:
```suggestion
```

src/WalletConnectSigner.ts
```diff
@@ -0,0 +1,142 @@
...
+    });
+    return <TransactionResponse>{
+      hash: txHash,
+      nonce: tx.nonce,
+      gasLimit: tx.gasLimit,
+      gasPrice: tx.gasPrice,
+      data: bytesLikeToString(tx.data) || "",
+      value: tx.value,
+      chainId: tx.chainId,
+      confirmations: 1,
```
cloudhead, line 93: Should be `0` no?

src/WalletConnectSigner.ts
```diff
@@ -0,0 +1,142 @@
...
+      gasLimit: maybeBigNumberToString(tx.gasLimit || 200 * 1000),
+      gasPrice: maybeBigNumberToString(tx.gasPrice || 0),
+      nonce,
+      data: bytesLikeToString(tx.data),
+    });
+    return signedTx;
+  }
+
+  connect(_provider: Provider): ethers.Signer {
+    console.log(_provider);
```
cloudhead, line 119:
```suggestion
```

src/base/registrations/registrar.ts
```diff
@@ -213,7 +211,6 @@ async function register(name: string, owner: string, salt: Uint8Array, config: C
   const tx = await registrar(config).connect(config.signer).register(
     name, owner, ethers.BigNumber.from(salt), { gasLimit: 150000 }
   );
-  console.log("Sent", tx);
 
   await tx.wait();
   window.localStorage.clear();
```
cloudhead, line 216: There's no reason to touch this file, it was fine as it was : )

src/components/Modal/QRCode.svelte
cloudhead: Looks like it isn't fixed..

src/base/registrations/registrar.ts
Khay-EMMA: Okay cool

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,61 @@
...
+      margin: var(--content-padding) auto;
+      padding: calc(var(--content-padding) / 2);
+      border-radius: 1rem;
+      background-color: white;
+    }
+    .wrapper {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+  }
```
cloudhead, line 37: Nor here.

src/config.json
```diff
@@ -42,6 +42,7 @@
     "tokens": []
   },
   "alchemy": { "key": "cQFlLK8EokIGlJhd_soImwEyUoC7Ec8r" },
+  "walletConnect" : {"bridge" : "https://radicle.bridge.walletconnect.org"},
```
cloudhead, line 45:
```suggestion
  "walletConnect" : { "bridge" : "https://radicle.bridge.walletconnect.org" },
```

src/config.ts
```diff
@@ -89,31 +87,71 @@ export class Config {
...
+
+}
 
 function isMetamaskInstalled(): boolean {
   const { ethereum } = window;
   return Boolean(ethereum && ethereum.isMetaMask);
 }
 
+function isWalletConnectConnected(): boolean {
+
```
cloudhead, line 100:
```suggestion
```

src/session.ts
```diff
@@ -3,13 +3,21 @@ import type { BigNumber } from 'ethers';
...
 export enum Connection {
   Disconnected,
   Connecting,
   Connected
 }
 
+export enum ModalStateType {
+  Open,
+  Close
+}
```
cloudhead, line 19: You forgot to rename this one

src/session.ts
cloudhead: There's still ModalState and ModalStateType that should be renamed

src/session.ts
```diff
@@ -28,29 +36,117 @@ export interface Session {
...
+
+  // instantiate wallet connect signer
+  let signer: any;
+
+  // Connect to a wallet using walletconnect
+  const connectWalletConnect = async (config: Config) => {
+    walletConnect = newWalletConnect(config);
+
+    provider = new ethers.providers.Web3Provider(window.ethereum);
+
```
cloudhead, line 103:
```suggestion
```

src/session.ts
```diff
@@ -28,29 +36,117 @@ export interface Session {
...
+  //ethereum provider
+  let provider: any;
+
+  // instantiate wallet connect signer
+  let signer: any;
+
+  // Connect to a wallet using walletconnect
+  const connectWalletConnect = async (config: Config) => {
+    walletConnect = newWalletConnect(config);
+
```
cloudhead, line 101:
```suggestion
```

src/session.ts
```diff
@@ -28,29 +36,117 @@ export interface Session {
...
+    store.set({ connection: Connection.Connecting });
+
+    try {
+      await walletConnect.connect();
+
+      const address = await signer.getAddress();
+
+      const tokenBalance: BigNumber = await config.token.balanceOf(address);
+
+      const session = { address, tokenBalance, tx: null };
```
cloudhead, line 120: Why

so


much


space

src/session.ts
```diff
@@ -220,7 +327,7 @@ export async function approveSpender(spender: string, amount: BigNumber, config:
 }
 
 export function disconnectWallet(): void {
-  window.localStorage.removeItem("session");
+  window.localStorage.clear();
```
cloudhead, line 330: wowowow, not a good idea.

cloudhead:
I'm trying the branch locally, and "Connect with Metamask" doesn't seem to work.

cloudhead:
Also getting an error when trying to register a name with wc:

image

cloudhead:
When connecting with a wallet that is set to Rinkeby network, it seems to work anyway by the way, but actually somehow connects you to Mainnet on the UI. Not sure how that works :thinking:

src/Connect.svelte
Khay-EMMA: it can never be initialized with null. I set that because when the modalProps type won't exist on ConnectionModalState when the ConnectionModalStateType is set to close

cloudhead: Hmm I see. Can we add that to the `if` condition instead then, somehow? The QRCode will not work properly with an empty URI, but this code makes it look like it's ok to pass "".

src/WalletConnectSigner.ts
Khay-EMMA: yeah. that's the number confirmation that the provider uses to wait for the transaction before returning the TX reciept

src/components/Modal/QRCode.svelte
Khay-EMMA: what's not supposed to be there?

cloudhead: Indentation not fixed I mean

Khay-EMMA:
> Currently, when you successfully connect your wallet through WalletConnect, the page reloads. This should be avoided!

Resolved

Khay-EMMA:
> When connecting with a wallet that is set to Rinkeby network, it seems to work anyway by the way, but actually somehow connects you to Mainnet on the UI. Not sure how that works 🤔

I think it's using the metamask web3 provider that is available on the browser which might be connected to mainnet. disconnect metamask before connecting to walletConnect

cloudhead:
Hey, most of the comments that were marked as resolved are not resolved. It's better if you let me mark them as resolved once that is the case.

Khay-EMMA:
> Hey, most of the comments that were marked as resolved are not resolved. It's better if you let me mark them as resolved once that is the case.

Can you reopen the comments that are not resolved yet?

src/Connect.svelte
cloudhead: Unresolved :warning:

src/WalletConnectSigner.ts
cloudhead: Unresolved :warning:

src/WalletConnectSigner.ts
cloudhead: Unresolved :warning:

src/components/Modal/QRCode.svelte
cloudhead: Unresolved :warning:

src/config.ts
cloudhead: Unresolved :warning:

src/config.ts
cloudhead: Unresolved :warning:

src/base/registrations/registrar.ts
cloudhead: Unresolved :warning:

src/session.ts
cloudhead: Unresolved :warning:

src/components/Modal/QRCode.svelte
cloudhead: The whole file content is somehow wrongly indented

src/config.json
```diff
@@ -42,6 +42,7 @@
     "tokens": []
   },
   "alchemy": { "key": "cQFlLK8EokIGlJhd_soImwEyUoC7Ec8r" },
+  "radicleBridge" : { "bridge" : "https://radicle.bridge.walletconnect.org" },
```
cloudhead, line 45:
```suggestion
  "walletconnect" : { "bridge" : "https://radicle.bridge.walletconnect.org" },
```

src/config.ts
cloudhead: Still unresolved!

src/session.ts
```diff
@@ -21,37 +29,114 @@ export type State =
...
   | { connection: Connection.Connecting }
   | { connection: Connection.Connected; session: Session };
-
 export interface Session {
   address: string;
   tokenBalance: BigNumber | null; // `null` means it isn't loaded yet.
   tx: TxState;
 }
 
+export type ConnectionModalState = { status: ConnectionModalStateType.Open; modalProps: ConnectionModalProps } | { status: ConnectionModalStateType.Close; modalProps: null };
```
cloudhead, line 38: Line needs wrapping

src/session.ts
```diff
@@ -21,37 +29,114 @@ export type State =
     { connection: Connection.Disconnected }
   | { connection: Connection.Connecting }
   | { connection: Connection.Connected; session: Session };
-
```
cloudhead, line 24: Please leave

src/session.ts
```diff
@@ -21,37 +29,114 @@ export type State =
...
+
+    return walletConnect;
+  }
   if (session) store.set({ connection: Connection.Connected, session: JSON.parse(session) });
 
+  let walletConnect: any;
+  let network: any;
+  //ethereum provider
+  let provider: any;
+  // instantiate wallet connect signer
```
cloudhead, line 86:
```suggestion
```

src/session.ts
```diff
@@ -21,37 +29,114 @@ export type State =
...
+      },
+    });
+
+    return walletConnect;
+  }
   if (session) store.set({ connection: Connection.Connected, session: JSON.parse(session) });
 
+  let walletConnect: any;
+  let network: any;
+  //ethereum provider
```
cloudhead, line 84:
```suggestion
```

src/session.ts
```diff
@@ -64,7 +149,12 @@ export const loadState = (initial: State): Store => {
       const signer = config.signer;
       const address = await signer.getAddress();
 
+      network = config.network;
+
```
cloudhead, line 153:
```suggestion
```

src/session.ts
```diff
@@ -223,7 +317,9 @@ export async function approveSpender(spender: string, amount: BigNumber, config:
 }
 
 export function disconnectWallet(): void {
-  window.localStorage.removeItem("session");
+  window.localStorage.removeItem('session');
+  window.localStorage.removeItem('walletconnect');
+  window.localStorage.removeItem('wallet_connect');
```
cloudhead, line 322: Why are there two keys for walletconnect?

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,59 @@
...
+  .wrapper {
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+  }
+</style>
+
+<div class="wrapper">
+  <Modal floating={true} center>
```
cloudhead, line 40:
```suggestion
  <Modal floating center>
```

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,59 @@
...
+      </a>
+    </p>
+    <div slot="body">
+      <div class="qrcode-wrapper">
+        <div data-cy="qr-code">
+          {@html svgString}
+        </div>
+      </div>
+      <p>Or Connect your metamask wallet</p>
+      <!-- svelte-ignore a11y-missing-attribute -->
```
cloudhead, line 55: What's this?

src/config.json
```diff
@@ -42,6 +42,7 @@
     "tokens": []
   },
   "alchemy": { "key": "cQFlLK8EokIGlJhd_soImwEyUoC7Ec8r" },
+  "walletConnect" : { "radicleBridge" : "https://radicle.bridge.walletconnect.org" },
```
cloudhead, line 45:
```suggestion
  "walletConnect" : { "bridge" : "https://radicle.bridge.walletconnect.org" },
```
Radicle is implied here, and we only have one bridge anyway.

src/session.ts
```diff
@@ -223,7 +314,8 @@ export async function approveSpender(spender: string, amount: BigNumber, config:
 }
 
 export function disconnectWallet(): void {
-  window.localStorage.removeItem("session");
+  window.localStorage.removeItem('session');
+  window.localStorage.removeItem('walletconnect');
```
cloudhead, line 318: Where is this key set?

src/components/Modal/QRCode.svelte
Khay-EMMA: the <a> tag requires a href attribute. putting a href="" will obstruct the connection of metamask. That was why i used a button initially

vite.config.ts
```diff
@@ -5,14 +5,16 @@ import rewriteAll from "vite-plugin-rewrite-all";
 
 const config: UserConfig = {
   optimizeDeps: {
-    exclude: ['svelte-routing']
+    exclude: ['svelte-routing', "@pedrouid/environment","@pedrouid/iso-crypto"]
```
cloudhead, line 8:
```suggestion
    exclude: ['svelte-routing', '@pedrouid/environment', '@pedrouid/iso-crypto']
```

src/session.ts
Khay-EMMA: That's done automatically by walletConnect library

cloudhead: Shouldn't we then use a library function to disconnect? How are we supposed to know that that is the key they use?

src/components/Modal/QRCode.svelte
cloudhead: Gotcha.

src/session.ts
Khay-EMMA: the library function that i used previously is an event that takes in a string called "disconnect" and a callback function to tell the app to do something when it's disconnected which will still result to me clearing the localstorage.

Khay-EMMA: The only item that's being set to the localStorage in the app is the session and walletconnect. so when a user disconnect, it's appropriate to clear the session and the wallet connect instance. That's why i was using the window.localStorage.clear() method

cloudhead:
WalletConnect still doesn't seem to work either:

image

Getting this when trying to create an Org on Rinkeby, with the Rainbow app connected through Rinkeby.

Khay-EMMA:
> WalletConnect still doesn't seem to work either:
> 
> image
> 
> Getting this when trying to create an Org on Rinkeby, with the Rainbow app connected through Rinkeby.
> 
> Additionally, if Metamask is available, it'll try to use Metamask to sign transactions, even though it's connected to the WalletConnect account.

okay. I think when i refreshed the page after connecting, it worked. somehow svelte doesn't update the state until after refresh. That's why i had the window.location.reload() method after connecting to walletConnect

cloudhead:
Yeah, refreshing does it, so we need to find a way around that.

However I get this error after a refresh:

image

The tx still goes through after that.

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  const onClickConnect = () => {
+    state.connectMetamask(config);
+  };
+</script>
+
+<style>
+  .qrcode-wrapper {
+    width: fit-content;
```
cloudhead, line 26:
```suggestion
    width: min-content;
```
Fixes the firefox issue

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,59 @@
...
+    <p slot="title">Connect your wallet</p>
+    <p slot="subtitle" style="text-align: center;">
+      Code Not working? <br />
+      <a href="https://walletconnect.org/wallets" class="link">
+        View compatible wallets.
+      </a>
+    </p>
+    <div slot="body">
+      <div class="qrcode-wrapper">
+        <div data-cy="qr-code">
```
cloudhead, line 50:
```suggestion
        <div>
```

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,59 @@
...
+    flex-direction: column;
+    height: 100%;
+  }
+</style>
+
+<div class="wrapper">
+  <Modal floating center>
+    <p slot="title">Connect your wallet</p>
+    <p slot="subtitle" style="text-align: center;">
+      Code Not working? <br />
```
cloudhead, line 43:
```suggestion
      Code not working? <br />
```

src/components/Modal/QRCode.svelte
```diff
@@ -0,0 +1,59 @@
...
+        View compatible wallets.
+      </a>
+    </p>
+    <div slot="body">
+      <div class="qrcode-wrapper">
+        <div data-cy="qr-code">
+          {@html svgString}
+        </div>
+      </div>
+      <p>Or Connect your metamask wallet</p>
```
cloudhead, line 54:
```suggestion
      <p>Or connect your Metamask wallet</p>
```

src/components/Modal/QRCode.svelte
Khay-EMMA: oh nice

src/session.ts
```diff
@@ -22,36 +30,116 @@ export type State =
...
+      config = new Config(network, provider, signer);
+
+      store.set({ connection: Connection.Connected, session });
+      saveSession({ ...session, tokenBalance: null });
+
+    } catch (e) {
+      console.log(e);
+      assertEq(state.connection, Connection.Disconnected);
+      store.set({ connection: Connection.Disconnected });
+      assert(e, "Could not connect to wallet connect");
```
cloudhead, line 131: What's going on here, these 3 lines of code don't really make sense..

Khay-EMMA: it's line 129 that doesnt make sense.. although I've removed it

Basically, when walletConnect initializes it sets the connection state to connecting, so If there's an error we should set the connection state back to disconnect

cloudhead: Yeah 129 didn't make sense. But 131 is also strange, are you trying to throw an error? `e` is always going to be true here so there's nothing to assert. You should just do `throw new Error("Could not connect to wallet")`

Khay-EMMA: oh yeah. you mentioned the assert method just checks for values and returns true or false

src/session.ts
```diff
@@ -22,36 +30,115 @@ export type State =
...
+      }
+      config = new Config(network, provider, signer);
+
+      store.set({ connection: Connection.Connected, session });
+      saveSession({ ...session, tokenBalance: null });
+
+    } catch (e) {
+      console.log(e);
+      store.set({ connection: Connection.Disconnected });
+      throw e `Could not connect to wallet connect`;
```
cloudhead, line 130: What's going on here? This isn't valid?

vite.config.ts
```diff
@@ -1,26 +1,28 @@
...
-import rewriteAll from "vite-plugin-rewrite-all";
+import path from 'path';
+import { UserConfig } from 'vite';
+import svelte from '@sveltejs/vite-plugin-svelte';
+import rewriteAll from 'vite-plugin-rewrite-all';
 
 const config: UserConfig = {
   optimizeDeps: {
-    exclude: ['svelte-routing']
+    exclude: ['svelte-routing', '@pedrouid/environment','@pedrouid/iso-crypto']
```
cloudhead, line 8:
```suggestion
    exclude: ['svelte-routing', '@pedrouid/environment', '@pedrouid/iso-crypto']
```

src/WalletConnectSigner.ts
Khay-EMMA: This has been resolved

src/WalletConnectSigner.ts
Khay-EMMA: this has been resolved also

src/base/registrations/registrar.ts
Khay-EMMA: resolved

src/session.ts
Khay-EMMA: resolved too

src/config.ts
Khay-EMMA: this has been resolved

src/WalletConnectSigner.ts
cloudhead: Should this be changed then? From the docs:

"The number of blocks that have been mined (including the initial block) since this transaction was mined."

Since no blocks are mined yet, it should be zero..

src/base/registrations/registrar.ts
cloudhead: It's not, just check the diff.

src/config.ts
```diff
@@ -89,31 +89,75 @@ export class Config {
...
   return Boolean(ethereum && ethereum.isMetaMask);
 }
 
+function isWalletConnectConnected(): boolean {
+  const newWalletConnect = (): WalletConnect => {
+    return new WalletConnect({
+      bridge: config.walletConnect.bridge,
+    });
+  };
+  Config.walletConnect = newWalletConnect();
```
cloudhead, line 107: Why don't we just assign the `new WalletConnect` to the var?

src/config.json
```diff
@@ -42,6 +42,7 @@
     "tokens": []
   },
   "alchemy": { "key": "cQFlLK8EokIGlJhd_soImwEyUoC7Ec8r" },
+  "walletConnect" : { "bridge" : "https://radicle.bridge.walletconnect.org" },
```
cloudhead, line 45:
```suggestion
  "walletConnect": { "bridge": "https://radicle.bridge.walletconnect.org" },
```

src/session.ts
```diff
@@ -22,36 +30,117 @@ export type State =
...
-
+  provider: ethers.providers.Provider;
+  signer: ethers.Signer;
+  connectWalletConnect(config: Config): Promise<void>;
   setTxSigning(): void;
   setTxPending(tx: TransactionResponse): void;
   setTxConfirmed(tx: TransactionReceipt): void;
   setChangedAccount(address: string): void;
 }
+let walletConnect: any;
```
cloudhead, line 57: Why is there another walletConnect global here when you have in Config? Why do you need two?

src/session.ts
```diff
@@ -22,36 +30,117 @@ export type State =
...
+        open: (uri: string, onClose, _opts?: unknown) => {
+          modalStore.set({ status: ConnectionModalStateType.Open, modalProps: { uri, config } });
+          console.log(_opts);
+          onClose();
+        },
+        close: () => {
+          modalStore.set({ status: ConnectionModalStateType.Close, modalProps: null });
+        },
+      },
+    });
```
cloudhead, line 79: This instantiation of wallet connect includes settings for the qrcodeModal, while the other in Config doesn't, why is that?

src/session.ts
```diff
@@ -22,36 +30,117 @@ export type State =
...
+
+  // Connect to a wallet using walletconnect
+  const connectWalletConnect = async (config: Config) => {
+    walletConnect = newWalletConnect(config);
+    provider = new ethers.providers.Web3Provider(window.ethereum);
+    signer = new WalletConnectSigner(walletConnect, provider);
+
+    config.setSigner(signer);
+
+    //Todo : check wallet state in the store before attempting to connect
```
cloudhead, line 97:
```suggestion
```

src/WalletConnectSigner.ts
```diff
@@ -0,0 +1,135 @@
...
+import {
+  Deferrable,
+  resolveProperties,
+} from "@ethersproject/properties";
+import { _TypedDataEncoder } from "ethers/lib/utils";
+import type { TypedDataDomain, TypedDataField } from "@ethersproject/abstract-signer";
+
+export class WalletConnectSigner extends ethers.Signer {
+  public walletConnect: WalletConnect;
+  public readonly provider: ethers.providers.JsonRpcProvider;
```
cloudhead, line 19:
```suggestion

  public readonly provider: ethers.providers.JsonRpcProvider;
```

src/session.ts
Khay-EMMA: The one in the config loads when the app renders or reloads and if it has the qrcode modal, it'll popup the qrcode modal every single time

Khay-EMMA: anyways with the global walletconnect that's now a attribute of the class, I don't need the wallet connect instantiation in the config.ts

cloudhead:
Moved to #64

Issue #54, opened 2021-07-19
User profile should show "own" projects


cloudhead:
The user's profile should show projects from all orgs where the user is *the* owner, ie. `org.owner == user.address`.

sebastinez:
77145b1 closes this issue

PR #55, opened 2021-07-19
Add projects to user profile

sebastinez:
This PR works on #54 

It works the same way as the project listing on the Orgs View in: https://github.com/radicle-dev/radicle-interface/blob/b30c5f75ec979dbce4bb98e20a0084e5d082dc51/src/base/orgs/View.svelte#L217

But instead of getting the org by address it gets all the orgs where the user is owner and fetches all projects of each org..

Eventual improvements: We could eventually generate a `getOrgsProjectsByMember` that would work as a wrapper for the former so we don't need to fetch first all the Orgs and then all Projects.

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -216,6 +216,14 @@ export class Org {
     }
   }
 
+  // Return only Orgs that have a specific user as owner
+  static async getOrgsByOwner(owner: string, config: Config): Promise<Org[]> {
+    const orgsResult = await utils.querySubgraph(config.orgs.subgraph, GetOrgsByOwner, { owners: [owner] });
+    return orgsResult.orgs.map((o: { id: string; owner: string }) => {
+      return new Org(o.id, o.owner);
+    });
+  }
+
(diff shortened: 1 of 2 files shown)


cloudhead:
The way this is supposed to work is that we only show projects of orgs where the user is the owner of the org. This disqualifies orgs that have Safes as their owner!

sebastinez:
> The way this is supposed to work is that we only show projects of orgs where the user is the owner of the org. This disqualifies orgs that have Safes as their owner, hence `org.owner == user.address`!

Ah I see, will take Safes into account! 👍

cloudhead:
> > The way this is supposed to work is that we only show projects of orgs where the user is the owner of the org. This disqualifies orgs that have Safes as their owner, hence `org.owner == user.address`!
> 
> Ah I see, will take Safes into account!

Well, that's not really necessary if you just check for orgs with the user as the owner :)

cloudhead:
Code looks good, will try it out!

cloudhead (approved):

cloudhead:
Merged in 77145b13593dcb1314cc83acc4f4d3c135a8297e

Issue #58, opened 2021-07-20
Add form validation to the registration view


cloudhead:
Currently it's possible to save bogus data in ENS, via `/registrations/:label`, eg. you can set a URL as the twitter handle. We should validate inputs on Save.

sebastinez:
Hmm I was thinking about this issue and we could eventually add regex expressions in the initialization of the fields array @ https://github.com/radicle-dev/radicle-interface/blob/8a6c2938a63a3f924d0a3c1de3dc536c3300e525/src/base/registrations/View.svelte#L46

For example:
```diff
  fields = [
-   { name: "owner", placeholder: "", value: r.owner, resolve: true, editable: false },
+  { name: "owner", placeholder: "", value: r.owner, resolve: true, editable: false, validation: "[a-zA-Z0-9]+" },
-   { name: "url", label: "URL", placeholder: "https://acme.org", value: r.url,editable: true }
+  { name: "url", label: "URL", placeholder: "https://acme.org", value: r.url,editable: true, validation: "^(https:|http:|www\.)\S*" }
]
``` 

And then in the `onSave` function in https://github.com/radicle-dev/radicle-interface/blob/8a6c2938a63a3f924d0a3c1de3dc536c3300e525/src/base/registrations/View.svelte#L74 we could iterate over the values and validate the values with `f.value.match(f.validation)`

cloudhead:
That's not a bad idea, though since there are only a few types of fields, what if we defined validation _types_ and had the validation logic inside Form?

Eg. `{ name: "url", validate: "url", ... }`, then in Form we'd check if `validate` is set. If it's undefined, we don't validate the field.

This would keep us from having regexes all over the place. We could have "url", "handle", "address" for now.

In terms of displaying errors we can highlight the field and show the value of the "validate" property like `Invalid input, expected ${validate}`.

sebastinez:
@cloudhead 
> show the value of the "validate" property like Invalid input, expected ${validate}.

Do you mean replacing the typed value with the mentioned text or replacing the tooltip?
IMO I would replace the tooltip to not alter the information entered by the user.

cloudhead:
Yeah the tooltip of course!

sebastinez:
#68 closed this issue

PR #68, opened 2021-08-17
Implement registration form validation

sebastinez:
This PR works on #58

It implements a `check` function that validates the entered value by the user with a predefined Regex pattern enum named `validationTypes` created in the `Form` component which can be one of the following types `address`, `url` or `handle`. Which validation should be used is defined in the `fields` array in the `View` component.

The validation happens on the `keyup` DOM Event in the form and defines an `error` value with the corresponding message `Must be a valid ${field.validate}`. This error message supersedes the `description` message under the input field and once corrected reverts to `undefined` so the `description` message can be shown.
It also disables the submit button until the failed validation is corrected by the user.

On mount of the form a copy of the `field` props is made to be reverted to if a user cancels the input of the entered information

This is a WIP, feel free to provide feedback

**Pending tasks:**
- [x] Improve the regex patterns
- [x] Review the error message style
- [x] When canceling the registration edit, revert the edits

### DIFF
--- a/src/Form.svelte
+++ b/src/Form.svelte
@@ -3,11 +3,28 @@
     name: string;
     value: string | null;
     label?: string;
+    validate?: string;
     placeholder?: string;
     description: string;
     resolve?: boolean;
     editable: boolean;
+    error?: string;
   }
+
+  const validationTypes: { [index: string]: RegExp } = {
+    url: /^(https:\/\/|http:\/\/|ipfs:\/\/)\S+/,
+    // Github
+    //   Username may only contain alphanumeric characters or hyphens.
+    //   Username cannot have multiple consecutive hyphens.
+    //   Username cannot begin or end with a hyphen.
+    //   Maximum is 39 characters.
+    // Twitter
+    //   Username may only contain alphanumeric characters or underscores.
+    //   Maximum is 15 characters.
+    // For simplification of the regex pattern we use a combined version of both requirements.
+    handle: /^[a-zA-Z0-9-_]{1,39}$/,
+    address: /^0x[a-zA-Z0-9]{40}$/
+  };
 </script>
 
 <script lang="ts">
@@ -22,6 +39,25 @@
   export let disabled = false;
   export let config: Config;
 
+  let formFields = fields;
+  let hasErrors = false;
+
+  const check = (event: Event) => {
+    const name = (<HTMLInputElement>event.target).name;
+    const value = (<HTMLInputElement>event.target).value;
+
+    formFields = formFields.map(field => {
+      if (field.name === name && field.validate) {
+        hasErrors = value.length > 0 ? !validationTypes[field.validate].test(value) : false;
+        return { ...field,
+          value: value,
+          error: hasErrors ? `Must be a valid ${field.validate}` : undefined
+        };
+      }
+      return field;
+    });
+  };
+
   const cleanup = (fields: Field[]): { name: string; value: string | null }[] => {
     return fields.filter(field => field.editable).map(field => {
       return {
@@ -31,8 +67,12 @@
     });
   };
   const dispatch = createEventDispatcher();
-  const save = () => dispatch('save', cleanup(fields));
-  const cancel = () => dispatch('cancel');
+  const save = () => dispatch('save', cleanup(formFields));
+  const validate = (event: Event) => dispatch('validate', check(event));
+  const cancel = () => {
+    formFields = fields;
+    dispatch('cancel');
+  };
 </script>
 
 <style>
@@ -72,6 +112,9 @@
     text-overflow: ellipsis;
     border-color: var(--color-secondary) !important;
   }
+  .description.invalid {
+    color: var(--color-negative) !important;
+  }
   input.field::placeholder {
     color: var(--color-secondary);
     font-style: italic;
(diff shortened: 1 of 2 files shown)


src/Form.svelte
```diff
@@ -22,6 +39,22 @@
   export let disabled = false;
   export let config: Config;
 
+  let formFields = fields;
+
+  const check = (e: any) => {
+    formFields = formFields.map(field => {
+      if (field.name === e?.target?.name) {
+        const hasErrors = field.validate && !validationTypes[field.validate].test(e?.target?.value);
+        submitDisabled = hasErrors ? true : false;
```
cloudhead, line 48: Can we just move `hasErrors` to the outer scope, and replace `submitDisabled` completely?

src/Form.svelte
```diff
@@ -3,11 +3,28 @@
...
+    validate?: string;
     placeholder?: string;
     description: string;
     resolve?: boolean;
     editable: boolean;
+    error?: string;
   }
+
+  const validationTypes: { [index: string]: RegExp } = {
+    url: /^(https:|http:|www\.)\S+/,
```
cloudhead, line 15: I wouldn't allow `www.` here as it's not technically valid and we don't want to have issues rendering these. Also make sure you add `\/\/` after the protocol.

src/Form.svelte
```diff
@@ -72,6 +110,9 @@
     text-overflow: ellipsis;
     border-color: var(--color-secondary) !important;
   }
+  .description.invalid {
+    color: red !important;
```
cloudhead, line 114: We have `--var(--color-negative)`

src/Form.svelte
```diff
@@ -22,6 +39,22 @@
...
+  let formFields = fields;
+
+  const check = (e: any) => {
+    formFields = formFields.map(field => {
+      if (field.name === e?.target?.name) {
+        const hasErrors = field.validate && !validationTypes[field.validate].test(e?.target?.value);
+        submitDisabled = hasErrors ? true : false;
+        return { ...field,
+          value: e?.target?.value,
+          error: hasErrors ? `Invalid input, expected ${field.validate}` : undefined
```
cloudhead, line 51: Since the user will get this error before they are done typing, perhaps we should phrase it a bit more softly, eg. `Must be a valid ${field.validate}`

cloudhead:
This is really nice. I think the only thing missing is that it should clear the error when the input is empty.

src/Form.svelte
cloudhead: Perhaps also worth adding `ipfs://` for the future ;)

src/Form.svelte
sebastinez: Yeah, we can add the `field.validate` check to the if statement and remove `submitDisabled`.

src/Form.svelte
sebastinez: Okay no problem, the `\/\/` is already taken care of with `\S` but we can be explicit and show it in the `(https:|http:)` group, like `(https:\/\/|http:\/\/)`.
Great idea with adding `ipfs://` :smiley:

src/Form.svelte
sebastinez: Yeah no problem :+1:

sebastinez:
> This is really nice. I think the only thing missing is that it should clear the error when the input is empty.

Yeah, this was missing, I'm checking now if `field.value` is a value before passing it through the regex test.

src/Form.svelte
cloudhead: The issue with only having `\S` is that `http://` is considered valid :shrug:

src/Form.svelte
```diff
@@ -31,8 +66,13 @@
...
   const dispatch = createEventDispatcher();
-  const save = () => dispatch('save', cleanup(fields));
-  const cancel = () => dispatch('cancel');
+  const save = () => dispatch('save', cleanup(formFields));
+  const validate = (event: Event) => dispatch('validate', check(event));
+  const cancel = () => {
+    formFields = fields;
+    dispatch('cancel');
+  };
+  let hasErrors = false;
```
cloudhead, line 75: Can you move this up so it's defined before it's used? Eg. next to `formFields`

src/Form.svelte
sebastinez: Hmm yeah, there are different regex rules I found that go into a lot of depth declaring all possible domains, but I think we should cover the basics.. Will build something more declarative

sebastinez: It's a interesting topic very plenty of regex have been made, see eg https://mathiasbynens.be/demo/url-regex.

cloudhead: Yeah it's better to be more permissive. Nothing worse than throwing an error on valid input.

cloudhead:
> > This is really nice. I think the only thing missing is that it should clear the error when the input is empty.
> 
> Yeah, this was missing, I'm checking now if `field.value` is a value before passing it through the regex test.

I'm still seeing the same issue :thinking: 

When I delete input with backspace, the error remains.

sebastinez:
> I'm still seeing the same issue 
> 
> When I delete input with backspace, the error remains.

Ok yeah, I think I squashed that now, the error property wasn't being reset properly

cloudhead (approved):
Nice!

PR #80, opened 2021-09-18
Add source information to profile information

sebastinez:
This PR adds the source origin to each property of the profile of each user.

This should allow us to eg. link in case of ENS to the registration form from the user view, and in case of IDX to the self.id app.
I can see that in the future with other identity providers this could help us also in other use cases.

Feel free to provide feedback, did test the PR and run all the checks

### DIFF
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -4,7 +4,6 @@
   import { ethers } from 'ethers';
   import { safeLink, explorerLink, identifyAddress, formatAddress, AddressType, parseEnsLabel } from '@app/utils';
   import { Profile, ProfileType } from '@app/profile';
-  import Loading from '@app/Loading.svelte';
   import Avatar from "@app/Avatar.svelte";
   import type { Config } from '@app/config';
 
@@ -46,9 +45,6 @@
   .address a:hover {
     color: var(--color-foreground);
   }
-  .loading {
-    height: 1.5rem;
-  }
 </style>
 
 <div class="address" title={address} class:no-badge={noBadge}>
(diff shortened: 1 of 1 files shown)


cloudhead:
Hmmm, this works, but I would have thought that a much simpler solution would just be to have a `url()` method on `Profile` that returns the right thing depending on whether it's ENS or IDX. The current code assumes that the name can come from one place and the avatar from a different places, but it's either/or, since we don't merge metadata currently.

sebastinez:
> Hmmm, this works, but I would have thought that a much simpler solution would just be to have a `url()` method on `Profile` that returns the right thing depending on whether it's ENS or IDX. The current code assumes that the name can come from one place and the avatar from a different places, but it's either/or, since we don't merge metadata currently.

Okay, yeah this would be much simpler. I see a benefit for users in loading something thats available on IDX but not on ENS.
But we could start using a `url()` method and see how to merge metadata in the future.

cloudhead:
> > Hmmm, this works, but I would have thought that a much simpler solution would just be to have a `url()` method on `Profile` that returns the right thing depending on whether it's ENS or IDX. The current code assumes that the name can come from one place and the avatar from a different places, but it's either/or, since we don't merge metadata currently.
> 
> Okay, yeah this would be much simpler. I see a benefit for users in loading something thats available on IDX but not on ENS.
> But we could start using a `url()` method and see how to merge metadata in the future.

Yeah, indeed, though maybe also a bit confusing to have a profile patched together from multiple sources :thinking:

sebastinez:
Will open this PR again on a separate branch and with the `url()` method

Issue #74, opened 2021-09-06
TheGraph query failures


cloudhead:
We need to handle query failures by retrying. Note that we get `200 OK` sometimes even though there is a failure.

```
Failed to get entities from store: relation "sgd86.project" does not exist, query = "/* qid: 30079ddb410a63f2-fa18b9efba7cf9d9 */\nselect 'Project' as entity, to_jsonb(c.*) as data from (select  * \n  from \"sgd86\".\"project\" c\n where c.\"block_range\" @> $1 and (\"org\" = $2)\n\n order by \"id\"\n limit 100) c -- binds: [13142350, \"0x8152237402e0f194176154c3a6ea1eb99b611482\"]
```

sebastinez:
@cloudhead I tried to get one failure or a false positive but no luck.

As I understand, we always get the `errors` array from `fetch`, in this case, I would propose to call recursively the `querySubgraph` function until there are no errors or we tried 3 times.
In case that 3 tries aren't enough we can throw a custom error eg. `Failure.GraphQueryFailed` and allow the user to retry manually.

Here a excerpt from `utils.ts`:
```diff
diff --git a/src/utils.ts b/src/utils.ts
index b2898ac..9829def 100644
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -155,6 +155,7 @@ export async function querySubgraph(
   url: string,
   query: string,
   variables: Record<string, any>,
+  intent = 0
 ): Promise<null | any> {
   const response = await fetch(url, {
     method: 'POST',
@@ -172,7 +173,8 @@ export async function querySubgraph(
     for (const e of json.errors) {
       console.error("querySubgraph:", e.message);
     }
-    return null;
+    if (intent < 3) querySubgraph(url, query, variables, intent += 1);
+    else return null;
   }
 
   return json.data;
```

cloudhead:
Yeah I like that solution.

The error is very intermitent. Unless they fixed something... but I haven't been getting the error either. Still, I think it's a good idea to do this, since the node that handles the query is randomly selected.

sebastinez:
PR #83 closes this issue

PR #83, opened 2021-09-21
Retry on Graph query failure

sebastinez:
This PR works on #74

In case that the Graph query for querying orgs and projects, returns with a `errors` array, the `querySubgraph` function retries 4 times in total before it returns null, and a error gets shown where this function fails.

I could only test it by changing the API key to something wrong and received an error `Unknown API key` eventually we'll have to see in the long run which errors can show up.
For that reason I avoided to create a new type of error and forcing it onto the views and components, since for the most part they manage their errors well.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -155,6 +155,7 @@ export async function querySubgraph(
   url: string,
   query: string,
   variables: Record<string, any>,
+  retries = 3
 ): Promise<null | any> {
   const response = await fetch(url, {
     method: 'POST',
@@ -172,7 +173,8 @@ export async function querySubgraph(
     for (const e of json.errors) {
       console.error("querySubgraph:", e.message);
     }
-    return null;
+    if (retries > 0) querySubgraph(url, query, variables, retries - 1);
+    else return null;
   }
 
   return json.data;


src/utils.ts
```diff
@@ -172,7 +173,8 @@ export async function querySubgraph(
     for (const e of json.errors) {
       console.error("querySubgraph:", e.message);
     }
-    return null;
+    if (intent < 3) querySubgraph(url, query, variables, intent += 1);
```
cloudhead, line 176: You can just do `intent + 1` here, no need to mutate it!

Also perhaps we call this `retries` to be clearer?

cloudhead: Actually hmm, it would be nicer if the count was decremented. This would allow the caller to choose the retry count, and you'd default it to 3, ie. `retries = 3` in the function signature.

sebastinez: Yeah sounds even better :+1:

cloudhead (approved):

cloudhead:
Looks good! Can you squash?

cloudhead:
And rebase actually!

cloudhead:
Merged in dad1906

PR #86, opened 2021-09-22
Add registry method to Profile

sebastinez:
This PR is the continuation of #80 

It adds a `registry(config: Config)` to the `Profile` class, which returns either a link to the client registration form or in case of IDX to the https://self.id webapp.

### DIFF
--- a/src/profile.ts
+++ b/src/profile.ts
@@ -1,7 +1,7 @@
 import type { EnsProfile } from "@app/base/registrations/registrar";
 import type { BasicProfile } from "@ceramicstudio/idx-constants";
 import {
-  isAddress, formatCAIP10Address, formatIpfsFile, resolveEnsProfile, resolveIdxProfile, parseUsername,
+  isAddress, formatCAIP10Address, formatIpfsFile, resolveEnsProfile, resolveIdxProfile, parseUsername, parseEnsLabel
 } from "@app/utils";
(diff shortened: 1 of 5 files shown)


sebastinez:
I changed on purpose from `Link` component to the `a` tag since we have to dynamically change between a internal link and a link to the https://self.id page

src/profile.ts
```diff
@@ -98,6 +98,13 @@ export class Profile {
     return config;
   }
 
+  // Returns the corresponding registration form to edit a user profile.
+  // We are not interested in a non-existant registry link, since we check before hand if the name exists.
+  registry(config: Config): string {
+    if (this.profile?.ens) return `/registrations/${parseEnsLabel(this.profile.ens.name, config)}`;
+    else return `https://self.id/${formatCAIP10Address(this.profile.address, "eip155", config.network.chainId)}`;
```
cloudhead, line 105: Could we keep `self.id` in `config.json`? Perhaps under `ceramic.registry`?

cloudhead (approved):

src/profile.ts
sebastinez: Yeah sure :+1:

sebastinez: Done @cloudhead

cloudhead (approved):

Issue #65, opened 2021-08-03
Registration doesn't load after name is registered


cloudhead:
An annoying caching issue: when the registration of a name is completed and you click "View", it sometimes goes to the page an shows "name not registered". This is probably caused by caching, either at the client level (hopefully?) or at the server level. We could retry or have a spinner until it loads, or find a way to disable caching for that request.

sebastinez:
Okay I did some tries and I couldn't reproduce the bug..

It seems to me that ethers.js is not using any cache with its queries, so I can only imagine that it's eventually something from the browser or the Graph.
One thing that comes to mind is when a `getRegistration` call ends up displaying 'name not registered' we could initiate a interval every 2 seconds or so that calls again to check if anything changed..

cloudhead:
Yes I think that's a good solution.

I think the situation is simply racey, and the cache is on the server somewhere, since it loads correctly after a refresh. So retrying a couple of times when coming from the registration flow would be a good option.

cloudhead:
Closed by #88

PR #88, opened 2021-09-22
Refresh registration form

sebastinez:
This PR works on #65 

It checks if the `status` of the view is `NotFound` and in which case it starts recursively to retry fetching the information, until reaching 3 retries.
To avoid refetching the resolver contract it caches it with the first try.

### DIFF
--- a/src/base/registrations/registrar.ts
+++ b/src/base/registrations/registrar.ts
@@ -1,7 +1,7 @@
 import { ethers } from 'ethers';
 import { writable } from 'svelte/store';
 import type { BigNumber } from 'ethers';
-import type { EnsResolver, Resolver } from '@ethersproject/providers';
+import type { EnsResolver } from '@ethersproject/providers';
 import type { TypedDataSigner } from '@ethersproject/abstract-signer';
 import * as session from '@app/session';
 import { Failure } from '@app/error';
@@ -57,7 +57,7 @@ state.subscribe((s: Connection) => {
   console.log("register.state", s);
 });
 
-export async function getRegistration(name: string, config: Config, resolver?: Resolver): Promise<Registration | null> {
+export async function getRegistration(name: string, config: Config, resolver?: EnsResolver): Promise<Registration | null> {
   name = name.toLowerCase();
 
   if (! resolver) {
@@ -96,7 +96,7 @@ export async function getRegistration(name: string, config: Config, resolver?: R
   };
 }
 
-export async function getAvatar(name: string, config: Config, resolver?: Resolver): Promise<string | null> {
+export async function getAvatar(name: string, config: Config, resolver?: EnsResolver): Promise<string | null> {
   name = name.toLowerCase();
 
   resolver = resolver ?? await config.provider.getResolver(name);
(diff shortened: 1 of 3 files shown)


src/base/registrations/registrar.ts
```diff
@@ -1,7 +1,7 @@
 import { ethers } from 'ethers';
 import { writable } from 'svelte/store';
 import type { BigNumber } from 'ethers';
-import type { EnsResolver, Resolver } from '@ethersproject/providers';
```
cloudhead, line 4: What's the difference?

cloudhead (commented):
The only problem I see with this is that it'll happen on all loads, not just when the user registered a name. I think ideally we would only retry in the registration flow, so perhaps the `Done` button when a registration is complete should set some hidden state that triggers the retries. We ideally don't want to pass it as a query param because that will be copied by users :thinking:

src/base/registrations/registrar.ts
sebastinez: There isn't a big difference, but since they are pretty similar we are having places where we use `Resolver` and places where it's `EnsResolver`.
So I took as reference.
https://github.com/radicle-dev/radicle-interface/blob/1fb59faf308d91ad2c77ea35b8a1129e36d838c7/src/base/registrations/registrar.ts#L14
And unified everything under `EnsResolver` since it's name is more explicit.

`EnsResolver` is the interface that Resolver implements and extends.
https://github.com/ethers-io/ethers.js/blob/d3b7130ed6ec50b192eb7f33905eaa327d65eee2/packages/providers/lib/base-provider.d.ts#L31

But I'm seeing now that there a different ethers.js functions that return `Resolver` so eventually we should go back to `Resolver` and unify it everywhere. What do you think?

sebastinez:
> The only problem I see with this is that it'll happen on all loads, not just when the user registered a name. I think ideally we would only retry in the registration flow, so perhaps the `Done` button when a registration is complete should set some hidden state that triggers the retries. We ideally don't want to pass it as a query param because that will be copied by users

Hmm the only way that comes to mind, is using the [Context API](https://svelte.dev/tutorial/context-api) since we have to pass the information in between routes without sharing it in the query  :sweat_smile:

src/base/registrations/registrar.ts
cloudhead: I see, I think using the interface is probably better. It doesn't really matter what the return value is, as it'll match EnsResolver.

cloudhead:
I think there may be a better way actually with the routing library: https://github.com/EmilTholin/svelte-routing#properties-1

See `state`

cloudhead:
https://developer.mozilla.org/en-US/docs/Web/API/History/state

sebastinez:
> I think there may be a better way actually with the routing library: https://github.com/EmilTholin/svelte-routing#properties-1
> 
> See `state`

Okay this works great, I was able to only subtract if the `window.history.state` has an object with a property of `retry` set to `true`.
Now it should only retry 3 times when coming from the registration flow.

cloudhead (approved):

PR #90, opened 2021-09-30
New home page

### LINKED ISSUE #89: Update orgs list to make more compelling as showcase
We currently have a list of organisations on the [Orgs page](https://app.radicle.network/orgs/) that show an avatar and name/address. 

I am planning to focus projects around the output of setting an org and org node being a listing on this page, so it may be worth optimising and making it a little more substantial. Immediate ideas for things I would like to see:

1. Orgs page moved to home page with sidebar moved to top. e.g. like [snapshot](https://snapshot.org/#/)
2. Sorting them to put most compelling ones at the top. 
3. number of repos anchored in each org
4. I see that there is already an issue for adding the ability to star. Would be good to have a star count here as well.

cloudhead:
Closes #89 

image

With blurb:

image

### DIFF
--- a/src/Avatar.svelte
+++ b/src/Avatar.svelte
@@ -47,5 +47,6 @@
 {#if isAddress(source) || isRadicleId(source)}
   <div class="avatar" class:inline bind:this={container} class:glowOnHover title={source}/>
 {:else}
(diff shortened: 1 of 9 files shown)


sebastinez:
Really like the new design :+1:

Issue #91, opened 2021-10-05
Create a fallback image when avatar url returns 404


sebastinez:
If a org or a user links to a image that returns a 404, we get the usual "image not found" rectangle.
I see two ways we could improve that.

**Number One:**
Provide the org or user address as a prop to the avatar, and if the image returns a 404 show a blockies instead of the given url.

**Number Two:**
If the image returns a 404 show a hardcoded 1x1 transparent pixel instead of the missing image.
eg.
```js
let emptyImage = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNgYAAAAAMAASsJTYQAAAAASUVORK5CYII=";
```

To avoid this in the future, we could also enforce the usage of IPFS CIDs, similar to what IDX does.
And use the IPFS gateway to fetch them.

![Screenshot from 2021-10-05 08-30-11](https://user-images.githubusercontent.com/7912302/135988952-5052e9c2-bd12-426f-8a51-41934a5d55ca.png)
![Screenshot from 2021-10-05 08-30-25](https://user-images.githubusercontent.com/7912302/135988981-e08f177b-0b08-44ae-b7e0-2dd6176b706c.png)

sebastinez:
Fixed by #92

PR #92, opened 2021-10-05
Show blockies when avatar image is missing

sebastinez:
This PR works on #91 

It allows images that return `404` or somehow are not able to be loaded into an `img` tag to be replaced by a blockies image.
To simplify the component I reduce it to a simple `img` tag that shows the avatar url or the blockies dataUrl depending on the initial props or the error handling function.

I defined `address` as a required prop, since we never show a avatar without providing an address.

There is still improvements possible.
- [x] Show blockies after a predefined loading time, so we don't wait minutes for a missing image.

Feel free to provide feedback

### DIFF
--- a/src/Avatar.svelte
+++ b/src/Avatar.svelte
@@ -1,25 +1,29 @@
 <script lang="ts">
-  import { onMount } from 'svelte';
   import { createIcon } from '@app/blockies';
   import { isAddress, isRadicleId } from '@app/utils';
 
+  export let address: string;
   export let source: string;
   export let inline = false;
   export let glowOnHover = false;
 
-  let container: HTMLElement;
+  function handleMissingFile() {
+    source = createContainer(address);
+  }
 
-  onMount(() => {
-    if (isAddress(source) || isRadicleId(source)) {
-      const seed = source.toLowerCase();
-      const avatar = createIcon({
-        seed,
-        size: 8,
-        scale: 16,
-      });
-      container.style.backgroundImage = `url(${avatar.toDataURL()})`;
-    }
-  });
+  function createContainer(source: string) {
+    const seed = source.toLowerCase();
+    const avatar = createIcon({
+      seed,
+      size: 8,
+      scale: 16,
+    });
+    return avatar.toDataURL();
+  }
+
+  $: if (isAddress(source) || isRadicleId(source)) {
+    source = createContainer(source);
+  }
 </script>
 
 <style>
@@ -44,9 +48,5 @@
   }
 </style>
 
-{#if isAddress(source) || isRadicleId(source)}
-  <div class="avatar" class:inline bind:this={container} class:glowOnHover title={source}/>
-{:else}
-  <!-- svelte-ignore a11y-missing-attribute -->
-  <img class="avatar" class:inline src={source} class:glowOnHover />
-{/if}
+<!-- svelte-ignore a11y-missing-attribute -->
+<img class="avatar" class:inline src={source} on:error={handleMissingFile} class:glowOnHover />
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -49,7 +49,7 @@
 
 <div class="address" title={address} class:no-badge={noBadge}>
   {#if !noAvatar}
-    <Avatar inline source={profile?.avatar ?? address} />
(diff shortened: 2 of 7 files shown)


sebastinez:
We could eventually add something like this, to allow the interface to show blockies if after 2 seconds no avatar has been able to be loaded.

```diff
diff --git a/src/Avatar.svelte b/src/Avatar.svelte
index 972dcd5..c51bf66 100644
--- a/src/Avatar.svelte
+++ b/src/Avatar.svelte
@@ -1,11 +1,15 @@
 <script lang="ts">
   import { createIcon } from '@app/blockies';
   import { isAddress, isRadicleId } from '@app/utils';
+  import { onMount } from 'svelte';
 
   export let address: string;
   export let source: string;
   export let inline = false;
   export let glowOnHover = false;
+  
+  let timer: any;
+  let isLoaded = false;
 
   function handleMissingFile() {
     source = createContainer(address);
@@ -21,9 +25,16 @@
     return avatar.toDataURL();
   }
 
+  onMount(() => {
+    timer = setTimeout(() => {
+      source = createContainer(address);
+    }, 3000);
+  });
+
 $: if (isAddress(source) || isRadicleId(source)) {
     source = createContainer(source);
   }
+$: if (isLoaded) clearTimeout(timer);
 </script>
 
 <style>
@@ -49,4 +60,4 @@ $: if (isAddress(source) || isRadicleId(source)) {
 </style>
 
 <!-- svelte-ignore a11y-missing-attribute -->
-<img class="avatar" class:inline src={source} on:error={() => handleMissingFile()} class:glowOnHover />
+<img class="avatar" class:inline src={source} on:load={() => isLoaded = true} on:error={() => handleMissingFile()} class:glowOnHover />
```

src/Avatar.svelte
```diff
@@ -1,25 +1,29 @@
...
-        seed,
-        size: 8,
-        scale: 16,
-      });
-      container.style.backgroundImage = `url(${avatar.toDataURL()})`;
-    }
-  });
+$: if (isAddress(source) || isRadicleId(source)) {
+    source = createContainer(source);
+  }
```
cloudhead, line 26: Funky indent

cloudhead:
> We could eventually add something like this, to allow the interface to show blockies if after 2 seconds no avatar has been able to be loaded.
> 
> ```diff
> diff --git a/src/Avatar.svelte b/src/Avatar.svelte
> index 972dcd5..c51bf66 100644
> --- a/src/Avatar.svelte
> +++ b/src/Avatar.svelte
> @@ -1,11 +1,15 @@
>  <script lang="ts">
>    import { createIcon } from '@app/blockies';
>    import { isAddress, isRadicleId } from '@app/utils';
> +  import { onMount } from 'svelte';
>  
>    export let address: string;
>    export let source: string;
>    export let inline = false;
>    export let glowOnHover = false;
> +  
> +  let timer: any;
> +  let isLoaded = false;
>  
>    function handleMissingFile() {
>      source = createContainer(address);
> @@ -21,9 +25,16 @@
>      return avatar.toDataURL();
>    }
>  
> +  onMount(() => {
> +    timer = setTimeout(() => {
> +      source = createContainer(address);
> +    }, 3000);
> +  });
> +
>  $: if (isAddress(source) || isRadicleId(source)) {
>      source = createContainer(source);
>    }
> +$: if (isLoaded) clearTimeout(timer);
>  </script>
>  
>  <style>
> @@ -49,4 +60,4 @@ $: if (isAddress(source) || isRadicleId(source)) {
>  </style>
>  
>  
> - handleMissingFile()} class:glowOnHover />
> + isLoaded = true} on:error={() => handleMissingFile()} class:glowOnHover />
> ```

Yeah not a bad idea, let's keep it for later

src/Avatar.svelte
sebastinez: Yeah this is strange, I tried to indent it like:
```js
$: if (isAddress(source) || isRadicleId(source)) {
  source = createContainer(source);
}
```
But eslint objects with: `Expected indentation of 2 spaces but found 0`, we could disable this eslint rule for this line..

sebastinez: Okay I think I found a nice solution, and eslint not objecting

Issue #99, opened 2021-10-22
Show treasury of orgs in their profiles


sebastinez:
For multi sigs and safes show the amount of ETH and other tokens that are stored on the contract.
Check Alchemy API for checking available tokens under a specific address.

sebastinez:
#102 closed this issue

PR #102, opened 2021-11-01
Add Org treasury to org profile

sebastinez:
This PR implements #99 

It completes the function `utils.getTokens` providing a way to obtain with the Alchemy Token API the top 100 tokens by 24 hour volume.

When an Multi-Sig Org profile is opened and the Org has ETH or some ERC20 top 100 token an additional field called `Treasury` is being shown and a horizontal list of the ETH and other treasured ERC20 tokens is being shown.
In case of a EOA or other address type nothing is shown.

In case that Metamask is being used nothing is shown since the Alchemy API is not being supported.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -31,6 +31,14 @@ export interface SafeTransaction {
     operation: number;
 }
 
+export interface Token {
+  name: string;
+  symbol: string;
+  logo: string;
+  decimals: number;
+  balance: BigNumber;
+}
+
 export async function isReverseRecordSet(address: string, domain: string, config: Config): Promise<boolean> {
   const name = await config.provider.lookupAddress(address);
   return name === domain;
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -380,12 +380,22 @@ export async function getOwnerSafes(owner: string, config: Config): Promise<stri
 
 // Get token balances for an address.
 export async function getTokens(address: string, config: Config):
-  Promise<Array<{ tokenName: string; tokenLogo: string }>>
+  Promise<Array<{ name: string; symbol: string; decimals: number; logo: string; balance: BigNumber }>>
```
cloudhead, line 383: Worth moving that object into its own type.

src/utils.ts
```diff
@@ -380,12 +380,22 @@ export async function getOwnerSafes(owner: string, config: Config): Promise<stri
...
+    // alchemy_getTokenBalances sometimes returns 0x and this does not work well with ethers.BigNumber
+    if (token.tokenBalance !== "0x") {
+      if (!BigNumber.from(token.tokenBalance).isZero()) {
+        return token;
+      }
+    }
+  }).map(async (token: any) => {
+    const tokenMetaData = await config.provider.send("alchemy_getTokenMetadata", [token.contractAddress]);
+    return { ...tokenMetaData, balance: BigNumber.from(token.tokenBalance) };
+  });
```
cloudhead, line 396: Does this fail gracefully if `alchemy_getTokenMetadata` is not supported by the endpoint? ie. just return an empty list.

sebastinez: Yes, if `balances` is just an empty array, the condition `{#if tokens}` in the Orgs View component avoids rendering anything.
And if the endpoint does not support `alchemy_getTokenMetadata` the Promise get's rejected and it gets catched in the `getOrgsTreasury` function and no balance is being shown.

PR #103, opened 2021-11-08
Fix support for new CAIP10 format

sebastinez:
This PR implements a workaround in the `anchorsAccount()` getter to support the caip10 format according to the newest specification, using `<namespace>:<reference>:<address>`


cloudhead (approved):

cloudhead:
Merged in f47ae92

Issue #69, opened 2021-08-18
Testnet Faucet


sebastinez:
I open this issue to track and talk about a possible RAD Faucet on Rinkeby.

- [x] Show faucet in `radicle-interface` app when connected to Rinkeby Testnet

The contract part can be found in https://github.com/radicle-dev/radicle-faucet

cloudhead:
Yeah perhaps we should rate-limit based on ENS names. Eg. each ENS name can only withdraw at most X per day.

sebastinez:
> Yeah perhaps we should rate-limit based on ENS names. Eg. each ENS name can only withdraw at most X per day.

I concur, this would allow use to set a rate-limit without relying on social media, or other potentially centralized services.

sebastinez:
This has been implemented with #104.

PR #104, opened 2021-11-08
Add RAD faucet

sebastinez:
This PR works on #69 

It implements a new route called `/faucet` which allows connected to Rinkeby and a wallet with some Rinkeby ETH.
To withdraw a amount of RAD tokens from the faucet contract.

It verifies on client the following things, to avoid gas unneccesarly spent:
- Don't withdraw zero tokens
- Don't withdraw more than the maximalWithdrawalAmount in one tx.
- If you already withdrawed, wait the requested amount of time.

The verification happens once the button is clicked.

Once passed the verification a new route is being used `/faucet/submit` which uses a modal to inform the user of the state of the transaction and in the end shows the success or eventual errors.
Allows the user to go back to the faucet inital site in the end.

Feel free to provide feedback.


src/base/faucet/Routes.svelte
```diff
@@ -0,0 +1,16 @@
...
+
+  export let config: Config;
+</script>
+
+<Route path="faucet">
+  <Index {config} />
+</Route>
+
+<Route path="faucet/submit">
+  <Submit {config} />
```
cloudhead, line 15: I think `claim` would be a better name for this route/component.

src/base/faucet/Submit.svelte
```diff
@@ -0,0 +1,86 @@
...
+        The amount of {window.history.state.amount.toString()} RAD tokens has been successfully transfered to
+        <span class="highlight">{requester}</span>
+      {:else}
+        <Loading small center />
+      {/if}
+    </span>
+
+    <span slot="actions">
+      {#if state.status === Status.Success}
+        <button on:click={back} class="register"> Back </button>
```
cloudhead, line 82:
```suggestion
        <button on:click={back}> Back </button>
```

src/base/faucet/Routes.svelte
cloudhead: Actually `withdraw`, since that's what you call it in the contract / UI

src/base/faucet/Index.svelte
```diff
@@ -0,0 +1,121 @@
...
+    if (state === true) navigate("/faucet/submit", { state: { amount } });
+    else error = message;
+  }
+
+  async function isAbleToWithdraw(amount: ethers.BigNumber): Promise<[boolean, string?]> {
+    if (amount.isZero()) { return [false, "Not able to withdraw zero tokens"]; }
+    if (parseUnits(amountBN.toString()).gt(maxWithdrawAmount)) return [false, `Reduce amount, max withdrawal is ${formatEther(maxWithdrawAmount)}`];
+    let currentTime = new Date().getTime();
+    let timelock = await calculateTimeLock(amountBN, config);
+    let nextAvailableWithdraw = lastWithdrawal.add(timelock).mul(Math.pow(10,3));
```
cloudhead, line 28: What's the mul by 1000 for?

src/base/faucet/Faucet.ts
```diff
@@ -0,0 +1,56 @@
+import * as ethers from 'ethers';
```
cloudhead, line 1: By convention, capitalized file names are for classes, eg. `class Faucet` would be in `Faucet.ts`. In this case we can just call it `faucet.ts` or `lib.ts`.

src/base/faucet/Faucet.ts
```diff
@@ -0,0 +1,56 @@
...
+export async function calculateTimeLock(amount: ethers.BigNumber, config: Config): Promise<ethers.BigNumber> {
+  assert(config.signer);
+
+  const faucet = new ethers.Contract(
+    config.radToken.faucet,
+    config.abi.faucet,
+    config.signer
+  );
+
+  return faucet.calculateTimeLock(parseUnits(amount.toString()));
```
cloudhead, line 41: What's this toString/parse about?

src/base/faucet/Index.svelte
sebastinez: The timestamp stored in the contract is a 10 digit timestamp based on seconds, while `new Date().getTime()` returns a 13 digit timestamp based on miliseconds.
I can either multiply by 1000 the sum of the lastWIthdrawal and timelock or divide by 1000 the `currentTime`

src/base/faucet/Faucet.ts
sebastinez: The whole thing with passing `toString` and then `parseUnits` is because the user will be providing the value in ether RAD while the values obtained and required by the contract are needed in wei RAD.
I think the easiest way is to already parse everything to wei so we have it ready when needed to send to the contract.

src/base/faucet/Index.svelte
sebastinez: Hmm the other option would be to multiply by 1000 the `block.timestamp` in the faucet contract..

src/base/faucet/Faucet.ts
sebastinez: Done

src/base/faucet/Submit.svelte
sebastinez: Done

src/base/faucet/Routes.svelte
sebastinez: Done

src/base/faucet/Index.svelte
cloudhead: Ok. I'd keep the contract as-is, but maybe add a comment explaining what's going on, and make the modification to `currentTime` instead, ie. divide it by 1000.

src/base/faucet/Faucet.ts
cloudhead: Ah gotcha. Can you maybe create a helper `toWei` that does this, so it's a bit clearer?

src/base/faucet/Index.svelte
sebastinez: In the end I left the multiplication, since JS is quite bad at figuring it out the difference between a 10 digit and a 13 digit timestamp... And if parsed to a string with some of the `Date` methods it shows a date around January 1970..
I added some comments for better understanding

src/base/faucet/Faucet.ts
sebastinez: Working with BigNumbers, made me appreciate strings for big numbers :smile: 
So I changed everything back to string, and I think it made all much more clearer.

cloudhead (approved):

cloudhead:
Good stuff!

PR #108, opened 2021-11-13
Change Gnosis Safe calls to Subgraph

sebastinez:
This PR allows us to reduce the amount of API calls we do to the Gnosis Safe API and use the new introduced Subgraph Schema were we index the safes of a org.

There is a small caveat that we need to query for org owners and org members separately in the `getOrgsByMember` function since we can't do a OR query for one or the other, left a `TODO` there if this gets implemented in the future.

It also moves the console errors filling IDX errors to a much less invasive `console.debug`

This PR depends on the PR on the `radicle-subgraph` repo so we should wait with merging until the subgraph got updated.
https://github.com/radicle-dev/radicle-subgraph/pull/4

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -31,13 +31,38 @@ const GetOrgs = `
   }
 `;
 
-const GetOrgsByOwner = `
-  query GetOrgsByOwner($owners: [String!]!) {
+const GetSafesByOwners = `
+  query GetSafesByOwner($owners: [String!]!) {
+    safes(where: { owners_contains: $owners }) {
+      id
+      owners
+      threshold
+    }
+  }
+`;
+
+const GetOrgsByOwners = `
+  query GetOrgsByOwners($owners: [String!]!) {
     orgs(where: { owner_in: $owners }) {
       id
       owner
+      safe {
+        id
+        owners
+        threshold
+      }
       creator
       timestamp
+    }
+  }
+`;
+
+export const GetSafe = `
+  query GetSafe($addr: ID!) {
+    safe(id: $addr) {
+      id
+      owners
+      threshold
     }
   }
 `;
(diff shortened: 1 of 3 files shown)


src/base/orgs/Org.ts
```diff
@@ -31,9 +31,21 @@ const GetOrgs = `
   }
 `;
 
+const GetOrgsByMembers = `
+  query GetOrgsByOwner($members: [String!]!) {
+    orgs(where: { members: $members }) {
```
cloudhead, line 36: `members_contains: $member`?

src/base/orgs/Org.ts
```diff
@@ -42,6 +54,26 @@ const GetOrgsByOwner = `
...
+      id
+      owners
+      threshold
+    }
+  }
+`;
+
+export const GetSafesByOwner = `
+  query GetSafesByOwner($owner: [String!]!) {
+    safes(where: { owners: $owner }) {
```
cloudhead, line 69: This also won't work

sebastinez:
@cloudhead Implementing the feedback from our talk today, I encountered issues to maintain the data of an org up to date if the safe changes and vice versa..

So I changed the schema, to define that a Safe can have multiple Orgs. IMO we can avoid redundant data and maintain the subgraph store easier up to date. Will look at it tomorrow with fresh eyes but it seems nice.

src/base/orgs/Org.ts
```diff
@@ -31,14 +31,17 @@ const GetOrgs = `
   }
 `;
 
-const GetOrgsByMembers = `
-  query GetOrgsByOwner($members: [String!]!) {
-    orgs(where: { members: $members }) {
+const GetOrgsByMember = `
+  query GetOrgsByMember($owners: [String!]!) {
+    safes(where: { owners_contains: $owners }) {
```
cloudhead, line 36: This is checking whether an array contains another array. Are you sure this is what you want? Shouldn't we query for a single member here?

I'm also wondering why we don't use `members_contains` instead of going through the safe?

sebastinez: We are checking currently if an array is a set of another array, since `_contains` does not accept strings.

cloudhead (approved):

PR #109, opened 2021-11-15
Handle WalletConnect network change

### LINKED ISSUE #66: Handle WalletConnect network change
Currently, the app doesn't respond to changes to the network, in a WalletConnect mobile app that is connected. And if the network doesn't match on connect, an error is thrown.

We should look into handling this differently, at least when Metamask is not present, or if there's a way to switch the Metamask network that's even better.

The issue that makes this not so simple is that with Metamask, you always have a wallet connected, and if the networks don't match, which one takes precedence?

sebastinez:
This PR fixes #66 

## Features
- Starts using Alchemy as Rinkeby provider, leaves Metamask as signer and for dev environment.
- Allows changing accounts and network through WalletConnect on the go, no need to manually refresh the page
  - In case of using Metamask, the UI requests Metamask for a network change, which the user has to confirm (**Does not work with Brave Wallets, only with Metamask**)
- In case of a network mismatch, forces the UI to change to the network of the linked WalletConnect app

## Smaller fixes
- Updates the available wallets link

As always feel free to drop your thoughts :smile:

### DIFF
--- a/src/components/Modal/ConnectWallet.svelte
+++ b/src/components/Modal/ConnectWallet.svelte
@@ -53,7 +53,7 @@
     <div slot="subtitle">
       <div class="text-small">
         WalletConnect<br/>
-        <a href="https://walletconnect.org/wallets" class="link">
+        <a href="https://walletconnect.com/registry/wallets" class="link">
           View compatible wallets
         </a>
       </div>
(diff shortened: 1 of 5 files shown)


sebastinez:
This PR also fixes #107

cloudhead (approved):
LGTM!

PR #110, opened 2021-11-15
Allow users set reverse records

### LINKED ISSUE #42: Allow the user to be able to reverse register its address to the ENS name
Currently the interface goes a long way to help the user to register its subdomain and fill out its profile, but the process of creating the reverse registration of a address to its ENS name does not form part of `radicle-interface` and if one doesn't have  knowledge of ENS can be a unnecessary hurdle.

We could implement this feature since without it the user subdomain will not be resolved and it is a big part of the user identity experience.

sebastinez:
This PR closes #42 

It uses the already working `SetName` component for orgs and implements it for users, to allow a fast reverse record setting after registering their subdomain

### DIFF
--- a/src/base/users/View.svelte
+++ b/src/base/users/View.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import type { SvelteComponent } from 'svelte';
   import type { Config } from '@app/config';
   import Icon from '@app/Icon.svelte';
   import Address from '@app/Address.svelte';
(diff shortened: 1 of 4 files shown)


src/base/users/View.svelte
```diff
@@ -8,9 +9,30 @@
...
+  const setName = () => {
+    setNameForm = SetName;
+  };
+
+  $: isAuthorized = (address: string): boolean => {
+    if ($session) {
+      if (isAddressEqual(address, $session.address)) {
+        return true;
+      }
+    }
```
cloudhead, line 33: `return $session && isAddressEqual(..)`

cloudhead (approved):

cloudhead:
Nice :)

PR #111, opened 2021-11-17
Show orgs in user profile

### LINKED ISSUE #98: Show in user profile which orgs the user is member of.
Show in the user profile the orgs a user is owner as EOA or part of a Gnosis Safe.

## Design:
Frame 1

## Implementation:
I propose to add to the subgraph the ability to return its members:
```diff
type Org @entity {
  id: ID!
  owner: Bytes!
+ members: [Bytes!]! 
  creator: Bytes!
  timestamp: BigInt!
}
```

In case of a EOA org the owner is already part of the subgraph and does not require additional development.
In case of a Gnosis Safe which we could check for with something like: https://github.com/radicle-dev/radicle-interface/blob/b73177ad06e2f48eb067c08eff1e76cc1e46b5ae/src/utils.ts#L338-L345
We should be able to obtain the owners using the following method:
https://github.com/gnosis/safe-contracts/blob/186a21a74b327f17fc41217a927dea7064f74604/contracts/base/OwnerManager.sol#L136
And add the result to the subgraph response.

sebastinez:
This PR closes #98 

It adds a listing equal to the org member listing, to show which orgs a user is member of, and allows
It uses multiple static functions of the Org class to get the information and lets the visitor jump directly to a specific org.

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -32,7 +32,7 @@ const GetOrgs = `
 `;
 
 const GetSafesByOwners = `
-  query GetSafesByOwner($owners: [String!]!) {
+  query GetSafesByOwners($owners: [String!]!) {
     safes(where: { owners_contains: $owners }) {
       id
       owners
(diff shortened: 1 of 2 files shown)


sebastinez:
This PR should be waited for until PR #108 is merged, so I can cleanup the commit history

cloudhead (approved):

Issue #12, opened 2021-06-08
Unit and E2E tests


sebastinez:
To be able to guarantee that the interface works as expected in between changes and fixes.

I propose to implement unit tests for some individual atomic functions and E2E tests for session management and transaction handling, etc.
Regarding packages I suggest the unit testing we could use Jest and the E2E I think Cypress would be great fit, since it's also used by the radicle-upstream repo.

I'm aware that the interface will be more focused on developers but a thoroughly tested product is always something good.

PR #112, opened 2021-11-18
Add Cypress for E2E testing

sebastinez:
This is a PR working on #12 

Its a first draft, to start talking about which integration tests should be part of a first PR.

Open issues:

- [ ] Some of the loading takes forever and can even trigger the current timeout of 20 seconds, we need some stub responses.
- [ ] The implementation of a mock `JsonRPCProvider` could also help

Tests to implement:
- [x] Check project widget to render meta info correctly

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -5,8 +5,12 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - name: Checkout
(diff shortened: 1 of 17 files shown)


sebastinez:
@cloudhead I had to do some changes to the Seed Class and due to that to the Profile methods.
It seems that the Typescript compiler has been awoken :sweat_smile:

sebastinez:
@cloudhead Is there the possibility to allow the Radicle Alchemy Key Mainnet key to be used for testing?
The current key does not work from localhost and the Rinkeby key does not work because of an out of date subgraph..

PR #114, opened 2021-11-19
Allow overriding eth.radicle.seed.host record

cloudhead:
Adds two new records:

* eth.radicle.seed.api
* eth.radicle.seed.git

When specified, these override `eth.radicle.seed.host` for Git
and the HTTP API, respectively.

### DIFF
--- a/src/base/registrations/registrar.ts
+++ b/src/base/registrations/registrar.ts
@@ -6,7 +6,7 @@ import type { TypedDataSigner } from '@ethersproject/abstract-signer';
 import * as session from '@app/session';
 import { Failure } from '@app/error';
 import type { Config } from '@app/config';
-import { unixTime } from '@app/utils';
+import { isDomain, unixTime } from '@app/utils';
 import { assert } from '@app/error';
 
 export interface Registration {
@@ -14,17 +14,38 @@ export interface Registration {
   resolver: EnsResolver;
 }
 
+export class Seed {
+  id?: string;
+  host?: string;
+  git?: string;
+  api?: string;
+
+  constructor(id?: string, host?: string, git?: string, api?: string) {
+    if (id && /^[a-z0-9]+$/.test(id)) {
+      this.id = id;
+    }
+    if (host && isDomain(host)) {
+      this.host = host;
+    }
+    if (api && isDomain(api)) {
+      this.api = api;
+    }
+    if (git && isDomain(git)) {
+      this.git = git;
+    }
+  }
+}
+
 export interface EnsProfile {
   name: string;
   owner?: string;
-  address: string | null;
-  seedId: string | null;
-  seedHost: string | null;
-  anchorsAccount: string | null;
-  url: string | null;
-  avatar: string | null;
-  twitter: string | null;
-  github: string | null;
+  address?: string;
+  seed: Seed;
+  anchorsAccount?: string;
+  url?: string;
+  avatar?: string;
+  twitter?: string;
+  github?: string;
 }
 
 export enum State {
@@ -75,13 +96,15 @@ export async function getRegistration(name: string, config: Config, resolver?: E
     resolver.getText('url'),
     resolver.getText('eth.radicle.seed.id'),
     resolver.getText('eth.radicle.seed.host'),
+    resolver.getText('eth.radicle.seed.git'),
+    resolver.getText('eth.radicle.seed.api'),
     resolver.getText('eth.radicle.anchors'),
     resolver.getText('com.twitter'),
     resolver.getText('com.github'),
   ]);
 
-  const [address, avatar, url, seedId, seedHost, anchorsAccount, twitter, github] =
-    meta.map(r => r.status == "fulfilled" ? r.value : null);
+  const [address, avatar, url, seedId, seedHost, seedGit, seedApi, anchorsAccount, twitter, github] =
+    meta.map(r => r.status == "fulfilled" && r.value ? r.value : undefined);
 
   return {
     resolver,
@@ -89,8 +112,12 @@ export async function getRegistration(name: string, config: Config, resolver?: E
       name,
       url,
       avatar,
-      seedId,
-      seedHost,
+      seed: new Seed(
+        seedId,
+        seedHost,
+        seedGit,
+        seedApi,
+      ),
       anchorsAccount,
       address,
       twitter,
@@ -109,26 +136,6 @@ export async function getAvatar(name: string, config: Config, resolver?: EnsReso
   return resolver.getText('avatar');
 }
 
-export async function getSeedHost(name: string, config: Config, resolver?: EnsResolver | null): Promise<string | null> {
-  name = name.toLowerCase();
-
-  resolver = resolver ?? await config.provider.getResolver(name);
-  if (! resolver) {
-    return null;
-  }
-  return resolver.getText('eth.radicle.seed.host');
-}
-
-export async function getSeedId(name: string, config: Config, resolver?: EnsResolver | null): Promise<string | null> {
-  name = name.toLowerCase();
-
-  resolver = resolver ?? await config.provider.getResolver(name);
-  if (! resolver) {
-    return null;
-  }
-  return resolver.getText('eth.radicle.seed.id');
-}
-
 export async function getAnchorsAccount(name: string, config: Config, resolver?: EnsResolver | null): Promise<string | null> {
   name = name.toLowerCase();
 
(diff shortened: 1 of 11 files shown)


src/profile.ts
```diff
@@ -145,13 +141,13 @@ export class Profile {
         const idx = await resolveIdxProfile(
           formatCAIP10Address(address, "eip155", config.network.chainId), config
         );
-        return { address, ens: null, idx };
+        return { address, idx: idx || undefined };
```
sebastinez, line 144: I think this is a good place to use the nullish coalescing operator, since we are checking for nullish values.

```suggestion
        return { address, idx: idx ?? undefined };
```

src/config.ts
```diff
@@ -138,14 +138,20 @@ export class Config {
...
     Object.assign(cfg, this);
-    cfg.seed.host = seedHost;
 
-    if (seedId) {
-      cfg.seed.id = seedId;
-    }
+    // The `git` and `api` keys being more specific take
+    // precedence over the `host`, if available.
+    const api = seed.api || seed.host;
+    const git = seed.git || seed.host;
```
sebastinez, line 148: I think this is a good place to use the nullish coalescing operator, since we are checking for nullish values.

```suggestion
    const api = seed.api ?? seed.host;
    const git = seed.git ?? seed.host;
```

sebastinez (approved):
LGTM, did just comment two details of using nullish coalescing operator instead of `||`. It works fine like it is, but I think while `??` checks for `null` or `undefined`, `||` has eventually a to wide scope

src/base/registrations/registrar.ts
```diff
@@ -75,22 +81,28 @@ export async function getRegistration(name: string, config: Config, resolver?: E
...
+    resolver.getText('eth.radicle.seed.api'),
     resolver.getText('eth.radicle.anchors'),
     resolver.getText('com.twitter'),
     resolver.getText('com.github'),
   ]);
 
-  const [address, avatar, url, seedId, seedHost, anchorsAccount, twitter, github] =
-    meta.map(r => r.status == "fulfilled" ? r.value : null);
+  const [address, avatar, url, seedId, seedHost, seedGit, seedApi, anchorsAccount, twitter, github] =
+    meta.map(r => r.status == "fulfilled" ? r.value : undefined);
```
sebastinez, line 92: I think we have an issue here since `r.value` can also be `null`, since `getText` returns `string | null`. Eventually we could do something like 
```suggestion
    meta.map(r => r.status == "fulfilled" ? r.value ?? undefined : undefined);
```

src/base/registrations/registrar.ts
```diff
@@ -99,44 +111,40 @@ export async function getRegistration(name: string, config: Config, resolver?: E
   };
 }
 
-export async function getAvatar(name: string, config: Config, resolver?: EnsResolver): Promise<string | null> {
+export async function getAvatar(name: string, config: Config, resolver?: EnsResolver): Promise<string | undefined> {
```
sebastinez, line 114: Here we also will have an issue since `resolver.getText('avatar')` returns `Promise<string | null>` we return `null` and handle it later other await the promise and return `undefined`?

src/base/registrations/registrar.ts
```diff
@@ -99,44 +111,40 @@ export async function getRegistration(name: string, config: Config, resolver?: E
...
   resolver = resolver ?? await config.provider.getResolver(name);
   if (! resolver) {
-    return null;
+    return;
   }
   return resolver.getText('avatar');
 }
 
-export async function getSeedHost(name: string, config: Config, resolver?: EnsResolver): Promise<string | null> {
+export async function getAnchorsAccount(name: string, config: Config, resolver?: EnsResolver): Promise<string | undefined> {
```
sebastinez, line 124: Here we also will have an issue since `resolver.getText('anchors')` returns `Promise<string | null>`

src/base/registrations/registrar.ts
```diff
@@ -99,44 +111,40 @@ export async function getRegistration(name: string, config: Config, resolver?: E
...
+    return;
   }
-  return resolver.getText('eth.radicle.seed.host');
+  return resolver.getText('eth.radicle.anchors');
 }
 
-export async function getSeedId(name: string, config: Config, resolver?: EnsResolver): Promise<string | null> {
-  name = name.toLowerCase();
-
+export async function getSeed(name: string, config: Config, resolver?: EnsResolver): Promise<Seed | undefined> {
```
sebastinez, line 134: Same here returning `Promise<string | null>`

sebastinez (changes requested):
I found some things that we should review before merging, since it will break the build.
All the functions that use `resolver.getText(something)` return `Promise<string | null>` but we want to return `Promise<string | undefined>`.
IMO it's nicer using `undefined` for various reasons. But to make it work, we would need to await the promises and handle the return null value or eventually create a PR on ethers.js to change this?

cloudhead:
Thanks for the review!

So I think in terms of nullish values, the convention is that if you want to return "nothing", you return `null`, but if you want to take "nothing" as a parameter, you use `param?`, ie. you take `undefined`.

This is nice because it avoids accidently returning nothing when you wanted to return a value.

sebastinez:
> So I think in terms of nullish values, the convention is that if you want to return "nothing", you return `null`, but if you want to take "nothing" as a parameter, you use `param?`, ie. you take `undefined`.

Yes this sounds nice and makes sense, but to comply with the types if you want to return "nothing" ie. `null` you would need to specify that as a return type instead of `undefined`.
IMO we should as you said in PR #115 we should agree on one nil type and use it everywhere instead of managing both.

cloudhead:
I've fixed some of the null/undefined issues. Let me know if I missed anything!

src/base/registrations/registrar.ts
```diff
@@ -109,34 +136,32 @@ export async function getAvatar(name: string, config: Config, resolver?: EnsReso
   return resolver.getText('avatar');
 }
 
-export async function getSeedHost(name: string, config: Config, resolver?: EnsResolver | null): Promise<string | null> {
+export async function getAnchorsAccount(name: string, config: Config, resolver?: EnsResolver | null): Promise<string | null> {
```
sebastinez, line 139: In line with what we talked today, shouldn't we manage optional parameters only with `?` and pass only the required arguments to the functions after checking against `null`?

```suggestion
export async function getAnchorsAccount(name: string, config: Config, resolver?: EnsResolver): Promise<string | null> {
```

src/base/registrations/registrar.ts
```diff
@@ -109,34 +136,32 @@ export async function getAvatar(name: string, config: Config, resolver?: EnsReso
...
   resolver = resolver ?? await config.provider.getResolver(name);
   if (! resolver) {
     return null;
   }
-  return resolver.getText('eth.radicle.seed.host');
+  return resolver.getText('eth.radicle.anchors');
 }
 
-export async function getSeedId(name: string, config: Config, resolver?: EnsResolver | null): Promise<string | null> {
+export async function getSeed(name: string, config: Config, resolver?: EnsResolver | null): Promise<Seed | null> {
```
sebastinez, line 149:
```suggestion
export async function getSeed(name: string, config: Config, resolver?: EnsResolver): Promise<Seed | null> {
```

src/base/registrations/registrar.ts
sebastinez: We would need to create some temporary value for the `config.provider.getResolver(name)` call though, since we won't be able to associate it to `resolver`, ie.

```ts
// using var here to shadow the function param
const newResolver = resolver ?? await config.provider.getResolver(name);
if (! newResolver) {
  return null;
}
```

src/base/registrations/registrar.ts
```diff
@@ -109,34 +136,32 @@ export async function getAvatar(name: string, config: Config, resolver?: EnsReso
...
+    resolver.getText('eth.radicle.seed.git'),
+    resolver.getText('eth.radicle.seed.api'),
+  ]);
 
-  resolver = resolver ?? await config.provider.getResolver(name);
-  if (! resolver) {
-    return null;
-  }
-  return resolver.getText('eth.radicle.anchors');
+  return new Seed(id, host, git, api);
```
sebastinez, line 164: This will be an issue, since `getText` could eventually return a `null` value, but in the creation of the `Seed` class we accept only `string | undefined`

Maybe we could do a map similar to.
```ts
const [host, id, git, api] =
    seedPromise.map(r => r.status == "fulfilled" && r.value ? r.value : undefined);
```

sebastinez (commented):
LGTM, just left some details on managing `null` values

src/base/registrations/registrar.ts
cloudhead: Hmm but what is the issue exactly? It seems like typescript is fine compiling this, and the code will work since we check for truthy-ness only?

cloudhead: I guess you mean when we upgrade to ethers 6? I think we can just deal with that issue when/if the time comes..

src/base/registrations/registrar.ts
cloudhead: Yeah, I wasn't going to do this in this PR, but let me check.

cloudhead: Yeah I'd rather fix this separately.

src/base/registrations/registrar.ts
sebastinez: Yeah I don't know why typescript compiles fine for you, for me it throws a typecheck error, that it can't assign a `string | null` value to `string | undefined`..
But if you want to go ahead, since the CI also doesn't throw errors and the build seems to work, why not :smiley:

sebastinez (approved):
If the typechecker doesn't throw errors I'm okay with it too :sweat_smile:

PR #115, opened 2021-11-20
Improve loading of Cards component

### LINKED ISSUE #113: Improve loading times org list
At the moment it takes around 5 seconds at best for the `/orgs` list to be loaded.
I propose to eventually in a first load show all the orgs with their address and a blockies avatar and once all are loaded we add the avatars and names where available.

I think I can recall that the idea was to show everything already resolved.. IMO sometimes after ~4 seconds a normal user already lost interest.

Here some animations to see the difference in UX
## Current Loading
orgs_current
## Improved Loading
orgs_faster

sebastinez:
This PR closes #113 

It does so by showing a simplified `Card` while loading the `Profile` details like avatar, name, etc.
```svelte
{#await Profile.get(org.name ?? org.address, ProfileType.Minimal, config)}
  <Card profile={{ address: org.address }} {config} path={`/orgs/${org.address}`} />
{:then profile}
  {#if orgMembers[profile.address]?.length}
    <Card {profile} {config} path={`/orgs/${profile.nameOrAddress}`} members={orgMembers[profile.address]} />
  {:else}
    <Card {profile} {config} path={`/orgs/${profile.address}`} />
  {/if}
{/await}
```

It also implements some fixes to `null | undefined` as arguments and return values.
We should eventually conform in the entire codebase to one of the two null types and try to cast them to the corresponding one where needed from third party libraries (looking at ethers.js :eyes: ).

To avoid mismatches between addresses this PR also introduces the usage of lowercase addresses in the codebase and uses checksummed address only for display purposes.

### DIFF
--- a/src/Cards.svelte
+++ b/src/Cards.svelte
@@ -1,23 +1,20 @@
 <script lang="ts">
-  import { ethers } from 'ethers';
   import { onMount } from 'svelte';
-  import type { Org } from '@app/base/orgs/Org';
-  import { Profile } from '@app/profile';
-  import type { Config } from '@app/config';
-  import Loading from '@app/Loading.svelte';
+  import { Profile, ProfileType } from '@app/profile';
   import Card from '@app/Card.svelte';
+  import type { Org } from '@app/base/orgs/Org';
+  import type { Config } from '@app/config';
 
   export let config: Config;
   export let orgs: Org[] = [];
   export let profiles: Profile[] = [];
 
   const orgMembers: Record<string, string[]> = {};
-  const getOrgProfiles = Profile.getMulti(orgs.map(o => o.name ?? o.address), config);
 
   onMount(async () => {
     const promises = orgs.map(org => {
       org.getMembers(config).then(members => {
-        orgMembers[ethers.utils.getAddress(org.address)] = members;
+        orgMembers[org.address] = members;
       });
     });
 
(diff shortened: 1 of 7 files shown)


src/Cards.svelte
```diff
@@ -31,31 +28,26 @@
...
-        <Card {profile} {config} path={`/orgs/${profile.nameOrAddress}`} />
-      {/if}
+    {#each orgs as org}
+      {#await Profile.get(org.name ?? org.address, ProfileType.Minimal, config)}
+        <Card profile={{ address: org.address }} {config} path={`/orgs/${org.address}`} />
+      {:then profile}
+        {#if orgMembers[profile.address]?.length}
+          <Card {profile} {config} path={`/orgs/${profile.nameOrAddress}`} members={orgMembers[profile.address]} />
+        {:else}
+          <Card {profile} {config} path={`/orgs/${profile.address}`} />
```
cloudhead, line 41: How come we're using `address` here instead of `addressOrName`? This will give us an address even for orgs that don't use gnosis safe, no?

src/Cards.svelte
```diff
@@ -31,31 +28,26 @@
...
+        {/if}
+      {/await}
     {/each}
 
     {#each profiles as profile}
-      <Card {profile} {config} path={`/users/${profile.nameOrAddress}`} />
+      <Card {profile} {config} path={`/users/${profile.address}`} />
```
cloudhead, line 47: Why?

src/base/registrations/registrar.ts
```diff
@@ -58,7 +58,7 @@ state.subscribe((s: Connection) => {
   console.log("register.state", s);
 });
 
-export async function getRegistration(name: string, config: Config, resolver?: EnsResolver): Promise<Registration | null> {
+export async function getRegistration(name: string, config: Config, resolver?: EnsResolver | null): Promise<Registration | null> {
```
cloudhead, line 61: We want to allow both `null` and `undefined`?

cloudhead:
Looks good, just wondering about a few things!

src/base/registrations/registrar.ts
sebastinez: The issue here is with ethers.js, on line 65 `config.provider.getResolver(name)` returns `EnsResolver | null` to `undefined` and Typescript doesn't like that.

IMO if we want to go back to only accept `EnsResolver | undefined` we could check on line 65 if `getResolver` returns null and return early in that case.

Something like

```ts
  if (! resolver) {
    const newResolver = await config.provider.getResolver(name);
    if (newResolver !== null) resolver = newResolver;

    if (! resolver) {
      return null;
    }
```

src/Cards.svelte
sebastinez: Yeah this is a leftover, we can safely use `nameOrAddress`

src/Cards.svelte
sebastinez: Yeah you're right, this is a leftover will fix it

src/base/registrations/registrar.ts
cloudhead: Ok let's handle this in a different PR, I think we should slowly move away from `null` as parameters or inside datastructures (it's fine as a return value though).

cloudhead (approved):
Good stuff.

PR #117, opened 2021-11-24
Manage anchor proposals

### LINKED ISSUE #39: Manage anchor proposals
If an anchor has been proposed, but is missing signatures, it should be shown in the org view.

* On `/orgs/:address`, there should be a section that shows proposed anchors
* Only org members should see the proposed anchors
* The page should allow members to add their signature/confirmation to the anchor proposal
* The number of required signatures should be shown
* The members who signed the proposal should be shown

### Notes

* `SafeServiceClient.getPendingTransactions` to get proposals
* `SafeServiceClient.decodeData` to find out if the tx is an anchor proposal
* `SafeServiceClient.confirmTransaction` to sign a proposal
* To execute the proposal, we can use the SafeCoreSDK

sebastinez:
This PR closes #39 

It allows the signing and execution of anchor proposals for gnosis safe orgs.

**Some interesting points:**
- At the moment a pending project requires at least the required confirmations, once received the threshold, we don't accept more confirmations, but allow to execute the safe transaction.
- The owners that signed the proposal, show up as blockies, we could eventually show the profile avatar or show something entirely different, IMO the blockies maintain the design of the rest of the page.

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -7,7 +7,7 @@ import { assert } from '@app/error';
 import * as utils from '@app/utils';
 import type { Safe } from '@app/utils';
 import type { Config } from '@app/config';
-import type { Project } from '@app/project';
+import type { PendingProject, Project } from '@app/project';
 
 const GetProjects = `
   query GetProjects($org: ID!) {
@@ -69,14 +69,16 @@ export const GetSafe = `
 export class Org {
   address: string;
   owner: string;
-  name?: string;
+  name?: string | null;
+  safe?: Safe | null;
 
-  constructor(address: string, owner: string, name?: string) {
+  constructor(address: string, owner: string, name?: string | null, safe?: Safe | null) {
     assert(ethers.utils.isAddress(address), "address must be valid");
 
     this.address = address.toLowerCase(); // Don't store address checksum.
     this.owner = owner;
     this.name = name;
+    this.safe = safe;
   }
 
   async setName(name: string, config: Config): Promise<TransactionResponse> {
@@ -162,7 +164,9 @@ export class Org {
   }
 
   async getMembers(config: Config): Promise<Array<string>> {
-    const safe = await utils.getSafe(this.owner, config);
+    if (this.safe) return this.safe.owners;
+
+    const safe = await this.getSafe(config);
     if (safe) {
       return safe.owners;
     }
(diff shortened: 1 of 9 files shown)


src/base/orgs/Org.ts
```diff
@@ -174,6 +180,40 @@ export class Org {
...
+    const orgAddr = ethers.utils.getAddress(this.address);
+    const response = await config.safe.client.getPendingTransactions(
+      ethers.utils.getAddress(this.owner)
+    );
+    const projects: PendingProject[] = [];
+
+    for (const tx of response.results || []) {
+      if (tx.data && tx.to === orgAddr) {
+        const project = parseAnchorTx(tx.data, config);
+
```
cloudhead, line 195:
```suggestion
```

src/base/orgs/Org.ts
```diff
@@ -336,3 +379,20 @@ export class Org {
     });
   }
 }
+
+export function parseAnchorTx(data: string, config: Config): Project | undefined {
```
cloudhead, line 383: Would prefer to return `null` here, as discussed, to be consistent.

src/base/orgs/View/Anchor.svelte
```diff
@@ -0,0 +1,211 @@
...
+  let state = State.Idle;
+  let action: null | Action = null;
+  const isSigned = project.confirmations.includes(
+    ethers.utils.getAddress(session.address)
+  );
+  const close = () => {
+    action = null;
+    state = State.Idle;
+  };
+  const pending = safe.threshold - project.confirmations.length;
```
cloudhead, line 40: I would check the confirmations is `<` than threshold first. It's possible for there to be more confirmations than the threshold.

cloudhead: Nevermind, you do a `<=` check lower.

src/base/orgs/View/Anchor.svelte
```diff
@@ -0,0 +1,211 @@
...
+    <span slot="actions">
+      {#if state == State.Confirm}
+        <button class="primary" on:click={() => confirmAnchor(project.safeTxHash)}>
+          Confirm
+        </button>
+        <button class="text" on:click={close}>
+          Cancel
+        </button>
+      {:else if state == State.Success}
+        <!-- TODO: See if there is a better way to refresh the state in the upper components. -->
```
cloudhead, line 166: You could use `dispatch` to dispatch an event.

sebastinez: Good idea, this would be better :+1:

cloudhead (approved):
Nice :+1:

PR #120, opened 2021-11-25
Add setOwner safe tx proposal creation

### LINKED ISSUE #35: Trigger Transfer transaction when org owner is Safe
Just like the for `Org.setName`, we should be able to trigger an `Org.setOwner` transaction through the Safe.

sebastinez:
This PR closes #35 

- Implements a new function called `setOwnerMultisig` in `Org.ts`, similar to `setNameMultisig`
- Adds Multisig proposal behaviour to `TransferOwnership.svelte`
- Checks if `session.address` is a member of the shown Org before allowing the transfer of ownership

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -133,6 +133,35 @@ export class Org {
     return org.setOwner(address);
   }
 
+  async setOwnerMultisig(owner: string, config: Config): Promise<void> {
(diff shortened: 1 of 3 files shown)


cloudhead (approved):
Very cool!

cloudhead:
Merged in 6588cca

Issue #119, opened 2021-11-25
Update the Rinkeby Subgraph


sebastinez:
With the Rinkeby Subgraph outdated there are features that do not work, when connected to Rinkeby.

![Screenshot from 2021-11-25 09-51-28](https://user-images.githubusercontent.com/7912302/143409688-0b7b426f-3844-48f3-a635-48acb0a108dc.png)

I propose to update the Rinkeby Subgraph to fix these issues

sebastinez:
The subgraph has been updated with the safes feature.

PR #123, opened 2021-11-30
Add unit & integration testing 

### LINKED ISSUE #12: Unit and E2E tests

sebastinez:
This is a PR that continues the work of #112 and closes #12

Somewhat reduced to define first and foremost the basic structure and some component tests.


sebastinez:
This PR is blocked by the issue #119, since the Subgraph is not returning usable information the tests are failing.

sebastinez:
@cloudhead To fix the Netlify errors we have to update the node version used from 12 to 14.
Since Node is already on LTS v16 I think that makes sense.

package.json
```diff
@@ -5,21 +5,28 @@
...
-    "@types/node": "^15.12.2",
+    "@cypress/code-coverage": "^3.9.12",
+    "@cypress/vite-dev-server": "^2.2.2",
+    "@sveltejs/vite-plugin-svelte": "^1.0.0-next.36",
+    "@testing-library/cypress": "^8.0.2",
+    "@tsconfig/svelte": "^1.0.13",
     "@typescript-eslint/eslint-plugin": "^4.26.1",
     "@typescript-eslint/parser": "^4.26.1",
+    "cypress": "^9.3.1",
+    "cypress-svelte-unit-test": "git+https://github.com/flakolefluk/cypress-svelte-unit-test.git",
```
sebastinez, line 20: This is not the best way to define a package, but since the repo from bahmutov is not yet upgraded to Cypress 7 we need to use this fork.
If the library turns out not maintained anymore, we can eventually create our own fork and maintain it.

src/SeedAddress.svelte
```diff
@@ -1,16 +1,14 @@
...
-  export let config: Config;
   export let id: string;
   export let host: string;
   export let port: number;
 
   let seedCopied = false;
 
   const copySeed = (seedId: string, seedHost: string) => {
-    return () => toClipboard(formatSeedAddress(seedId, seedHost, config)).then(() => {
+    return () => toClipboard(formatSeedAddress(seedId, seedHost, port)).then(() => {
```
sebastinez, line 11: We we're passing the whole`config` object into this really simple component, only to show the port number, so I took the liberty to simplify it by only accepting the port number and updating the `formatSeedAddress` to accept a number instead of `config`.

src/base/projects/BranchSelector.svelte
```diff
@@ -101,14 +102,14 @@
...
+    <div data-cy="branch-name" class="stat branch not-allowed">
       {project.meta.defaultBranch}
     </div>
-    <div class="hash">
+    <div data-cy="hash" class="hash">
       {formatCommit(commit)}
     </div>
   {:else}
-    <div class="hash desktop">
+    <div data-cy="hash" class="hash desktop">
```
sebastinez, line 112: A comment in general for the component test I think the best way is to use this `data-cy` attribute, since classes and tags can change but this attribute we can keep and is not scoped on a specific component, but is also valid for nested components.

src/base/users/View.svelte
```diff
@@ -161,96 +161,96 @@
         </div>
       </div>
     </header>
-      <div class="fields">
-        <!-- Address -->
-        <div class="label">Address</div>
-        <div class="desktop"><Address {config} address={profile.address} /></div>
-        <div class="mobile"><Address compact {config} address={profile.address} /></div>
+    <div class="fields">
```
sebastinez, line 164: Here we had wad seems like a wrong indentation, feels great to find them myself :sweat_smile:

src/base/users/View.svelte
```diff
@@ -161,96 +161,96 @@
...
+              {/each}
+            {:catch err}
+              <Message error>
+                <strong>Error: </strong> failed to load projects: {err.message}.
+              </Message>
+            {/await}
+          {/if}
+        {/await}
+      {/if}
+    </div>
```
sebastinez, line 253: Here we had wad seems like a wrong indentation, feels great to find them myself :sweat_smile:

src/utils.ts
```diff
@@ -111,13 +111,14 @@ export function formatIpfsFile(ipfs: string | undefined): string | undefined {
 }
 
 export function formatHash(hash: string): string {
+  if (hash.length < 10) return hash;
```
sebastinez, line 114: If the `hash` length is lesser than 10 digits it doesn't make much sense to shorten it..
It's not something that should happen, but the edge case that we pass in a already shortened hash would be taken into account here.

src/utils.ts
```diff
@@ -130,7 +131,7 @@ export function formatOrg(input: string, config: Config): string {
 
 export function capitalize(s: string): string {
   if (s === "") return s;
-  return s[0].toUpperCase() + s.substr(1);
```
sebastinez, line 133: `substr` is a command that will be deprecated, and is being replaced for `substring`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

src/utils.ts
```diff
@@ -196,7 +197,7 @@ export function isENSName(input: string, config: Config): boolean {
   return regEx.test(input);
 }
 
-// Check whether the input is an Ethereum address.
+// Check whether the input is an checksummed or all lowercase Ethereum address.
```
sebastinez, line 200: Being a bit more explicit here since, on the format of the Ethereum address

vite.config.ts
```diff
@@ -1,18 +1,30 @@
+/// <reference types="vitest" />
 import path from 'path';
 import { UserConfig } from 'vite';
-import svelte from '@sveltejs/vite-plugin-svelte';
+import { svelte } from '@sveltejs/vite-plugin-svelte';
```
sebastinez, line 4: By updating `@sveltejs/vite-plugin-svelte` to `1.0.0-next.36` the library moved to named exports.

vite.config.ts
```diff
@@ -1,18 +1,30 @@
...
 import rewriteAll from 'vite-plugin-rewrite-all';
+import IstanbulPlugin from 'vite-plugin-istanbul';
```
sebastinez, line 6: Since we'll now have a testing env, it seemed like a little stretch on to add code coverage.
We should discuss on what the requirements should be, what's important to us. Lines, 100% coverage, etc?

cloudhead (approved):
Great work :tada:

PR #128, opened 2021-12-03
Add seed page

sebastinez:
This PR adds a page to explore the projects of a particular seed/org node.

If the seed is on v0.2.0 or above it shows the peer id and the projects that the seed node follows.
If not it shows a text, that some funcionality is limited, and one should upgrade to v0.2.0

It also does the following;
- moves the Seed class to the seed folder
- Adds some null types, to allow passing null as argument eg. into the Seed class 
- Changes to the Widget component, to allow projects that don't have a anchor hash.
- Link from the org page to the new seed page

### DIFF
--- a/src/base/orgs/View/Projects.svelte
+++ b/src/base/orgs/View/Projects.svelte
@@ -7,6 +7,7 @@
   import Message from "@app/Message.svelte";
   import Widget from '@app/base/projects/Widget.svelte';
   import Anchor from './Anchor.svelte';
+  import { formatCommit } from "@app/utils";
 
   export let org: Org;
   export let config: Config;
@@ -47,6 +48,10 @@
       <div class="project">
         {#if "safeTxHash" in project} <!-- Pending project -->
           <Widget {project} org={org.address} {config} faded>
+            <span slot="stateHash">
+              <span class="mobile">commit {formatCommit(project.anchor.stateHash)}</span>
+              <span class="desktop">commit {project.anchor.stateHash}</span>
+            </span>
             <span class="anchor" slot="actions">
               {#if org.safe && $session}
                 <Anchor {project} safe={org.safe} on:success={() => updateRecords()} account={$session.address} {config} />
@@ -55,7 +60,12 @@
             </span>
           </Widget>
         {:else} <!-- Anchored project -->
-          <Widget {project} org={org.address} {config} />
+          <Widget {project} org={org.address} {config}>
+            <span slot="stateHash">
+              <span class="mobile">commit {formatCommit(project.anchor.stateHash)}</span>
+              <span class="desktop">commit {project.anchor.stateHash}</span>
+            </span>
+          </Widget>
         {/if}
       </div>
     {/each}
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,7 +7,7 @@ import { Core } from '@self.id/core';
 import WalletConnect from "@walletconnect/client";
 import config from "@app/config.json";
 import { WalletConnectSigner } from "./WalletConnectSigner";
-import type { Seed } from "@app/base/registrations/registrar";
+import type { Seed } from '@app/base/seeds/Seed';
 
 declare global {
   interface Window {
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -8,6 +8,7 @@
   import Registrations from '@app/base/registrations/Routes.svelte';
   import Orgs from '@app/base/orgs/Routes.svelte';
   import Users from '@app/base/users/Routes.svelte';
+  import Seeds from '@app/base/seeds/Routes.svelte';
   import Faucet from '@app/base/faucet/Routes.svelte';
   import Projects from '@app/base/projects/Routes.svelte';
   import Resolver from '@app/base/resolver/Routes.svelte';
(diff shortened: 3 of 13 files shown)


src/base/orgs/View/Projects.svelte
```diff
@@ -55,7 +56,9 @@
             </span>
           </Widget>
         {:else} <!-- Anchored project -->
-          <Widget {project} org={org.address} {config} />
+          <Widget {project} org={org.address} {config}>
+            <span slot="stateHash">commit {project.anchor.stateHash}</span>
+          </Widget>
```
cloudhead, line 61: What do these changes do?

cloudhead: Ah I see, they allow passing the stateHash from outside.

src/base/seed/Seed.ts
```diff
@@ -0,0 +1,47 @@
...
+    return api.get("peer", {}, this.config);
+  }
+  async getProject(urn: string): Promise<Info> {
+    return getInfo(urn, this.config);
+  }
+  async getProjects(): Promise<Project[]> {
+    const result = await getProjects(this.config);
+    return result.map((project: any) => ({ ...project, id: project.urn }));
+  }
+}
```
cloudhead, line 47: Nice, I like this.

src/base/seed/View.svelte
```diff
@@ -0,0 +1,148 @@
...
+        <div class="projects">
+          {#each projects as project}
+            <div class="project">
+              <Widget {project} {config} />
+            </div>
+          {/each}
+        </div>
+      {/await}
+    {:else}
+    <div class="projects">For seed project listing, update http-api to v0.2.0</div>
```
cloudhead, line 145: Indentation

src/base/seed/View.svelte
```diff
@@ -0,0 +1,148 @@
...
+              Copy ✓
+            {:else}
+              Copy
+            {/if}
+          </button>
+        </div>
+        {/await}
+      {:else}
+      <div class="seed-address">Not supported, update http-api to v0.2.0</div>
+      <div />
```
cloudhead, line 122: Indent

src/base/seed/View.svelte
```diff
@@ -0,0 +1,148 @@
...
+            disabled={seedCopied}
+            on:click={copySeed(peer.id, seed.host)}
+          >
+            {#if seedCopied}
+              Copy ✓
+            {:else}
+              Copy
+            {/if}
+          </button>
+        </div>
```
cloudhead, line 118: Indent

src/config.ts
```diff
@@ -57,9 +57,9 @@ export class Config {
   };
   abi: { [contract: string]: string[] };
   seed: {
-    api: { host?: string; port: number };
-    git: { host?: string; port: number };
-    link: { host?: string; id?: string; port: number };
+    api: { host?: string | null; port: number };
+    git: { host?: string | null; port: number };
+    link: { host?: string | null; id?: string | null; port: number };
```
cloudhead, line 62: How come?

src/base/seed/View.svelte
```diff
@@ -0,0 +1,148 @@
...
+        {/await}
+      {:else}
+      <div class="seed-address">Not supported, update http-api to v0.2.0</div>
```
cloudhead, line 121:
```suggestion
      <div class="seed-address subtle">Not supported, update http-api to v0.2.0</div>
```

src/base/seed/View.svelte
cloudhead, line 145:
```suggestion
    <div class="projects subtle">For seed project listing, update http-api to v0.2.0</div>
```

src/config.ts
cloudhead: Can you point to the code that passes null?

src/base/orgs/View/Projects.svelte
sebastinez: Yes exactly, since the seed node doesn't provide anchors, we should have a way to show Projects without anchors or relation to orgs contracts.
Eventually this could be part of what the seed node provides?

src/config.ts
sebastinez: > Can you point to the code that passes null?

in `base/seed/View.svelte:L11` we have 
`const seed = new Seed(config, null, seedAddress);`
Since we are not passing a seed Id and we don't want to pass explicit `undefined` I had to start adding nulls.
If we don't want to do this, we can:
1. Pass `undefined` as argument for the seed id
2. Pass the seed id as query param eg. `/seed/hyyqpngdoe4x4oto3emfdppbw7sj1pfaghbpmmhz5rqiuqg8uofmeo@seed.alt-clients.radicle.xyz`

sebastinez: > Can you point to the code that passes null?

in `base/seed/View.svelte:L11` we have 
`const seed = new Seed(config, null, seedAddress);`
Since we are not passing a seed Id and we don't want to pass explicit `undefined` I had to start adding nulls.
If we don't want to do this, we can:
1. Pass `undefined` as argument for the seed id
2. Pass the seed id as query param eg. `/seed/hyyqpngdoe4x4oto3emfdppbw7sj1pfaghbpmmhz5rqiuqg8uofmeo@seed.alt-clients.radicle.xyz`

cloudhead: Aha, I think solution 3. is probably the cleanest for now. We may run into this again but we can deal with it then if so. Just don't forget all the other call sites for `new Seed`

sebastinez: Okay will do

Issue #32, opened 2021-06-22
Indicate if a project was previously anchored


cloudhead:
Path: `/orgs/:address/:project/:urn/:commit`

Currently, we either show whether the project hash is anchored, or not. There is no indication whether a project was previously anchored at this commit, if the current hash doesn't match the latest anchor.

What we'd like, is to have a third state which is basically "anchored (stale)", meaning this commit *was* anchored, but a newer anchor has replaced it.

We also want to have a link to the latest anchor, if we're not currently at it, and it's in our history. Eg. let's say we're 3 commits ahead of the latest anchor, and that anchor is in our commit history, then it should say "3 commits ahead of anchor", and clicking on that should go to the anchored commit.

sebastinez:
Okay I think for the most part this issue should be implemented quickly.

The only issue I'm seeing is the following sentence:
> Eg. let's say we're 3 commits ahead of the latest anchor, and that anchor is in our commit history, then it should say "3 commits ahead of anchor"

IMO the http-api is missing information to implement this, is it possible for the org node to pass a listing of all commits to the http-api?

@cloudhead wdyt to start somewhere if we start by showing how many anchors behind a previous anchored commit is?

sebastinez:
Okay here a idea what a first improvement could look like without touching the org-node or http-api.
https://gist.github.com/sebastinez/6cf4b7ddd8063c629f218fb18a12f440

IMO if we could get a list of all commits and do the `3 commits behind master` would be awesome.

cloudhead:
If we can solve part of this issue already, it's an improvement. The "x commits ahead" can be solved later.

PR #129, opened 2021-12-06
Improve anchor feature

sebastinez:
This PR is a improvement for the issue #32 

It queries all anchors of a project and if the user visits a previous anchored commit it shows how many anchors are in between this commit and the latest, and lets the user navigate to the latest anchor.

_Example of display:_
Screenshot from 2021-12-06 09-53-31

Works also on responsive
Screenshot from 2021-12-06 09-54-40

### DIFF
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -2,9 +2,9 @@
   import { navigate } from 'svelte-routing';
   import type { Config } from '@app/config';
   import type { Profile } from '@app/profile';
+  import { ethers } from "ethers";
   import * as proj from '@app/project';
   import Loading from '@app/Loading.svelte';
-  import { Org } from '@app/base/orgs/Org';
   import * as utils from '@app/utils';
 
   import Tree from './Tree.svelte';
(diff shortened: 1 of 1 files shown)


src/base/projects/Browser.svelte
```diff
@@ -353,18 +392,18 @@
...
                 </span>
               {:else}
                 <span class="anchor-widget" on:click={() => navigateBrowser(project.head)}>
-                  <span class="anchor-label" title="{anchors}">anchored 🔒</span>
+                  <span class="anchor-label" title="{anchors}">{getAnchorIndexByCommit(anchor, commit)} anchor behind latest 🔒</span>
                 </span>
               {/if}
             {:else if anchor}
-              <span class="anchor-widget not-anchored" on:click={() => navigateBrowser(anchor)}>
+              <span class="anchor-widget not-anchored" on:click={() => navigateBrowser(anchor[0])}>
```
cloudhead, line 406: How are the anchors sorted? Maybe worth sorting them explicitly in the GQL query.

src/base/projects/Browser.svelte
```diff
@@ -353,18 +392,18 @@
...
-            {#if anchor === commit}
+            {#if anchor?.includes(commit)}
               {#if commit === project.head}
                 <span class="anchor-widget anchor-latest">
                   <span class="anchor-label" title="{anchors}">anchored 🔒</span>
                 </span>
               {:else}
                 <span class="anchor-widget" on:click={() => navigateBrowser(project.head)}>
-                  <span class="anchor-label" title="{anchors}">anchored 🔒</span>
+                  <span class="anchor-label" title="{anchors}">{getAnchorIndexByCommit(anchor, commit)} anchor behind latest 🔒</span>
```
cloudhead, line 402: Let's leave the copy here as "anchored" for now. I need to think a little bit more about how we should differentiate this case.

sebastinez: To avoid misunderstandings, you mean that all anchors should simply show up as `anchored 🔒` ?

src/base/projects/Browser.svelte
```diff
@@ -87,10 +87,49 @@
...
+      .map((anchor: AnchorObject) => utils.formatProjectHash(ethers.utils.arrayify(anchor.multihash)));
+  }
+
+  function sortAnchorsByTimestamp(a: AnchorObject, b: AnchorObject) {
+    if (a.timestamp > b.timestamp) {
+      return -1;
+    } else if (a.timestamp < b.timestamp) {
+      return 1;
+    }
+  }
```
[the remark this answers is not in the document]
sebastinez: You're right, will move that to the subgraph query

src/base/projects/Browser.svelte
cloudhead: Oh, no I mean that if a commit was anchored in the past, ie. `anchors.include(commit)`, it should be marked as `anchored`. Before, this would be `not anchored`.

sebastinez: Hmm ok, maybe we can name it `anchored (stale) 🔒` and eventually give it a different background color, to differentiate a past anchor and the current one?

cloudhead: Ok how about `stale 🔒` and `latest 🔒`? The lock shows that it's anchored, while the text tells you whether it's an old anchor or the latest?

cloudhead: maybe latest has the key as well:  `stale 🔒` and `latest 🔐`

sebastinez: Okay yeah I'm fine with that :+1:

sebastinez: Okay I'm having another issue now with this PR, how do we want to handle the interaction with this button..
- commit == head && latest anchor, it should show anchor(latest) and not lead anywhere
- commit != head && anchor == latest, it should lead to the latest commit (head), current behaviour.
- commit != head && anchor != latest, should it lead to latest anchor or latest head?
- commit != head && !anchor, where should it lead?

cloudhead (approved):

PR #135, opened 2021-12-14
Add peer id to project api routes

sebastinez:
This PR adds the peer id of the seed to the project blob, readme and tree routes.

It breaks the current http-apis and require an update by the maintainers to keep seeing projects in the web client

### DIFF
--- a/src/project.ts
+++ b/src/project.ts
@@ -95,7 +95,7 @@ export async function getTree(
   if (path === "/") {
     path = "";
   }
-  return api.get(`projects/${urn}/tree/${commit}/${path}`, {}, config);
+  return api.get(`projects/${urn}/tree/${config.seed.link.id}/${commit}/${path}`, {}, config);
 }
 
 export async function getBlob(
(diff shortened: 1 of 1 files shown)


src/project.ts
```diff
@@ -95,7 +95,7 @@ export async function getTree(
   if (path === "/") {
     path = "";
   }
-  return api.get(`projects/${urn}/tree/${commit}/${path}`, {}, config);
+  return api.get(`projects/${urn}/tree/${config.seed.link.id}/${commit}/${path}`, {}, config);
```
cloudhead, line 98: I don't think this is going to do it, as this is trying to get the remote identified by the seed id, which most likely has no remotes of its own. By default, we should be using the first maintainer/delegate's peer-id which I think was the previous behavior of the http-api, and is what is returned as `remote` in the project_info endpoint.

sebastinez:
Will close this PR, since we won't use peer ids on endpoints related to source code

PR #136, opened 2021-12-20
Show project commit listing

sebastinez:
This PR implements a commit listing, which can be accessed by clicking on the commit count in the project view.

It also:
- Refactors the `Browser` component, extracting the Header in its own component.
- Adds value binding for `commit` and `content` between `View` and lower components `History` and `Header`
- Moves anchor fetching to `Header` component

- [x] Still working on improvements for responsive..

### DIFF
--- a/src/base/projects/Header.svelte
+++ b/src/base/projects/Header.svelte
@@ -1,0 +1,292 @@
+<script lang="ts">
+  import type { Config } from '@app/config';
+  import * as utils from '@app/utils';
+  import Loading from '@app/Loading.svelte';
+  import { Org } from '@app/base/orgs/Org';
+  import type { Info, Tree } from '@app/project';
+  import type { Profile } from '@app/profile';
+  import { createEventDispatcher } from "svelte";
+  
+  export let config: Config;
+  export let anchors: string | null = null;
+  export let urn: string;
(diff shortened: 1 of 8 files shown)


src/base/projects/Header.svelte
```diff
@@ -323,12 +334,8 @@
       {/if}
     </div>
   </span>
-  <div class="stat commit-count" on:click={switchContent}>
-    {#if content == ProjectContent.Browser}
-      <strong>{tree.stats.commits}</strong> commit(s)
-    {:else}
-      <strong>Back to Browser</strong>
```
sebastinez: @cloudhead I think it's not that clear that one can navigate back by clicking again the commits count.
Should we rely on the browser back button or give the user some other way to do it?

cloudhead: Yeah I think eventually there should be several ways, but for now this is kind of a hidden feature anyway. What I'd like to see is:

1) back button
2) click on project name (will work for free once URL changes when commits is clicked)
3) click on the commits button again (current behavior)

sebastinez: Ok great, yeah I think that makes sense, will provide it with separate PR

PR #139, opened 2021-12-23
Adds content param to project urls

sebastinez:
By adding a content param to the project urls and use the Browser content as default, we are able to control better the url shown in the project views as well in the future implement other views, eg. commit detail view, etc.

With this commit we have full URL support thus allowing navigation with:

- Back button
- Clicking on project name
- Clicking again on commit count

For future improvements we could allow to pass any string as `content` and parse it to an available view to match, and if not found return 404.

**Some thoughts on the why:**
To use `navigate` for pushing history state and avoid a re rendering by svelte-routing, we have to avoid route switching i.e.
If we are on the URL
`/users/<user>/projects/<urn>/<sha>` and switch to `/users/<user>/projects/<urn>/commits/<sha>` we are adding a route param `commits` which makes `svelte-routing` jump between Routes and thus re render. If we define a `content` param which changes its value based on interaction with the component and we pass that to the component or update it inside the component, we can avoid re rendering the entire component.

### DIFF
--- a/src/base/projects/Routes.svelte
+++ b/src/base/projects/Routes.svelte
@@ -2,16 +2,21 @@
   import { Route } from "svelte-routing";
   import View from '@app/base/projects/View.svelte';
   import type { Config } from '@app/config';
+  import { ProjectContent } from "@app/project";
 
   export let config: Config;
+
+  function parseContent(content: string): ProjectContent {
+    return content == "history" ? ProjectContent.History : ProjectContent.Tree;
+  }
 </script>
 
-<Route path="/projects/:urn/head/*" let:params>
-  <View {config} urn={params.urn} path={params['*'] || "/"} />
+<Route path="/projects/:urn/:content/head/*" let:params>
+  <View {config} urn={params.urn} content={parseContent(params.content)} path={params['*'] || "/"} />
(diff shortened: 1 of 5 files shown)


src/base/projects/Routes.svelte
```diff
@@ -2,16 +2,21 @@
...
 
-<Route path="/projects/:urn/head/*" let:params>
-  <View {config} urn={params.urn} path={params['*'] || "/"} />
+<Route path="/projects/:urn/:content/head/*" let:params>
+  <View {config} urn={params.urn} content={parseContent(params.content)} path={params['*'] || "/"} />
 </Route>
 
 <Route path="/projects/:urn/:commit/*" let:params>
-  <View {config} urn={params.urn} commit={params.commit} path={params['*'] || "/"} />
+  <View {config} urn={params.urn} content={parseContent(params.content)} commit={params.commit} path={params['*'] || "/"} />
```
cloudhead, line 19: Something's missing from the path here

sebastinez: Thanks for the catch :+1:

sebastinez:
Closing this PR in favor of a static Route solution

PR #140, opened 2021-12-23
Show project with anchor when no commit hash passed

### LINKED ISSUE #132: When viewing project with org context and no hash, try to use anchor as hash

sebastinez:
This PR closes #132 

- Moves anchor retrieval back into the `View` component.
- Passes resolved anchors into `Header` component, if no anchors available, returns an empty array which is falsy and `Header` won't show any anchor
- If no commit is passed to `View` component, checks first for latest anchor, and else uses head commit.

### DIFF
--- a/src/base/projects/View.svelte
+++ b/src/base/projects/View.svelte
@@ -10,6 +10,7 @@
   import { formatOrg, formatSeedId } from '@app/utils';
   import { getOid } from '@app/project';
   import { Seed } from '@app/base/seeds/Seed';
+  import { getAllAnchors } from '@app/anchors';
 
   import Header from '@app/base/projects/Header.svelte';
   import ProjectContentRoutes from '@app/base/projects/ProjectContentRoutes.svelte';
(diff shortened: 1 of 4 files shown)


src/base/projects/View.svelte
```diff
@@ -74,6 +76,31 @@
...
+  async function getAllAnchors(urn: string, anchors?: string | null): Promise<string[]> {
+    if (! anchors) {
+      return [];
+    }
+    const unpadded = decodeRadicleId(urn);
+    const id = ethers.utils.hexZeroPad(unpadded, 32);
+    const allAnchors = await querySubgraph(config.orgs.subgraph, GetAllAnchors, { project: id, org: anchors });
+    return allAnchors.anchors
+      .map((anchor: AnchorObject) => formatProjectHash(ethers.utils.arrayify(anchor.multihash)));
+  }
```
cloudhead, line 102: Can we perhaps move this in a `lib.ts` perhaps? Seems like it shouldn't belong in a view.

cloudhead:
Needs rebase.

PR #141, opened 2021-12-30
Project url support and branch browsing

sebastinez:
This PR provides:

- Adds URL support for project routes:
  - For commit listing under `/users/:user/projects/:urn/history/:commit`
  - The tree view you'll find under `/users/:user/projects/:urn/tree/:commit`
  - and as fallback we show the tree view
  - Adds branches selector to the `Header` that allows selecting the different head commits wrt to the selected branch

### DIFF
--- a/src/base/projects/BranchSelector.svelte
+++ b/src/base/projects/BranchSelector.svelte
@@ -1,0 +1,139 @@
+<script lang="ts">
+  import { Info, getOid } from "@app/project";
+  import { formatCommit, isOid } from "@app/utils";
+  import { createEventDispatcher } from "svelte";
+
+  export let branches: [string, string][];
+  export let project: Info;
+  export let revision: string;
+
+  const dispatch = createEventDispatcher();
+
+  let branchesDropdown = false;
+  const switchBranch = (name: string) => {
+    dispatch("revisionChanged", name);
+  };
+
+  // Sort branches array alphabetically
+  const sortBranches = ([firstBranchName,]: [string, string], [secondBranchName,]: [string, string]) => {
+    if (firstBranchName < secondBranchName) return -1;
+    if (firstBranchName > secondBranchName) return 1;
+    return 0;
+  };
+
+  branches = branches.sort(sortBranches);
+
+  // Casting commit to string, since the commit will always be defined here
+  $: commit = getOid(project.head, revision, branches);
+</script>
+
+<style>
+  .commit {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    font-family: var(--font-family-monospace);
+  }
+  .commit .branch {
+    cursor: pointer;
+    padding: 0.5rem 0.75rem;
+    color: var(--color-secondary);
+    background-color: var(--color-secondary-background);
+    border-radius: 0.25rem 0 0 0.25rem;
+  }
+  .commit .branch.not-allowed {
+    cursor: not-allowed;
+  }
+  .branch:hover {
+    background-color: var(--color-foreground-background-lighter);
+  }
+  .commit .hash {
+    display: inline-block;
+    color: var(--color-secondary);
+    background-color: var(--color-secondary-background-darker);
+    padding: 0.5rem 0.75rem;
+    border-radius: inherit;
+  }
+  .item {
+    cursor: pointer;
+    padding: 0.3rem;
+  }
+  .item:hover {
+    background-color: var(--color-foreground-background-lighter);
+  }
+  .dropdown {
+    background-color: var(--color-foreground-background);
+    padding: 1rem;
+    margin-top: 0.5rem;
+    border-radius: 0.25rem;
+    display: none;
+    position: absolute;
+  }
+  .branch-dropdown.branch-dropdown-visible {
+    display: block;
+  }
+  .stat {
+    font-family: var(--font-family-monospace);
+    padding: 0.5rem 0.75rem;
+    background: var(--color-foreground-background);
+  }
+  @media (max-width: 720px) {
+    .dropdown {
+      left: 32px;
+      z-index: 10;
+    }
+  }
+</style>
+
+<div class="commit">
+  <!-- Check for branches listing feature -->
+  {#if branches.length > 0}
+    <span>
+      <div on:click={() => branchesDropdown = !branchesDropdown} class="stat branch" class:not-allowed={!branches}>
+        {#if commit === project.head}
+          {project.meta.defaultBranch}
+        <!-- If commit is no sha1 commit show branch or tag name -->
+        {:else if !isOid(revision)}
+          {revision}
+        {:else}
+          Browse...
+        {/if}
+      </div>
+      <div
+        class="dropdown branch-dropdown"
+        class:branch-dropdown-visible={branchesDropdown}
+      >
+        {#each branches as [name,]}
+          <div class="item" on:click={() => switchBranch(name)}>{name}</div>
+        {/each}
+      </div>
+    </span>
+    {#if commit === project.head || !isOid(revision)}
+      <div class="hash">
+        {formatCommit(commit)}
+      </div>
+    {:else}
+      <div class="hash desktop">
+        {commit}
+      </div>
+      <div class="hash mobile">
+        {formatCommit(commit)}
+      </div>
+    {/if}
+  <!-- If there is no branch listing available, show default branch name if commit is head and else show entire commit -->
+  {:else if commit === project.head}
+    <div class="stat branch not-allowed">
+      {project.meta.defaultBranch}
+    </div>
+    <div class="hash">
+      {formatCommit(commit)}
+    </div>
+  {:else}
+    <div class="hash desktop">
+      {commit}
+    </div>
+    <div class="hash mobile">
+      {formatCommit(commit)}
+    </div>
+  {/if}
+</div>
--- a/src/base/projects/Commit/History.svelte
+++ b/src/base/projects/Commit/History.svelte
@@ -1,18 +1,26 @@
 <script lang="ts">
+  import { createEventDispatcher, onMount } from "svelte";
   import CommitTeaser from "./CommitTeaser.svelte";
-  import { getCommits } from "@app/project";
+  import { getCommits, Info, getOid, ProjectContent } from "@app/project";
   import type { Config } from "@app/config";
   import Loading from "@app/Loading.svelte";
   import { groupCommitHistory, GroupedCommitsHistory } from "./lib";
 
-  export let commit: string;
+  export let revision: string;
   export let urn: string;
   export let config: Config;
+  export let project: Info;
+  export let branches: [string, string][];
 
   async function fetchCommits(): Promise<GroupedCommitsHistory> {
-    const commitsQuery = await getCommits(urn, commit, config);
+    const commitsQuery = await getCommits(urn, getOid(project.head, revision, branches), config);
     return groupCommitHistory(commitsQuery);
   }
+
+  const dispatch = createEventDispatcher();
+  onMount(() => {
+    dispatch("routeParamsChange", { content: ProjectContent.History, revision, path: "/" });
+  });
 </script>
 
 <style>
@@ -31,7 +39,7 @@
   .commit {
     padding: 0.25rem 0;
   }
-  @media (max-width: 720px) {
+  @media (max-width: 960px) {
     .history {
       padding-left: 2rem;
     }
--- a/src/base/projects/Header.svelte
+++ b/src/base/projects/Header.svelte
@@ -3,18 +3,24 @@
   import * as utils from '@app/utils';
   import Loading from '@app/Loading.svelte';
   import { ethers } from "ethers";
-  import type { Info, Tree } from '@app/project';
-  import { ProjectContent } from "@app/project";
+  import { ProjectContent, getOid } from '@app/project';
+  import type { Info, Tree } from "@app/project";
   import type { Profile } from '@app/profile';
+  import BranchSelector from './BranchSelector.svelte';
+  import { createEventDispatcher } from 'svelte';
+
+  const dispatch = createEventDispatcher();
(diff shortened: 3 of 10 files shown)


src/base/projects/View.svelte
```diff
@@ -16,11 +16,13 @@
...
   export let org = "";
   export let user = "";
-  export let commit = "";
   export let config: Config;
-  export let path: string;
+  export let peer: string;
+  export let location: any;
 
-  let content = proj.ProjectContent.Browser;
+  let content = location.pathname.indexOf("history") !== -1 ? proj.ProjectContent.History : proj.ProjectContent.Tree;
```
sebastinez, line 23: I think there is a better way to check for the current ProjectView. Will improve

src/base/projects/View.svelte
```diff
@@ -16,11 +16,13 @@
...
+  let path = "/";
+  let commit = "";
```
sebastinez, line 25: Not sure if defining the default commit as empty string is a good idea..

src/base/projects/View.svelte
```diff
@@ -144,9 +151,12 @@
           </a>
           <span class="divider">/</span>
         {/if}
-        <Link to={projectRoot}>{result.project.meta.name}</Link>
+        <a href={projectRoot}>{result.project.meta.name}</a>
```
sebastinez, line 154: This can be reverted to a `Link`

src/base/projects/View.svelte
```diff
@@ -156,16 +166,33 @@
...
       />
-      {#if content == proj.ProjectContent.Browser}
-        <Browser {urn} {org} {user} {path} {tree}
-          commit={commit}
-          config={result.config} />
-      {:else if content == proj.ProjectContent.Commits}
-        <History {urn} config={result.config} bind:commit={commit}  />
-      {/if}
+      <Router>
+        <Route path="tree//*" let:params>
```
sebastinez, line 174:
```suggestion
        <Route path="tree/*" let:params>
```

sebastinez (commented):
Some things I should improve before merging

src/base/projects/View.svelte
sebastinez: Okay I found a more robust way by creating a regex that matches the available paths.

src/base/projects/Commit/History.svelte
```diff
@@ -31,7 +31,7 @@
   .commit {
     padding: 0.25rem 0;
   }
-  @media (max-width: 720px) {
+  @media (max-width: 960px) {
```
sebastinez, line 34: This is a minor style improvement, the commit listing was changing its padding much later than the header and other components.

src/utils.ts
```diff
@@ -165,6 +166,14 @@ export function parseUsername(input: string): string {
   return parts[parts.length - 1];
 }
 
+export function parseProjectURL(pathname: string): { content: ProjectContent; commit: string; path: string } {
```
sebastinez, line 169: Ok this could be a interesting discussion point.. Trying to obtain information from the nested routes inside the `View` component, the easiest and cleanest solution I found was to parse the pathname starting before `tree` or `history` and parse the commit and path that one passes after that, and in case of missing parts to provide fallbacks..
A URL without `tree` or `history` does not exist since the page redirects either to `history` or `tree` as fallback.

src/base/projects/Header.svelte
```diff
@@ -222,23 +198,9 @@
...
-    {:else}
-      <div class="hash desktop">
-        {commit}
-      </div>
-      <div class="hash mobile">
-        {utils.formatCommit(commit)}
-      </div>
-    {/if}
-  </div>
+  <BranchSelector {branches} {project}
```
sebastinez, line 201: The branch selection logic got too complicated to be inlined here, so I extracted it.

src/base/projects/BranchSelector.svelte
```diff
@@ -0,0 +1,123 @@
+<script lang="ts">
+  import { getBranchesByCommit, Info } from "@app/project";
+  import { formatCommit } from "@app/utils";
+
+  export let branches: [string, string][];
+  export let project: Info;
+  export let selectedBranch: [string, string] | null;
```
sebastinez, line 7: The `selectedBranch` information here is persisted in the `View` component and gets passed through the `Header` component until here, this is to allow being persistent while the Header and the BranchSelector get rerendered with changing commits.

src/base/projects/BranchSelector.svelte
```diff
@@ -0,0 +1,123 @@
...
+  export let commit: string;
+
+  let branchesDropdown = false;
+  const switchBranch = (name: string, head: string) => {
+    commit = head;
+    selectedBranch = [name, head];
+  };
+
+  $: branchNames = getBranchesByCommit(commit, branches);
```
sebastinez, line 16: Since there could be commits that are head of multiple branches, we return here a stringarray with all the branch names that fit the commit. and show them in the UI as eg. `master|dev`

cloudhead:
I've noticed an unintended change in the styling: there used to be a separation of bg color between branch name and commit, which is gone:
image

src/base/projects/BranchSelector.svelte
cloudhead: How is this used? I'm not able to try this out? This should only be relevant when viewing a commit, which we don't have right now.

cloudhead: Or perhaps it's used when viewing the anchored state, and that happens to belong to be the same across multiple branches? In that case I think only the default branch is relevant.

sebastinez: Screenshot from 2022-01-10 13-49-28
The issue is when you have two branches pointing to the same commit, shouldn't be something that occurs often, but I think it can be a possible edge case.
In that case which branch name should we display? I went for both :sunglasses:

src/base/projects/Header.svelte
```diff
@@ -222,63 +202,49 @@
...
-            <span class="anchor-widget" on:click={() => commit = project.head}>
-              <span class="anchor-label" title="{anchors}">latest 🔐</span>
-            </span>
-          <!-- commit is not head a stale anchor  -->
-          {:else if anchor?.includes(commit)}
-            <span class="anchor-widget" on:click={() => commit = anchor[0]}>
-              <span class="anchor-label" title="{anchors}">stale 🔒</span>
-            </span>
-          <!-- commit is not anchored, could be head or any other commit  -->
+  {#if commit}
```
sebastinez, line 205: The if clause here allows us to have initially an undefined commit, and after mounting the nested routes render this part.

src/base/projects/BranchSelector.svelte
cloudhead: Hmm I think the problem here is that we actually don't support branch names in the route. If a user selects `dev`, it should show `dev`. It doesn't matter that other branches point to the same commit. But because the branch selector just selects commits, we have this problem. Could we have a branch path? eg. `/tree/master`?

src/base/projects/Commit/History.svelte
```diff
@@ -13,6 +13,9 @@
     const commitsQuery = await getCommits(urn, commit, config);
     return groupCommitHistory(commitsQuery);
   }
+
+  // Updates the project store with route params
+  projectStore.update(project => ({ ...project, commit, path: project.path ?? "/", content: ProjectContent.History }));
```
cloudhead, line 18: Shouldn't we use `||` here, since `""` should be changed to `"/"`?

src/base/projects/Header.svelte
```diff
@@ -23,7 +23,12 @@
 
   // Switches between the browser and commit view
   const switchContent = () => {
-    content = content == ProjectContent.Browser ? ProjectContent.Commits : ProjectContent.Browser;
+    projectStore.update(project => ({ ...project, content: project.content === ProjectContent.History ? ProjectContent.Tree : ProjectContent.History }));
+  };
```
cloudhead, line 27: Hmm I think this function should take a `ProjectContent` as parameter, and toggle it. The logic should be:

*if the current content equals the one passed in, set to history (default)*
*else, set it to the content passed in*

I'd also name it `toggleContent`

src/project.ts
```diff
@@ -140,6 +151,12 @@ export function path(
   }
   result.push("projects", urn);
 
+  if (content == ProjectContent.History) {
+    result.push("history");
+  } else {
+    result.push("tree");
```
cloudhead, line 157: I'd use a switch here, there will be other cases in the future.

src/utils.ts
```diff
@@ -47,7 +47,7 @@ export enum Status {
     Pending,
     Success,
     Failed,
-  }
+}
```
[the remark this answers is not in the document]
sebastinez: fixed in c1e2717cae73295799262c6a6b02e52d988653c4

src/project.ts
sebastinez: done in c1e2717cae73295799262c6a6b02e52d988653c4

src/base/projects/Header.svelte
sebastinez: done in c1e2717cae73295799262c6a6b02e52d988653c4

src/base/projects/BranchSelector.svelte
sebastinez: ok this has been implemented in c1e2717cae73295799262c6a6b02e52d988653c4

src/base/projects/BranchSelector.svelte
```diff
@@ -0,0 +1,139 @@
...
+  }
+  @media (max-width: 720px) {
+    .dropdown {
+      left: 32px;
+      z-index: 10;
+    }
+  }
+</style>
+
+  <div class="commit">
```
cloudhead, line 88: Needs to be un-indented.

src/base/projects/BranchSelector.svelte
```diff
@@ -0,0 +1,139 @@
...
+        </div>
+      {:else}
+        <div class="hash desktop"> 
+          {commit}
+        </div>
+        <div class="hash mobile">
+          {formatCommit(commit)}
+        </div>
+      {/if}
+    <!-- If there are is no branch listing available, show default branch name if commit is head and else show entire commit -->
```
cloudhead, line 123:
```suggestion
    <!-- If there is no branch listing available, show default branch name if commit is head and else show entire commit -->
```

src/base/projects/Header.svelte
```diff
@@ -330,7 +300,7 @@
       {/if}
     </div>
   </span>
-  <div class="stat commit-count" class:active={content == ProjectContent.Commits} on:click={switchContent}>
+  <div class="stat commit-count" class:active={content == ProjectContent.History} on:click={() => toggleContent(ProjectContent.Tree)}>
```
cloudhead, line 303: This should toggle `ProjectContent.History`, no?

src/project.ts
```diff
@@ -152,3 +172,28 @@ export function path(
   }
   return "/" + result.join("/");
 }
+
+// We need a sha1 commit for the API call so we return early if rev is a sha1 and else we look into branches.
+// As fallback we go for the head
+// TODO: Add tags parsing
+export function parseRevision(head: string, revision: string, branches?: [string, string][]): string {
```
cloudhead, line 179: It took me a while to understand what this function did, even after reading the comment.. Maybe calling it `getOid` would be more appropriate? Since we're not really parsing the revision.

src/project.ts
```diff
@@ -152,3 +172,28 @@ export function path(
...
+  if (isOid(revision)) return revision;
+  if (branches) {
+    const branch = branches.find(([name,]) => name === revision);
+    return branch ? branch[1] : head;
+  }
+  return head;
+}
+
+// Splits the path consisting out of an eventual sha1 commit or branch name and the path into a tuple [revision, path]
```
cloudhead, line 188:
```suggestion
// Splits the path consisting of a revision (eg. branch or commit) and file path into a tuple [revision, file-path]
```

src/project.ts
```diff
@@ -152,3 +172,28 @@ export function path(
...
+export function splitPrefixFromPath(input: string, branches: [string, string][], head: string): [string, string] {
+  const branch = branches.find(([branchName,]) => input.includes(branchName));
```
cloudhead, line 190: We can be more specific here, because the branch should always be the prefix, so I think `input.indexOf(branchName) === 0` is better. It may also prevent some errors.

src/utils.ts
```diff
@@ -218,6 +223,11 @@ export function safeLink(addr: string, config: Config): string {
   return explorerLink(addr, config);
 }
 
+// Splits a string at a defined index.
+export function splitAt(input: string, index: number): [string, string] {
+  return [input.slice(0, index), input.slice(index + 1)];
+}
```
cloudhead, line 229: This function skips a character, is that intended?
```
splitAt("abcdef", 3)
[ "abc", "ef" ]
```

src/project.ts
sebastinez: Yeah, I think eventually to be more explicit we could use `String.startsWith`

src/utils.ts
sebastinez: Yes, this skips a character because we don't want the path to start with a "/" and I thought better do it here..
But maybe to maintain the logic of the function we could move that to `splitPrefixFromPath`

cloudhead: Hmm, but this function is exported as a general utility. Can we instead embed it inside the function that uses it?

src/base/projects/BranchSelector.svelte
```diff
@@ -0,0 +1,139 @@
...
+    </div>
+  {:else}
+    <div class="hash desktop"> 
+      {commit}
+    </div>
+    <div class="hash mobile">
+      {formatCommit(commit)}
+    </div>
+  {/if}
+</div>
```
cloudhead, line 139: Can you strip the trailing whitespace in this file?

PR #147, opened 2022-01-13
Reactivity fixes

cloudhead:
Fixes lack of reactivity when you:

1. open file browser, choose a file, then press back
2. open history browser, switch between branches, then press back

Also simplifies things by using two-way bindings instead of events.

### DIFF
--- a/src/base/projects/View.svelte
+++ b/src/base/projects/View.svelte
@@ -174,7 +174,8 @@
         branches={result.branches}
         config={result.config}
         bind:content={content}
-        on:routeParamsChange={updateRouteParams} />
+        bind:revision={revision}
+        bind:path={path} />
     {:catch err}
       <div class="container center-content">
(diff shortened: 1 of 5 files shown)


src/base/projects/View.svelte
```diff
@@ -169,11 +169,13 @@
         branches={result.branches}
         profile={result.profile}
         on:routeParamsChange={updateRouteParams} />
```
sebastinez, line 171: Okay I tried some ways to remove the `routeParamsChange` event from the `Header` but I think for now it works and it does what it supposed to do quite nicely.
Maybe in the future we could remove/improve

sebastinez (approved):

src/base/projects/View.svelte
cloudhead: Yeah it makes a bit more sense there because it's triggered by a click, vs. just `onMount`

PR #151, opened 2022-01-13
Add plausible analytics

### LINKED ISSUE #146: Analytics
It would be good to start collecting some analytics on the web app in a privacy-preserving manner.

One popular open source solution is https://github.com/plausible/analytics which we can self host or pay to host. It doesn't store IPs and doesn't use cookies.

sebastinez:
This PR adds the JS snippet for using Plausible as analytics provider for `app.radicle.network`

Closes #146

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -64,6 +64,9 @@
 <svelte:window on:keydown={handleKeydown} />
 <svelte:head>
   <title>Radicle</title>
+  {#if import.meta.env.PROD}
+    <script defer data-domain="app.radicle.network" src="https://plausible.io/js/plausible.js"></script>
+  {/if}
 </svelte:head>
 
 <div class="app">


cloudhead:
Does this require us to create an account or something?

Also maybe this should only be included in prod?

sebastinez:
> Does this require us to create an account or something?

To use the hosted service we need to create an account (https://plausible.io/docs/register-account)

> Also maybe this should only be included in prod?

Looking into how add the analytic snippet just in case of building for prod, and found this package what do you think?
Looked over the source code and seems pretty straight forward
https://github.com/trapcodeio/vite-plugin-ejs#readme
The idea is using the template language ejs add the snippet like 
```
<% if(!isDev){ %>
  <script defer data-domain="app.radicle.network" src="https://plausible.io/js/plausible.js"></script>
<% } %>
```
another way would be to do a index.html for dev and one for prod, but seems way too redundant

cloudhead (approved):

cloudhead:
Merged.

PR #152, opened 2022-01-13
Seed address component

### LINKED ISSUE #142: Small CSS issue on seed page
image

On the right is how it should look (org page), on the left is the seed page.

I fixed it in the org page, but maybe it's a good idea to create a component for it, eg. `SeedID.svelte`, and share it accross user/org/seed.

sebastinez:
- Refactors seed address into separate component.
- Adds SeedAddress component to Orgs and Seed View page
- Adds SeedAddress to user page

Closes #142


src/base/orgs/View.svelte
```diff
@@ -17,6 +17,7 @@
   import { Profile, ProfileType } from '@app/profile';
   import Projects from './View/Projects.svelte';
   import Link from '@app/Link.svelte';
+  import SeedId from '@app/SeedID.svelte';
```
cloudhead, line 20: Maybe name the file `SeedId` as well

cloudhead: Actually it's called a "SeedAddress", the ID is just the PeerID without the hostname.

PR #157, opened 2022-01-14
Support for browsing seed projects

cloudhead:
Eg. /seeds/seed.cloudhead.io/projects/:urn

Signed-off-by: Alexis Sellier <self@cloudhead.io>

### DIFF
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -21,6 +21,7 @@
   export let config: Config;
   export let org: string | null = null;
   export let user: string | null = null;
+  export let seed: string | null = null;
   export let tree: proj.Tree;
   export let project: proj.Info;
   export let branches: [string, string][];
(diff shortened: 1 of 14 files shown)


src/base/projects/Browser.svelte
```diff
@@ -21,6 +21,7 @@
   export let config: Config;
   export let org: string | null = null;
   export let user: string | null = null;
+  export let seed: string | null = null;
```
sebastinez, line 24: Why are we using a fallback to `null` here, it seems that neither in `View` or `ProjectContentRoutes` we could encounter us with a missing org, user or seed, since they are `""` from the beginning

sebastinez (approved):

src/base/projects/Browser.svelte
cloudhead: hmm good question, will look into this.

PR #158, opened 2022-01-14
Fix avatar aspect ratio

### LINKED ISSUE #155: Avatar images don't respect the aspect ratio
Some avatar images from user and orgs show up as ellipse instead of circles.
Probably an issue in the `Avatar` component

Screenshot from 2022-01-14 10-13-20

sebastinez:
This PR uses the `object-fit` attribute to the Avatar component, so that the avatars fill their container, if they are not perfectly square. Which in most cases should keep it the same but in some cases maintains the form factor of the Avatars

Closes #155

### DIFF
--- a/src/Avatar.svelte
+++ b/src/Avatar.svelte
@@ -33,6 +33,8 @@
     min-width: 1rem;
     min-height: 1rem;
     height: 100%;
+    width: inherit;
+    object-fit: cover;
     background-size: cover;
     background-repeat: no-repeat;
   }


cloudhead:
This is broken for me. Did you test it?
image

sebastinez:
> This is broken for me. Did you test it? image

Hey @cloudhead, yes I did test it with multiple avatars, blockies and images but the one you pointed out, did escape me.
I think I found a better solution by using `inherit` width from the parent, it seems to work with images and blockies in both scenarios.
And the `object-fit: cover` avoids that images get stretched thin or too big.

PR #165, opened 2022-01-21
Audit fixes and improved anchors

sebastinez:
Ok this PR is a bit mixed..

Started out to fix the high severity issues from `npm audit` touching gnosis safe sdk and others.
And in the way I found some issues with the anchor proposal design so I took the liberty to address those too.

I feel that the anchor proposal system in the web ui could have more work done but I would like to tackle that in another PR.


sebastinez:
Okay it seems there are some issues with dependencies of `@gnosis.pm/safe-core-sdk` and typescript will look into it.

sebastinez:
Okay this PR on the typechain repo should fix the gnosis/safe-core-sdk issue with importing variables as values instead of types.
https://github.com/dethcrypto/TypeChain/pull/607

PR #170, opened 2022-01-31
Orgs / User Route Refactor

sebastinez:
This PR refactors the `/users/:user` and `/orgs/:orgs` routes and replaces them with `/:addressOrName` routes.
Also the projects in org or user context are simplified to `/:addressOrName/:projects`.

Big points:
- Extract all the anchored project querying into `base/orgs/View/Projects.svelte`
- Join the `orgs/View` and `/users/View` components into `Profile.svelte`


src/base/orgs/View/Projects.svelte
```diff
@@ -40,36 +43,65 @@
...
-          </Widget>
-        {/if}
-      </div>
-    {/each}
-  {:catch err}
-    <Message error>
-      <strong>Error: </strong> failed to load projects: {err.message}.
-    </Message>
-  {/await}
+  {#if profile.org}
```
sebastinez, line 46: Okay this is to fetch all the org projects and else query the `profile.anchorsAccount` projects..
We could also separate them into separate components.. eg. `PendingProjects` `AnchoredProjects`

src/profile.ts
```diff
@@ -128,31 +141,52 @@ export class Profile {
...
+    let type = AddressType.EOA;
+    let org: Org | null = null;
     const ens = await resolveEnsProfile(addressOrName, profileType, config);
 
     if (ens) {
       if (ens.address) {
+        type = await identifyAddress(ens.address, config);
+
+        if (type === AddressType.Org) {
+          org = await Org.get(ens.address, config);
```
sebastinez, line 153: The nicest way I found for Orgs is to be part of the Profile type as a kind of subset, to allow us to query for orgs and get all the properties and methods linked to.

src/base/projects/Widget.svelte
```diff
@@ -35,13 +34,13 @@
     }
   });
 
+  $: console.log(addressOrName);
```
cloudhead, line 37:
```suggestion
```

src/Profile.svelte
```diff
@@ -0,0 +1,380 @@
...
+  };
+  $: isOrgAuthorized = async (org: Org): Promise<boolean> => {
+    if ($session) {
+      if (isOwner(org)) {
+        return true;
+      }
+      return await org.isMember($session.address, config);
+    }
+    return false;
+  };
```
cloudhead, line 69: I know this wasn't introduced in this PR, but why is it that all these functions are defined as reactive?

src/Profile.svelte
```diff
@@ -0,0 +1,380 @@
...
+  let setNameForm: typeof SvelteComponent | null =
+    action === "setName" ? SetName : null;
+  const setName = () => {
+    setNameForm = SetName;
+  };
+
+  let transferOwnerForm: typeof SvelteComponent | null = null;
+  const transferOwnership = () => {
+    transferOwnerForm = TransferOwnership;
+  };
```
cloudhead, line 34:
```suggestion
  };
  
```

src/Profile.svelte
sebastinez: They are required to change the permissions on setting profile and changing owner, eg when the logged in user changes.

src/base/orgs/Routes.svelte
```diff
@@ -12,6 +11,6 @@
   <Index {config} />
 </Route>
 
-<Route path="/orgs/:address" let:params let:location>
-  <View {config} addressOrName={params.address} action={getSearchParam("action", location)} />
```
sebastinez, line 16: I removed the action param here, since this is only a redirection.

src/Profile.svelte
cloudhead: Hmm not sure that answers my question, since they are functions that would have to be called anyway.. I'll look into it.

cloudhead (approved):

PR #175, opened 2022-02-01
Move signer into session state

sebastinez:
This PR moves the signer "clones" the signer property from the `Config` object into the `session` state, to allow Svelte reactivity, e.g. in the Testnet Faucet, where we rely on `config.signer` but since it only changes when the `Config` object in `App.svelte` changes we can't react on it.

For now I only cloned the signer property, if we see in the future that there are other properties that need to change with session we can add them.

### DIFF
--- a/src/Header.svelte
+++ b/src/Header.svelte
@@ -17,7 +17,6 @@
   export let session: Session | null;
   export let config: Config;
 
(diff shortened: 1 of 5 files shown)


cloudhead (approved):

cloudhead:
Merged.

PR #178, opened 2022-02-02
Toggle peer and branch dropdown and fix header commit btn

sebastinez:
If the branch or peer dropdown toggle have been selected, and the user wants to select the same peer or branch again, the page does not re render and the dropdown stays open.
This PR hides the dropdown when clicking on any peer or branch, re-rendering or not.

And the commit button in a navigation situation of project -> commit -> [back] was not highlighted, which should be fixed.


src/base/projects/Commit/CommitTeaser.svelte
```diff
@@ -68,6 +68,8 @@
   <div class="right">
     <span class="bold author">{commit.committer.name}</span>
     <span class="desktop-inline time">{formatCommitTime(commit.committerTime)}</span>
-    <Icon name="browse" clickHandler={() => browseCommit(commit.sha1)} width={17} inline fill />
+    <div style:height="1.6rem" on:click={() => browseCommit(commit.sha1)}>
```
cloudhead, line 71: Where is this 1.6rem coming from?

cloudhead:
Hey, I've taken the liberty to simplify this by moving the `<span>` inside `Icon` and using the default event propagation for `on:click`.

Issue #171, opened 2022-01-31
Support project names


cloudhead:
Instead of `/seeds/:seed/projects/:urn` we should also support `/seeds/:seed/projects/:name`, where `:name` is for eg. `nakamoto`. This will require support in `http-api` (it already is supported in `git-server`).

sebastinez:
Ok have a PR in `http-api` for this, https://github.com/radicle-dev/radicle-client-services/pull/110
I'm not sure if we want to do `/seeds/:seed/projects/:urn` wouldn't it be nicer to have `/seeds/:seed/:urn` or `/:name` as suffix?

PR #182, opened 2022-02-03
Allow users to pass project names into URL

sebastinez:
This PR closes issue #171

### DIFF
--- a/src/base/projects/View.svelte
+++ b/src/base/projects/View.svelte
@@ -7,7 +7,7 @@
   import Avatar from '@app/Avatar.svelte';
   import { Profile, ProfileType } from '@app/profile';
   import type { Info } from '@app/project';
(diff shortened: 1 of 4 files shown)


src/base/projects/View.svelte
```diff
@@ -43,6 +43,9 @@
     const seedInstance = profile?.seed ?? result?.seed;
     const cfg = seedInstance ? config.withSeed(seedInstance) : config;
     const info = await proj.getInfo(urn, cfg);
+
+    // Overwrite URN with a Radicle ID in case that the user has passed a project name
+    urn = isRadicleId(urn) ? urn : info.meta.urn;
```
cloudhead, line 48: Can we split this so that the input is `export nameOrUrn: string`, then set `urn` from that?

PR #184, opened 2022-02-04
Add link to seed node from registration page

### LINKED ISSUE #168: Link to seed node from registrations page
In eg. `/registrations/cloudhead`, the `seed host` value should link to the seed page.

sebastinez:
Closes #168

### DIFF
--- a/src/Form.svelte
+++ b/src/Form.svelte
@@ -10,6 +10,7 @@
     editable: boolean;
     error?: string;
     example?: string;
+    url?: string;
   }
 
   const validationExamples: Record<string, string> = {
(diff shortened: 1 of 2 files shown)


src/Form.svelte
```diff
@@ -191,6 +192,10 @@
               <div class="mobile">
                 <Address compact resolve={field.resolve ?? false} address={field.value} {config} />
               </div>
+            {:else if (field.label === "Seed Host")}
+              <div>
+                <a use:link href="/seeds/{field.value}">{field.value}</a>
+              </div>
```
cloudhead, line 198: This (and the Seed ID) check don't really belong here, as the field label isn't supposed to be an API. Will look into something cleaner.

sebastinez: Right, maybe we can add a check for `isSeedHost` and `isSeedId`? Similar to `isAddress` and `isUrl`?
So it doesn't depend on the field label

sebastinez: Okay that's a clean solution, thanks!

Issue #167, opened 2022-01-29
Support non-radicle registrations


cloudhead:
In the `/registrations/:name` page, we should support ALL ens domains, not just radicle.eth ones.

To do this, `/registrations/cloudhead` would become `/registrations/cloudhead.radicle.eth`.

Having the ability to edit any registration on that page would also be great.

One thing we *don't* want is to allow people to register new non-radicle ENS names from the website, since the registration process is different.

PR #185, opened 2022-02-04
Improve support for non-radicle ENS domains

### LINKED ISSUE #190: ENS names not supported in search bar
image

sebastinez:
This PR allows user to open non-radicle ENS domains in registration form and also edit them.
It also fixes an issue that wasn't allowing editing ENS information, due to passing `undefined` values for empty input fields, we now cast `undefined` to an empty string.

Closes #190, #167

### DIFF
--- a/src/Form.svelte
+++ b/src/Form.svelte
@@ -80,7 +80,8 @@
     return fields.filter(field => field.editable).map(field => {
       return {
         name: field.name,
-        value: field.value && field.value.trim(),
+        // We only allow to have a trueish value or an empty string.
+        value: field.value ? field.value.trim() : "",
       };
     });
   };
@@ -179,7 +180,7 @@
     <div>
       {#if field.editable && editable}
         <input name={field.name} class="field" placeholder={field.placeholder}
-               on:change={validate} value={field.value} type="text" {disabled} />
+               on:change={validate} value={field.value || ""} type="text" {disabled} />
       {:else}
         <span class="field">
           {#if field.value}
--- a/src/base/registrations/New.svelte
+++ b/src/base/registrations/New.svelte
@@ -20,11 +20,11 @@
   }
 
   export let config: Config;
-  export let subdomain: string;
+  export let name: string;
   export let owner: string | null;
 
   // We only support lower-case names.
-  subdomain = subdomain.toLowerCase();
+  name = name.toLowerCase();
 
   let fee: string;
   let state = State.CheckingAvailability;
@@ -32,7 +32,7 @@
   $: registrationOwner = owner || ($session && $session.address);
 
   function begin() {
-    navigate(`/registrations/${subdomain}/submit?${
+    navigate(`/registrations/${name}/submit?${
       registrationOwner ? new URLSearchParams({ owner: registrationOwner }) : ''
     }`);
   }
@@ -41,7 +41,7 @@
     try {
       const [_fee, isAvailable] = await Promise.all([
         registrationFee(config),
-        registrar(config).available(subdomain),
+        registrar(config).available(name),
       ]);
 
       fee = formatBalance(_fee);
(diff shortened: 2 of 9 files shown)


src/Form.svelte
```diff
@@ -179,7 +180,7 @@
     <div>
       {#if field.editable && editable}
         <input name={field.name} class="field" placeholder={field.placeholder}
-               on:change={validate} value={field.value} type="text" {disabled} />
+               on:change={validate} value={field.value || ""} type="text" {disabled} />
```
sebastinez, line 183: This is to avoid showing an "undefined" as value of an input field, we should prefer to show an empty string which allows us to show the placeholder as default

src/base/registrations/New.svelte
```diff
@@ -69,17 +69,17 @@
   </span>
 
   <span slot="subtitle">
-    {subdomain}.{config.registrar.domain}
+    {name}.{config.registrar.domain}
```
cloudhead, line 72: Won't this show up as eg. `cloudhead.eth.radicle.eth`?

src/base/registrations/resolver.ts
```diff
@@ -10,7 +10,7 @@ export async function setRecords(name: string, records: EnsRecord[], resolver: E
   assert(config.signer, "no signer available");
 
   const resolverContract = new ethers.Contract(resolver.address, config.abi.resolver, config.signer);
-  const node = ethers.utils.namehash(`${name}.${config.registrar.domain}`);
+  const node = ethers.utils.namehash(`${name}`);
```
cloudhead, line 13:
```suggestion
  const node = ethers.utils.namehash(name);
```

src/utils.ts
```diff
@@ -312,7 +312,7 @@ export async function identifyAddress(address: string, config: Config): Promise<
 
 // Resolve a label under the radicle domain.
 export async function resolveLabel(label: string | undefined, config: Config): Promise<string | null> {
-  if (label) return config.provider.resolveName(`${label}.${config.registrar.domain}`);
+  if (label) return config.provider.resolveName(`${label}`);
```
cloudhead, line 315:
```suggestion
  if (label) return config.provider.resolveName(label);
```

src/base/registrations/New.svelte
```diff
@@ -69,17 +69,17 @@
...
-        The name <strong>{subdomain}</strong> is available for registration
+        The name <strong>{name}</strong> is available for registration
         under account <strong>{formatAddress(registrationOwner)}</strong>
         for <strong>{fee} RAD</strong>.
       {:else}
-        The name <strong>{subdomain}</strong> is available
+        The name <strong>{name}</strong> is available
         for <strong>{fee} RAD</strong>.
       {/if}
     {:else if state === State.NameUnavailable}
```
cloudhead, line 85: Not sure why this file and the other was changed? `subdomain` seems more correct than `name`, since it's only part of an ENS name.

cloudhead (commented):
Looks good, except for the change of some of the variable names.

src/Form.svelte
cloudhead: Is this a new issue? Never seen this happen so far :thinking:

src/base/registrations/New.svelte
sebastinez: I changed it to name since for me the subdomain actually is `.radicle` in this case, and i took the name as eg `cloudhead` or `sebastinez`.
`<name>.<subdomain>.<TLD>`

src/Form.svelte
sebastinez: Yeah it seems that it was introduced somewhere on the way.
Since it currently is not possible to edit ones records. And it seems in part due to passing around undefined values or showing "undefined" as field value

src/base/registrations/New.svelte
cloudhead: Ok, I can live with that.

src/Form.svelte
sebastinez: Instead of doing it in the `Form.svelte` component, we could also move this to the `registrations/View.svelte` component and define an empty string for values that could be `undefined` but I thought it would be also be applicable for other forms when doing it here

PR #192, opened 2022-02-08
Anchors refactor

cloudhead:
* Anchors are loaded alongside projects, but project listings always come from seeds now
* Removing project info loading from `Widget`
* Start unifying anchor handling
* Rename "Project*" to "Anchor*" when appropriate
* Move to new project api with embedded metadata

### DIFF
--- a/src/base/orgs/View/Projects.svelte
+++ b/src/base/orgs/View/Projects.svelte
@@ -55,29 +55,31 @@
 </style>
 
 <div class="projects">
-  {#await Seed.getProjects(config)}
-    <Loading center />
-  {:then projects}
-    {#each projects as project}
-      {@const anchor = anchors[project.urn]}
-      {@const pendingAnchor = pendingAnchors[project.urn]}
-      <div class="project">
-        <Widget {project} {anchor} on:click={() => onClick(project)}>
-          <span class="actions" slot="actions">
-            {#if profile.org?.safe && account && anchor}
-              {#if pendingAnchor} <!-- Pending anchor -->
-                <AnchorActions
-                  {account} {config} anchor={pendingAnchor} safe={profile.org.safe}
-                  on:success={() => loadAnchors()} />
+  {#if config.seed.api.host}
+    {#await Seed.getProjects(config)}
+      <Loading center />
+    {:then projects}
+      {#each projects as project}
+        {@const anchor = anchors[project.urn]}
+        {@const pendingAnchor = pendingAnchors[project.urn]}
+        <div class="project">
+          <Widget {project} {anchor} on:click={() => onClick(project)}>
+            <span class="actions" slot="actions">
+              {#if profile.org?.safe && account && anchor}
+                {#if pendingAnchor} <!-- Pending anchor -->
+                  <AnchorActions
+                    {account} {config} anchor={pendingAnchor} safe={profile.org.safe}
+                    on:success={() => loadAnchors()} />
+                {/if}
               {/if}
-            {/if}
-          </span>
-        </Widget>
-      </div>
-    {/each}
-  {:catch err}
-    <Message error>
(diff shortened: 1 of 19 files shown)


sebastinez (commented):
To avoid fetching anchors from an ENS domain I commented on one line something.
Moving the logic to something more specific on Anchors and away from Projects makes sense

src/base/projects/View.svelte
```diff
@@ -40,18 +40,18 @@
     const seedInstance = profile?.seed ?? result?.seed;
     const cfg = seedInstance && seedInstance.valid ? config.withSeed(seedInstance) : config;
     const info = await proj.getInfo(id, cfg);
-    const urn = isRadicleId(id) ? id : info.meta.urn;
+    const urn = isRadicleId(id) ? id : info.urn;
     const anchors = await getAllAnchors(config, urn, profile?.anchorsAccount ?? addressOrName);
```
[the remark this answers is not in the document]
cloudhead: Fixed here: 8d18b0d

src/base/projects/View.svelte
```diff
@@ -40,18 +39,19 @@
     const seedInstance = profile?.seed ?? result?.seed;
     const cfg = seedInstance && seedInstance.valid ? config.withSeed(seedInstance) : config;
     const info = await proj.getInfo(id, cfg);
-    const urn = isRadicleId(id) ? id : info.meta.urn;
-    const anchors = await getAllAnchors(config, urn, profile?.anchorsAccount ?? addressOrName);
-    let branches = Array([info.meta.defaultBranch, info.head]) as [string, string][];
+    const urn = isRadicleId(id) ? id : info.urn;
+    const anchors = await profile?.confirmedProjectAnchors(urn, config);
```
sebastinez, line 43: Anchors can eventually be `undefined` here and in the `Source` object we only accept anchors as `string[]`
Maybe we can convert `confirmedProjectAnchors` into a static method?

cloudhead: ah I see, not sure how converting it into a static method would help... but we can just do `profile ? profile... : []`

sebastinez: Yeah we can do that too.

src/project.ts
```diff
@@ -8,12 +8,19 @@ export type Urn = string;
 export type Peer = string;
 export type Branch = { [key: string]: string };
 
-export interface ProjectListing {
-  name: string;
-  urn: Urn;
+export interface Anchor {
+  confirmed: true;
```
sebastinez, line 12: I understand the idea behind this new property, but I'm not sure where we are using it. Since we already have the `confirmations` array on the `PendingAnchor` type.

src/base/orgs/View/Projects.svelte
```diff
@@ -1,32 +1,44 @@
...
-        return result ? profile.org.getAllProjects(config) : profile.org.getProjects(config);
-      }
-    }
-    return [];
+  const onClick = (project: ProjectInfo) => {
+    navigate(
+      proj.path({
+        urn: project.urn,
+        addressOrName: profile.name ?? profile.address,
+        revision: project.head,
```
sebastinez, line 36: With this we go again to the `project.head` of a project instead of the latest anchor. Is this desired? Maybe we could navigate to the latest anchor of the project like `revision: anchors[project.urn].anchor.stateHash ?? project.head` or something similar.

cloudhead: Yeah, I think latest anchor is actually going to be confusing for users. I'd rather they click on the anchor button to get to that state, but users will expect to see the latest code whether anchored or not, imo.

sebastinez: Ok good for me 👍

PR #201, opened 2022-02-14
Some project refactors

sebastinez:
Creates a new `Project` class that has some useful static functions, and once created an instance with `Project.get` it allows us to pass the instance through components and have a unified API.

Also removes the `ProjectContentRoutes` and integrates the `Router` into the `View` component.


src/base/orgs/View/Projects.svelte
```diff
@@ -2,12 +2,12 @@
...
   import { onMount } from "svelte";
   import type { Config } from "@app/config";
-  import * as proj from "@app/project";
+  import * as proj from "@app/base/projects/Project";
   import Loading from "@app/Loading.svelte";
   import Message from "@app/Message.svelte";
   import Widget from '@app/base/projects/Widget.svelte';
   import type { Profile } from "@app/profile";
-  import type { ProjectInfo, Anchor, PendingAnchor } from "@app/project";
+  import type { ProjectInfo, Anchor, PendingAnchor } from "@app/base/projects/Project";
```
sebastinez, line 10: Personally I prefer the destructured way of importing, but I left the default `* as proj` imports. Maybe we should define when it makes sense and when not.

src/base/orgs/View/Projects.svelte
```diff
@@ -62,11 +62,11 @@
   {#await seed.getProjects(profile?.id)}
```
sebastinez, line 62: We have the option to return `Project` instances from this function, similar to the `Org.getMulti` function.
I did not implement it, under the reason that we would be fetching and doing a lot of overhead to get just `projectInfo` and we can leave that for when a user visits the project.

src/base/seeds/Seed.ts
```diff
@@ -1,6 +1,6 @@
 import * as api from '@app/api';
 import type { Config } from '@app/config';
-import * as proj from '@app/project';
+import * as proj from '@app/base/projects/Project';
```
cloudhead, line 3: Same comment as below, we shouldn't be importing things between sub-apps (`base/*`), this creates a dependency spaghetti.

src/base/projects/Project.ts
```diff
@@ -0,0 +1,338 @@
...
+    parsed.path = commitPath[1] ? commitPath[1] : "/";
+  } else {
+    parsed.path = input;
+  }
+  return parsed;
+}
+
+export class Project {
+  urn: string;
+  info: ProjectInfo;
```
cloudhead, line 199: I would extend `Project` with `ProjectInfo`. Then you can pass `Project` where `ProjectInfo` is expected, eg. BranchSelector.

cloudhead (changes requested):
This is very nice and clean! Left some comments that should make things even tighter.

src/base/projects/BranchSelector.svelte
```diff
@@ -1,11 +1,11 @@
 <script lang="ts">
   import { createEventDispatcher } from "svelte";
-  import { ProjectInfo, Branches, getOid } from "@app/project";
+  import { Project, Branches, getOid } from "@app/project";
   import { formatCommit } from "@app/utils";
   import Dropdown from "@app/Dropdown.svelte";
 
   export let branches: Branches;
-  export let project: ProjectInfo;
+  export let project: Project;
```
cloudhead, line 8: Could we not leave this with `ProjectInfo`, since `Project` implements it?

cloudhead (approved):
Looks great! Just one question.

cloudhead (approved):
Great, can you squash & rebase?

PR #204, opened 2022-02-21
Add delegate info to peer selector

### LINKED ISSUE #203: Improved peer selector
It would be cool if the peer selector:
* Showed the peer name next to the id, eg. `cloudhead <peer-id>`
* Showed whether the peer was a delegate or not for the project, eg. `cloudhead <peer-id> [delegate]`

sebastinez:
This PR closes #203

It adds the delegate metadata obtained from the seed nodes at the route `<host>/v1/projects/<urn>/remotes`
Which has the following format
```json
[
  {
    "id": "hyyg555wwkkutaysg6yr67qnu5d5ji54iur3n5uzzszndh8dp7ofue",
    "name": "sebastinez",
    "delegate": true
  }
]
```
And it shows it the following way.
Bildschirmfoto 2022-02-22 um 18 01 53

If there is no metadata provided by the seed node it maintains the current format showing the peer ids without additional metadata
Bildschirmfoto 2022-02-22 um 18 03 43

Also the current selected peer is being highlighted, the same applies to the branch selector which uses the same `Dropdown` component underneath.
Bildschirmfoto 2022-02-22 um 18 04 49

### DIFF
--- a/src/base/projects/PeerSelector.svelte
+++ b/src/base/projects/PeerSelector.svelte
@@ -1,13 +1,28 @@
 <script lang="ts">
-  import { createEventDispatcher } from "svelte";
+  import { createEventDispatcher, onMount } from "svelte";
   import Icon from "@app/Icon.svelte";
   import Dropdown from "@app/Dropdown.svelte";
   import { formatSeedId } from "@app/utils";
+  import type { Peer } from "@app/project";
 
   export let peer: string | null = null;
-  export let peers: (string | { id: string })[];
+  export let peers: Peer[];
   export let toggleDropdown: (input: string) => void;
   export let peersDropdown = false;
+
+  let meta: Peer | undefined;
+  // List of items to be created for the Dropdown component.
+  let items: { key: string; value: string; badge: string | null }[] = [];
+
+  onMount(() => {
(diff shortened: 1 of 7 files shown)


src/base/projects/PeerSelector.svelte
```diff
@@ -1,14 +1,29 @@
...
+
+  onMount(() => {
+    meta = peers.find(p => p.id === peer);
+    items = peers.map(p => {
+      if (! p.name) console.debug("Not able to resolve peer identity for: ", p.id);
+      let key = p.name ? `${p.name} ${p.id}` : p.id;
+
+      return { key, value: p.id, badge: p.delegate ? "delegate" : null };
+    });
+  });
```
cloudhead, line 25: Is there a reason these vars are set in `onMount` vs. just being set directly as consts?

sebastinez: Hmm I wanted to be explicit when the setting should happen, but it could be also outside of a `onMount` callback

cloudhead: Ok, we can change it in a later commit. Looks good!

PR #211, opened 2022-02-23
Add code activity graph to project widget

sebastinez:
Already working but there are some optimizations left to do.

- General refactoring to make the code simpler.
- Improve the display of the graphs (we have some graphs of peaks where 1 or 2 commits seem the same as 100+ commits in a week)

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -16,6 +16,7 @@
   import Header from '@app/Header.svelte';
   import Loading from '@app/Loading.svelte';
   import Modal from '@app/Modal.svelte';
+  import LinearGradient from "@app/LinearGradient.svelte";
 
   const loadConfig = getConfig().then(async cfg => {
     if ($state.connection === Connection.Connected) {
@@ -112,4 +113,6 @@
       </Modal>
     </div>
   {/await}
+  <!-- Adds a svg linear gradient that can be used by any other svg under the id #gradient -->
+  <LinearGradient id="gradient" />
 </div>
--- a/src/base/orgs/View/Projects.svelte
+++ b/src/base/orgs/View/Projects.svelte
@@ -67,7 +67,7 @@
       {@const pendingAnchor = pendingAnchors[project.urn]}
       {#if project.head}
         <div class="project">
-          <Widget {project} {anchor} on:click={() => onClick(project)}>
+          <Widget {project} {seed} {anchor} on:click={() => onClick(project)}>
             <span class="actions" slot="actions">
               {#if profile?.org?.safe && account && anchor}
                 {#if pendingAnchor} <!-- Pending anchor -->
--- a/src/base/projects/Widget.svelte
+++ b/src/base/projects/Widget.svelte
@@ -1,26 +1,69 @@
 <script lang="ts">
   import type * as proj from '@app/project';
   import AnchorBadge from '@app/base/profiles/AnchorBadge.svelte';
+  import Diagram from '@app/Diagram.svelte';
+  import { groupCommitsByWeek } from '@app/commit';
+  import { Project } from '@app/project';
+  import type { Seed } from '@app/base/seeds/Seed';
 
   export let project: proj.ProjectInfo;
+  export let seed: Seed;
   export let faded = false;
   export let anchor: proj.Anchor | null = null;
+
+  const getTimestampOneYearAgo = () => {
+    const now = new Date();
+    const oneYearAgo = new Date(now.getFullYear() - 1, now.getMonth(), now.getDate());
+    return Math.floor(oneYearAgo.getTime() / 1000).toString();
+  };
+
+  const loadCommits = async () => {
+    const commits = await Project.getCommits(project.urn, seed.api, project.head ?? undefined, getTimestampOneYearAgo(), undefined, "500", "0");
+    return groupCommitsByWeek(commits.headers);
+  };
 </script>
 
 <style>
   article {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
     padding: 1rem;
     border: 1px solid var(--color-secondary-faded);
     border-radius: 0.25rem;
     min-width: 36rem;
+    min-height: 140px;
     cursor: pointer;
+  }
+  article .right {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    align-items: flex-end;
+  }
+  article .left {
+    width: 50%;
+  }
+  div .description {
+    overflow-x: hidden;
+    text-overflow: ellipsis;
   }
   article.project-faded {
     border: 1px dashed var(--color-foreground-subtle);
     cursor: not-allowed;
   }
+  .activity {
+    width: 100%;
+    max-width: 14rem;
+  }
   article:hover {
     border-color: var(--color-secondary);
+  }
+  article:hover .anchor {
+    display: block;
+  }
+  article:hover .activity {
+    display: none !important;
   }
   article.project-faded:hover {
     border-color: var(--color-foreground-faded);
@@ -34,7 +77,7 @@
     margin-bottom: 0.25rem;
     font-size: 0.75rem;
   }
-  article .anchor {
+  article .stateHash {
     color: var(--color-secondary);
     font-size: 0.75rem;
     min-height: 2rem;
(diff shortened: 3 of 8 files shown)


sebastinez:
Ok it's ready, I think it's stable enough now.
The only issue is that e.g. one commit in a new project gets a massive spike while a project with many commits per week seem less spiky..

I think there is a solution there to distribute the activity graphs evenly between all projects, but I think thats for another PR.

src/commit.ts
```diff
@@ -108,3 +111,57 @@ export function groupCommits(commits: CommitHeader[]): CommitGroup[] {
...
+
+  commits = commits.sort((a, b) => {
+    if (a.committerTime > b.committerTime) {
+      return -1;
+    } else if (a.committerTime < b.committerTime) {
+      return 1;
+    }
+
+    return 0;
+  });
```
cloudhead, line 127: Note that this is sorting in reverse order, ie. from latest commit to earliest. Is this intended?

src/base/seeds/Seed.ts
```diff
@@ -100,12 +100,14 @@ export class Seed {
...
-  async getProjects(id?: string): Promise<proj.ProjectInfo[]> {
+  async getProjects(config: Config, id?: string): Promise<proj.Project[]> {
     const result = id
       ? await proj.Project.getDelegateProjects(id, this.api)
       : await proj.Project.getProjects(this.api);
 
-    return result.map((project: proj.ProjectInfo) => ({ ...project, id: project.urn }));
+    return Promise.all(
+      result.map((project: proj.ProjectInfo) => proj.Project.get(project.urn, null, null, this.api.host, config))
+    );
```
cloudhead, line 110: I think it would be best to separate these api queries, otherwise we have to wait for lots of requests to complete before the page can render.

src/base/orgs/View/Projects.svelte
```diff
@@ -59,25 +60,27 @@
...
       {#if project.head}
         <div class="project">
-          <Widget {project} {anchor} on:click={() => onClick(project)}>
-            <span class="actions" slot="actions">
-              {#if profile?.org?.safe && account && anchor}
-                {#if pendingAnchor} <!-- Pending anchor -->
-                  <AnchorActions
-                    {account} {config} anchor={pendingAnchor} safe={profile.org.safe}
-                    on:success={() => loadAnchors()} />
+          {#await project.getCommits(project.head) then points}
```
cloudhead, line 71: We shouldn't be waiting for this to load for the widget to render, ideally this is set on the widget once it is done loading in the background.

I also thought we wanted to only use the last 52 weeks of commits?

src/commit.ts
sebastinez: I tried different ways of building the graph and the easiest way I found was to start the graph from the right side = latest, filling up the weeks with no commits until the latest and from there go back to the earliest.

src/base/orgs/View/Projects.svelte
sebastinez: > Another thing is that using this method forces us to fetch the project before getting its history, which shouldn't be necessary. Some options would be:

Yeah makes sense 👍 

> I also thought we wanted to only use the last 52 weeks of commits?

Yeah this escaped me, will implement it.

sebastinez: Ok added the last 52 weeks of commits, and moved `getCommits` to be a class method like
`static async getCommits(urn: string, host: api.Host, parent: string, since?: string, until?: string): Promise<CommitsHistory>`

src/base/seeds/Seed.ts
sebastinez: Ok reverted this to just map the urn to `id`

sebastinez:
Also found a edge case when a project didn't have a commit in the last year the svg path wasn't drawed because of a calculation error in the normalization formula

src/base/projects/Widget.svelte
```diff
@@ -81,37 +106,62 @@
...
         </slot>
       </span>
-    </span>
+    </div>
+  </div>
+  <div class="right">
+    <div class="id">
+      <span class="urn desktop">{project.urn}</span>
+    </div>
+    {#if widgetHover}
```
cloudhead, line 137: Can CSS not be used for this? `article:hover .anchor { display: block }` `article:hover .activity { display: none }`?

cloudhead:
Made some tweaks to the visuals, can you apply this?

```diff
diff --git a/src/Diagram.svelte b/src/Diagram.svelte
index a80326a..72ece40 100644
--- a/src/Diagram.svelte
+++ b/src/Diagram.svelte
@@ -75,9 +75,7 @@
   <defs>
     <linearGradient id="gradient" x1="0" y1="1" x2="0" y2="0">
       <stop offset="0%" stop-color={fillColor} stop-opacity="0" />
-      <stop offset="10%" stop-color={fillColor} stop-opacity="0.1" />
-      <stop offset="25%" stop-color={fillColor} stop-opacity="0.2" />
-      <stop offset="50%" stop-color={fillColor} stop-opacity="0.3" />
+      <stop offset="100%" stop-color={fillColor} stop-opacity="0.4" />
     </linearGradient>
   </defs>
   <g>
diff --git a/src/base/projects/Widget.svelte b/src/base/projects/Widget.svelte
index da0c3d2..18008ae 100644
--- a/src/base/projects/Widget.svelte
+++ b/src/base/projects/Widget.svelte
@@ -43,6 +43,7 @@
   }
   .activity {
     width: 100%;
+    max-width: 14rem;
   }
   article:hover {
     border-color: var(--color-secondary);
@@ -154,9 +155,9 @@
         </span>
       </div>
     {:else}
-      <div class="activity">
+      <div class="activity desktop">
         <Diagram
-          strokeWidth={2}
+          strokeWidth={1.25}
           points={groupedCommits}
           viewBoxHeight={100}
           viewBoxWidth={600}

```

sebastinez:
> Made some tweaks to the visuals, can you apply this?
> 
> ```diff
> diff --git a/src/Diagram.svelte b/src/Diagram.svelte
> index a80326a..72ece40 100644
> --- a/src/Diagram.svelte
> +++ b/src/Diagram.svelte
> @@ -75,9 +75,7 @@
>    <defs>
>      <linearGradient id="gradient" x1="0" y1="1" x2="0" y2="0">
>        <stop offset="0%" stop-color={fillColor} stop-opacity="0" />
> -      <stop offset="10%" stop-color={fillColor} stop-opacity="0.1" />
> -      <stop offset="25%" stop-color={fillColor} stop-opacity="0.2" />
> -      <stop offset="50%" stop-color={fillColor} stop-opacity="0.3" />
> +      <stop offset="100%" stop-color={fillColor} stop-opacity="0.4" />
>      </linearGradient>
>    </defs>
>    <g>
> diff --git a/src/base/projects/Widget.svelte b/src/base/projects/Widget.svelte
> index da0c3d2..18008ae 100644
> --- a/src/base/projects/Widget.svelte
> +++ b/src/base/projects/Widget.svelte
> @@ -43,6 +43,7 @@
>    }
>    .activity {
>      width: 100%;
> +    max-width: 14rem;
>    }
>    article:hover {
>      border-color: var(--color-secondary);
> @@ -154,9 +155,9 @@
>          </span>
>        </div>
>      {:else}
> -      <div class="activity">
> +      <div class="activity desktop">
>          <Diagram
> -          strokeWidth={2}
> +          strokeWidth={1.25}
>            points={groupedCommits}
>            viewBoxHeight={100}
>            viewBoxWidth={600}
> ```

Ok addressed all the open issues, just one thing, here
```diff
diff --git a/src/base/projects/Widget.svelte b/src/base/projects/Widget.svelte
index da0c3d2..18008ae 100644
--- a/src/base/projects/Widget.svelte
+++ b/src/base/projects/Widget.svelte
@@ -154,9 +155,9 @@
         </span>
       </div>
     {:else}
-      <div class="activity">
+      <div class="activity desktop">
         <Diagram
           strokeWidth={1.25}
           points={groupedCommits}
           viewBoxHeight={100}
           viewBoxWidth={600}
```

The definition of `desktop` is
```css
.desktop {
  display: block !important;
}
```
if we add this here, we won't have the hover functionality and just the anchor showing up when hovering over the widget, is this your idea?

sebastinez:
> My intention was just to hide the activity graph on mobile, since it overlaps with everything else. Am I missing something?

Okay did some additional changes to your diff, in short:
- The Widget was resizing its height when hovering, I increased the min-height to 140px.
- There was an issue with using the same class `anchor` for the anchor actions and the `stateHash` so I added some differentiation.
- I extracted the linear gradient from the Diagram svg and put it inside `App.svelte` this on purpose so we can reuse the same gradient inside different components and it doesn't screw with the Widget layout.
- I added an `!important` tag to the `article:hover .activity` selector, since the `desktop` class would else overwrite the toggling visbility behaviour by hovering over the widget.

Let me know what you think, but I think we are close 😄

src/base/projects/LinearGradient.svelte
```diff
@@ -0,0 +1,13 @@
...
+</script>
+
+  <svg style="height: 0; width: 0;" xmlns="http://www.w3.org/2000/svg">
+    <defs>
+      <linearGradient {id} x1="0" y1="1" x2="0" y2="0">
+        <stop offset="0%" stop-color={fillColor} stop-opacity="0" />
+        <stop offset="100%" stop-color={fillColor} stop-opacity="0.4" />
+      </linearGradient>
+    </defs>
+  </svg>
```
cloudhead, line 13: Indentation :eyes:

src/base/orgs/View/Projects.svelte
```diff
@@ -67,17 +75,19 @@
...
       {#if project.head}
         <div class="project">
-          <Widget {project} {anchor} on:click={() => onClick(project)}>
-            <span class="actions" slot="actions">
-              {#if profile?.org?.safe && account && anchor}
-                {#if pendingAnchor} <!-- Pending anchor -->
-                  <AnchorActions
-                    {account} {config} anchor={pendingAnchor} safe={profile.org.safe}
-                    on:success={() => loadAnchors()} />
+          {#await Project.getCommits(project.urn, seed.api, project.head, getTimestampOneYearAgo()) then points}
```
cloudhead, line 78: I think we discussed this, but I'm not too happy with waiting for the commit data to load before we are able to show the widget. Can this be moved so that the widget renders without waiting?

src/App.svelte
```diff
@@ -16,6 +16,7 @@
   import Header from '@app/Header.svelte';
   import Loading from '@app/Loading.svelte';
   import Modal from '@app/Modal.svelte';
+  import LinearGradient from "@app/base/projects/LinearGradient.svelte";
```
cloudhead, line 19: If this is to be used as a global gradient, then maybe move it to `@app/LinearGradient.svelte`?

PR #213, opened 2022-02-25
Add link to code line numbers

sebastinez:
This PR adds the ability to pass a link to a `Browser` with a location hash pointing to a specific line in the document.
For example:
`/seeds/<host>/<projectId>/tree/<peerId>/<path>#L27`

There are other places where this eventually can come handy, but for now I just add the `hash` properties to the `browserStore` and work with the `Blob` component.

I'm seeing possibilities to link to files and lines of code in a specific commit detail view..

### DIFF
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -173,7 +173,7 @@
           {#if utils.isMarkdownPath(blob.path)}
             <Readme content={blob.content} />
           {:else}
-            <Blob {blob} />
+            <Blob line={browser.line} {blob} />
           {/if}
         {:catch}
           <div class="error error-message file-not-found">
--- a/src/base/projects/ProjectRoute.svelte
+++ b/src/base/projects/ProjectRoute.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import type { Writable } from 'svelte/store';
   import type { Config } from "@app/config";
+  import { formatLocationHash } from '@app/utils';
   import * as proj from '@app/project';
 
   import Project from './Project.svelte';
@@ -12,8 +13,10 @@
   export let content: proj.ProjectContent = proj.ProjectContent.Tree;
   export let project: proj.Project;
   export let config: Config;
+  export let hash: string | null = null;
 
-  const browse: proj.BrowseTo = { content, peer, path: "/" };
+  const line = formatLocationHash(hash);
+  const browse: proj.BrowseTo = { content, peer, path: "/", line };
   const head = project.branches[project.defaultBranch];
 
   // route is passed when the URL has more params after e.g. /tree or /history
--- a/src/project.ts
+++ b/src/project.ts
@@ -97,6 +97,7 @@ export interface Browser {
   revision: string | null;
   peer: string | null;
   path: string | null;
+  line: number | null;
 }
 
 export const browserStore = writable({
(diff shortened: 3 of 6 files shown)


src/project.ts
```diff
@@ -97,6 +97,7 @@ export interface Browser {
   revision: string | null;
   peer: string | null;
   path: string | null;
+  hash: string | null;
```
cloudhead, line 100: Isn't it better to store `line: number | null`?

src/base/projects/Browser.svelte
```diff
@@ -173,7 +173,7 @@
           {#if utils.isMarkdownPath(blob.path)}
             <Readme content={blob.content} />
           {:else}
-            <Blob {blob} />
+            <Blob line={browser.hash?.slice(2)} {blob} />
```
cloudhead, line 176: we should check what we're parsing, this will pass in a line even with `#abcd`

src/base/projects/ProjectRoute.svelte
```diff
@@ -12,8 +12,9 @@
   export let content: proj.ProjectContent = proj.ProjectContent.Tree;
   export let project: proj.Project;
   export let config: Config;
+  export let hash: string | null = null;
 
-  const browse: proj.BrowseTo = { content, peer, path: "/" };
+  const browse: proj.BrowseTo = { content, peer, path: "/", hash };
```
cloudhead, line 17: We should parse the hash here into a line number

src/project.ts
sebastinez: I didn't want to commit myself to `line: number | null` since we eventually could use the location hash for pointing to a commit diff (similar to github using e.g. https://github.com/radicle-dev/radicle-client-services/pull/122/files#diff-0b5c61cb7e37875016664c483ddc7eb6947e7019b710684c9ffc956a79a80491) or other places

src/base/projects/ProjectRoute.svelte
sebastinez: Same here, eventually it wont always be a line number but could be a diff hash or another type of pointer

src/base/projects/Browser.svelte
sebastinez: I could add a function that would do the slicing of the location hash here and  verify the result.
Although if you pass `#abcd` or anything else it just won't stop no harm done, not sure why we would need to parse and check the result

cloudhead: The point is that currently `Blob` needs to handle a string, which could be anything, ie. not a number, whereas `Blob` only cares about line numbers at the moment, so it should accept the smallest possible input space that the type-checker can check, which is `number | null`. As a rule of thumb, you want to convert unstructured data into structured data as soon as possible, ie.  at the edges, to keep most of your code dealing with structured/typed data.

src/project.ts
cloudhead: When that happens we can turn the type into an enum or something, eg. `{ line: number } | { diff: ... } | null`. But you're currently mixing domains: revision, peer and path are all meaningful data about a code resource, and then you have `hash` which has to do with how web browsers work. These fundamentally don't belong in the same type.

src/base/projects/Browser.svelte
sebastinez: Ok awesome! Understood, will provide a fix for it

src/project.ts
sebastinez: Ok got it, yeah makes sense 👍

cloudhead (approved):
Great!

PR #215, opened 2022-02-25
Highlight entire code line

sebastinez:
Just working in desktop for now `width > 960`.
Would work even better if `radicle_source` would provide us with each line instead of a blob.


cloudhead:
Rebased and merged in 49169eeeeb31bd1f52fe919b41ee7e7a14b9af5c

PR #217, opened 2022-03-14
Provide new foundation for testing

sebastinez:
This PR improves the base layer for our testing suite, it does:

- Continue with the use of Cypress for component testing, and as of this PR also uses some Integration tests for more complex navigation states (e.g. the project source browsing)
- Implement the use of Vitest for unit testing single functions (utils functions for now, more later)

IMO we should review this PR carefully specially the way we implement the testing suite, since we will be adding tests as time goes on, and should have a good foundation without the need to rewrite the base.


src/config.ts
```diff
@@ -246,8 +251,7 @@ export async function getConfig(): Promise<Config> {
   if (metamask) {
     // If Metamask is detected, we use the network configured there.
     const ready = await checkMetaMask(metamask);
-    if (ready === null) throw new Error("Not able to connect to Metamask, check for multiple Web3 providers");
-    network = ready;
+    if (ready) network = ready;
```
sebastinez, line 254: Ok this is maybe a weird one, I started having issues without having Metamask installed just with Brave Wallets that this error has been thrown around..
So IMO we should be fine if we just check for `ready` and if something returns we can overwrite `network` and else we stay on `homestead`

src/config.ts
```diff
@@ -5,11 +5,16 @@ import type { TypedDataSigner } from '@ethersproject/abstract-signer';
 import SafeServiceClient from "@gnosis.pm/safe-service-client";
 import { Core } from '@self.id/core';
 import WalletConnect from "@walletconnect/client";
-import config from "@app/config.json";
+import configProd from "@app/config.json";
+import configTest from "@app/configTest.json";
 import { WalletConnectSigner } from "./WalletConnectSigner";
 
+const config = window.Cypress ? configTest : configProd;
```
sebastinez, line 12: Allows us to have a smaller `config.json` for test purposes.

tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "@tsconfig/svelte/tsconfig.json",
   "include": [
-    "src",
-    "cypress"
+    "src"
```
sebastinez, line 4: not necessary to compile the `cypress` folder

tsconfig.json
```diff
@@ -26,8 +25,8 @@
     "importsNotUsedAsValues": "error",
     "skipLibCheck": true,
     "paths": {
+      "@public/*": ["./public/*"],
```
sebastinez, line 28: Also here the `@test` folder was not that used by the codebase but had to call plenty the `public` folder

sebastinez:
I'm still working on some things in the CI that seem to freeze the pipeline, but feel free to review the dependencies and the way I lined the e2e, integration and unit tests up

cypress/support/index.ts
```diff
@@ -1,19 +1,50 @@
...
+  if (params[0].to === "0x00000000000c2e074ec69a0dfb2997ba6c7d2e1e") {
+    return "0x0000000000000000000000004976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41";
+  // Get ENS Attributes
+  } else if (params[0].to === "0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41" && params[0].data === "0x59d1d43c567c364804de7bbedb53f583e483f6b73513fd2f44299e281024e4719da0b332000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000066176617461720000000000000000000000000000000000000000000000000000") {
+    return "0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001f68747470733a2f2f636c6f7564686561642e696f2f6176617461722e706e6700";
+  // Get Org informations
+  } else if (params[0].to === "0x8152237402e0f194176154c3a6ea1eb99b611482" && params[0].data === "0x8da5cb5b") {
+    return "0x000000000000000000000000ceab094641905c209cc796fc8037dd9ecc87ca2f";
+  } else {
+    console.log("Not implementend");
```
cloudhead, line 47: Should this be removed?

scripts/test
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh
 
-npm test && npx nyc report
+npm run test:unit
+npm run test:ct-unit
+npm start & wait-on http://localhost:3000
+npm run test:e2e
```
cloudhead, line 6: Is this just for us to run? Maybe it makes sense either way to add `set -e` at the top of the file, so it exits if any of the commands fail?

src/utils.ts
```diff
@@ -98,6 +98,9 @@ export function formatSeedId(id: string): string {
 }
 
 export function formatRadicleUrn(id: string): string {
+  // If the passed id has no prefix, add the default prefix
+  if (! id.startsWith("rad:git:")) id = "rad:git:" + id;
```
cloudhead, line 102: Maybe better to assert that it *does* start, otherwise the output will be strange anyway.

cloudhead: Or just `assert(isRadicleUrn(id))` if we have a function like that.

src/utils.ts
```diff
@@ -125,6 +128,7 @@ export function formatIpfsFile(ipfs: string | undefined): string | undefined {
   return undefined;
 }
 
+// If the string is less than 9 characters the entire string is returned.
```
cloudhead, line 131:
```suggestion
// If the string is less than 10 characters the entire string is returned.
```

src/configTest.json
```diff
@@ -0,0 +1,145 @@
+{
```
cloudhead, line 1: Can we put this in a test folder of sorts? Eg. in under `/test/` in the root?

src/base/projects/BranchSelector.svelte
```diff
@@ -19,7 +19,7 @@
 
   $: branchList = Object.keys(branches).sort().map(b => ({ key: b, value: b, badge: null }));
   $: showSelector = branchList.length > 1;
-  $: head = branches[project.defaultBranch];
+  $: head = project.head ?? branches[project.defaultBranch];
```
cloudhead, line 22: Is this currently a bug?

scripts/test
sebastinez: hmm the CI scripts already end the node process with a non zero exit code if any command fails, but we can add set -e

cypress/support/index.ts
sebastinez: This is needed for the mocking of the `request` function when `eth_call` is being called, we need to resolve it according to the params passed into.
It's not complete but has enough functionality to get a resolver, ENS attributes, Org information or just return an address as fallback.

cloudhead: I meant the `console.log` here :)

scripts/test
cloudhead: I'm not aware of such functionality, are you sure about this?

cypress/support/index.ts
sebastinez: > I meant the `console.log` here :)

Yeah sure! :)

scripts/test
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh
 
-npm test && npx nyc report
+npm run test:unit
+npm run test:ct-unit
+npm start & wait-on http://localhost:3000
```
cloudhead: I think here you mean `&&`?

cloudhead: I guess this file no longer exists?

cloudhead (approved):
Nice work.

scripts/test
sebastinez: I added `set -e` to all script files, since a successful command down the line can still return a 0 exit code..

scripts/test
sebastinez: > I guess this file no longer exists?

Exactly this file exists no more, I split the unit and integration tests.

PR #220, opened 2022-03-23
Change config signer during account change

sebastinez:
This is a fix that originated from an issue on Discord
https://discord.com/channels/841318878125490186/841342872082579466/956242980942008340

It seems that while changing accounts the signer property `config.signer` was used for `session.config.signer` instead of `config.metamask.signer`

This PR fixes this.

### DIFF
--- a/src/session.ts
+++ b/src/session.ts
@@ -2,7 +2,7 @@ import { get, writable, derived } from "svelte/store";
 import type { Readable } from "svelte/store";
 import type { BigNumber } from 'ethers';
 import type { TransactionReceipt, TransactionResponse } from '@ethersproject/providers';
-import { Config, getConfig, isMetamaskInstalled } from "@app/config";
+import { Config, isMetamaskInstalled } from "@app/config";
 import { Unreachable, assert, assertEq } from "@app/error";
 import type { TypedDataSigner } from '@ethersproject/abstract-signer';
 import type { WalletConnectSigner } from "./WalletConnectSigner";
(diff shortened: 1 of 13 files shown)


cloudhead:
What if we're using walletconnect though?

sebastinez:
> What if we're using walletconnect though?

Yep you're right, I think I found a better way, we can use the `metamaskSigner` if not `null` in `config.signer` during the constructor and if one uses walletconnect in the `getWalletConnectSigner` function we would overwrite `config.signer` then. And the faucet should still work.

sebastinez:
Ok it seems it's a bit more complicated than that, did some testing with WalletConnect, will provide a commit

src/session.ts
```diff
@@ -21,12 +21,23 @@ export type TxState =
...
-// TODO: Eventually there will be more reactive config needs, e.g. seeds
-export interface SessionConfig {
-  signer: ethers.Signer & TypedDataSigner | WalletConnectSigner | null;
-}
+// The token property isn't reactive, but reduces the passing down of the config object.
+export type SessionConfig = {
+  signer: Signer;
+  token: ethers.Contract;
+  type: SignerType;
+ };
```
cloudhead, line 40:
```suggestion
};
```

src/session.ts
```diff
@@ -21,12 +21,23 @@ export type TxState =
...
+}
+
 // Definitions made in `Config` that need to be part of the session,
 // to provide reactivity based on events i.e. accountsChanged
-// TODO: Eventually there will be more reactive config needs, e.g. seeds
-export interface SessionConfig {
-  signer: ethers.Signer & TypedDataSigner | WalletConnectSigner | null;
-}
+// The token property isn't reactive, but reduces the passing down of the config object.
+export type SessionConfig = {
```
cloudhead, line 36: Can this stay an `interface`?

src/session.ts
```diff
@@ -128,6 +139,7 @@ export const loadState = (initial: State): Store => {
           }
 
           try {
+            config.getWalletConnectSigner();
```
cloudhead, line 142: Can you add a comment explaining why we call this?

sebastinez: Done, added more comments in general, to make it clearer for the next time

src/session.ts
```diff
@@ -43,30 +56,40 @@ export interface Store extends Readable<State> {
...
   setTxSigning(): void;
   setTxPending(tx: TransactionResponse): void;
   setTxConfirmed(tx: TransactionReceipt): void;
-  setChangedAccount(address: string, config: Config): void;
+  setChangedAccount(address: string, signer: Signer): void;
 }
 
+// This is a buffer for the Config class to be available in this file, without making it a store just yet.
+// TODO: We should be thinking on making config a store, similar to session, to be more accesible.
+let sessionConfig: Config | null = null;
```
sebastinez, line 68: I accept this to be a not so nice workaround, but there are places like in the `window.ethereum` event handlers, were I really wanted to have a current `Config` instance.

This shows IMO that we should eventually move `Config` to be a store similar to `session` to allow us to be more easily accesible..

src/session.ts
```diff
@@ -126,7 +153,14 @@ export const loadState = (initial: State): Store => {
...
+              // When session_update address doesn't match the signer address, we update the signer and pass it to changeAccounts to update the session.
+              const signer = config.getWalletConnectSigner();
+              changeAccounts(accounts[0], signer);
+
+              // We update the address variable in scope of the connectWalletConnect function async here, to be used in the future when session_updates triggers.
+              signer.getAddress().then(a => address = a);
+            }
             // Check the current chainId, and request Metamask to change, or reload the window to get the correct chain.
             if (chainId !== config.network.chainId) {
               if (isMetamaskInstalled()) {
```
cloudhead, line 166: Should check signerType

cloudhead (approved):

PR #222, opened 2022-03-28
Verified commits

sebastinez:
This PR adds a verified badge to all the commits that have additional peer information, which means that the `http-api` was able to match the commit signature fingerprint to a key available on the seed.

**Things to fix:**
- [x] Fix test suite
- [x] Give some types better names 🤷‍♂️

### DIFF
--- a/src/commit.ts
+++ b/src/commit.ts
@@ -1,5 +1,6 @@
 import type { Stats } from "@app/project";
 import type { Diff } from "@app/diff";
+import { ApiError } from "@app/api";
 
 export interface CommitsHistory {
   headers: CommitMetadata[];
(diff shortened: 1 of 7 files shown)


cloudhead:
Looks good so far.

sebastinez:
Okay test suite works now and I think names make sense.

sebastinez:
Also moves the `navigateHistory` helper function from the `History` page to the `CommitTeaser`.
Allows us to remove one of div wrappers and just have one `div.commit` instead of two.

src/commit.ts
```diff
@@ -27,6 +32,20 @@ export interface GroupedCommitsHistory {
   stats: Stats;
 }
 
+export interface Person {
+  name: string;
+}
+
+export interface CommitContext {
+  committer: {
```
cloudhead, line 40: Should this be optional?

cloudhead (approved):

PR #231, opened 2022-04-21
Seeds SIWE implementation

sebastinez:
This PR adds the possibility to sign into a seed node with an Ethereum wallet.

**Authentication Process:**
- It sends a POST request to create an `Unauthorized` session
- With the returned `nonce` it creates a SIWE message and requests the wallet holder to sign in.
- The signed message and the raw message gets sent through a PUT request to the seed node
- If the verification on the seed node passes it updates the session to a `Authorized` session and returns the `session_id` and the session metadata.

Once authenticated the session ids get stored in `localStorage` and the currently active sessions get showed in the corresponding seed pages and the app header.

**TODOs left:**
- [x] Remove sessions from `localStorage` and the `session` store if expiration time is in the past.
- [x] Improve UI of session dropdown in the app header


src/base/seeds/View.svelte
```diff
@@ -106,10 +119,36 @@
...
+              <Address address={siweSession.address} {config} resolve />
+            </div>
+            <div class="mobile">
+              <Address address={siweSession.address} {config} compact resolve />
+            </div>
+          {:else}
+            <SiweConnect
+              {seed}
+              {config}
+              connection={siweSession ? Connection.Connected : Connection.Disconnected}
```
cloudhead, line 137: This will always be `false`

src/base/seeds/View.svelte
```diff
@@ -106,10 +119,36 @@
...
+            />
+          {/if}
+        {:else}
+          <div class="subtle">Please sign in to your wallet.</div>
```
cloudhead, line 141: I assume this is what you mean?
```suggestion
          <div class="subtle">Please connect your wallet.</div>
```

src/session.ts
```diff
@@ -293,6 +304,27 @@ export async function changeAccounts(address: string): Promise<void> {
   state.refreshBalance(config);
 }
 
+export function loadSeedSessions(): { [key: string]: SeedSession } {
+  const siweStorage = localStorage.getItem("siwe");
+
+  if (siweStorage) {
+    const siwe: { [key: string]: SeedSession } = JSON.parse(siweStorage);
+
+    // We only keep the sessions that are still valid and remove expired from localStorage
```
cloudhead, line 313:
```suggestion
    // We only keep the sessions that are still valid, and remove expired ones from `localStorage`.
```

src/session.ts
```diff
@@ -293,6 +304,27 @@ export async function changeAccounts(address: string): Promise<void> {
...
+    // For a session to be valid the expiration time has to be bigger or equal than the current Date.
```
cloudhead, line 314:
```suggestion
    // For a session to be valid the expiration time has to be bigger or equal than the current time.
```

src/siwe.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (! config.signer) {
+    return null;
+  }
+
+  const result = await createUnauthorizedSession(seed.api);
+  const address = await config.signer.getAddress();
+  const message = createSiweMessage(seed, address, result.nonce, config);
+  const signature = await config.signer.signMessage(message);
+
+  const auth: { id: string; session: SeedSession } = await new Request(`sessions/${result.id}`, seed.api).put({ message, signature: removePrefix(signature) });
```
cloudhead, line 55: Wrap this line?

src/siwe.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  await new Request(`sessions/${result.id}`, seed.api).get({});
```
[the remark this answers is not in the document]
sebastinez: At some point I was only returning, the nonce when updating a session, so I had to fetch the session object.
But can be removed now.

src/siwe.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface SeedSession {
+  domain: string;
+  address: string;
+  statement: string;
+  uri: string;
+  version: string;
+  chain_id: string;
+  nonce: string;
+  issued_at: number;
+  expiration_time: number;
```
cloudhead, line 18: Why are these snake_case?

PR #232, opened 2022-04-26
Add patch listing and detail view

sebastinez:
This PR adds a listing for all patches of a project and a detailed view for each patch with the related commit patch history.


src/utils.ts
```diff
@@ -15,6 +15,8 @@ import { ProfileType } from '@app/profile';
 import { parseUnits } from "@ethersproject/units";
 import { GetSafe } from "@app/base/orgs/Org";
 
+export const patchPrefix = "radicle-patch/";
```
rudolfs, line 18: 👀

rudolfs:
Nice! 👏 

Could we have an `Open in Upstream` button somewhere?

sebastinez:
> Nice! 👏
> 
> Could we have an `Open in Upstream` button somewhere?

Yeah sure you can, was on my todo list 😎

PR #244, opened 2022-06-22
Update issue state to new serialization

### LINKED ISSUE #242: Rewrite issue state due to new serialization in `radicle_common`
`radicle_common` has received a new way to serialize the issue state.
https://github.com/radicle-dev/radicle-cli/commit/2d723290455b98eb52a0ddb57c594847dda0d96d

At the moment `radicle-client-services` still needs to update the commit they are patching from `radicle_common` but once that happens we should change the issue state we're using from 
```
"state" | { closed: { reason: string } }
```
to
```
{ "status": "open" }
| {
  "status": "closed",
  "reason": "other"
}
```

sebastinez:
This PR closes #242

### DIFF
--- a/src/base/projects/Issue.svelte
+++ b/src/base/projects/Issue.svelte
@@ -90,7 +90,7 @@
     padding: 0.5rem 1rem;
     border-radius: 1.25rem;
   }
-  .opened {
+  .open {
     color: var(--color-positive);
     background-color: var(--color-positive-background);
   }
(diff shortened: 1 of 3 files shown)


src/base/projects/Issue.svelte
```diff
@@ -126,10 +125,10 @@
         </div>
         <div
           class="summary-state"
-          class:closed={issue.state !== "open"}
-          class:opened={issue.state == "open"}
+          class:closed={issue.state.status === "closed"}
+          class:opened={issue.state.status === "open"}
```
cloudhead, line 129: We're using `class:open` bellow, maybe we should do the same here?

sebastinez: done ✅

cloudhead (approved):

PR #246, opened 2022-06-27
Add patches

sebastinez:
This PR adds a patch listing and patch detail view.


sebastinez:
Still more design work pending

src/utils.ts
```diff
@@ -246,15 +246,42 @@ export function unixTime(): number {
...
+    year: 24 * 60 * 60 * 1000 * 365,
+    month: 24 * 60 * 60 * 1000 * 365/12,
+    day: 24 * 60 * 60 * 1000,
+    hour: 60 * 60 * 1000,
+    minute: 60 * 1000,
+    second: 1000
+  };
+
+  const rtf = new Intl.RelativeTimeFormat('en', { numeric: "auto", style: 'long' });
+  const elapsed = (timestamp * 1000) - current;
```
cloudhead, line 260: We should check if `current >= timestamp` and just return `"now"` in that case. We can then also drop the `Math.abs` usage.

src/cobs.ts
```diff
@@ -0,0 +1,20 @@
...
+  peer: PeerId;
+  identity?: PeerIdentity;
+}
+
+export interface PeerIdentity {
+  urn: string;
+  name: string;
+  ens?: {
+    name: string;
+  } | null;
```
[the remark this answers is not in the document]
sebastinez: This should only be `null` fixed in #288

src/utils.ts
sebastinez: Ok did rewrite this, the only thing we have to maintain is when calling `rtf.format` we have to pass it a negative number which I do by multiplying by `-1`, else we get `in 1 day` or similar output
While before doing `timestamp - current` we obtained the elapsed time with the right sign

src/diff.ts
```diff
@@ -90,3 +90,8 @@ export interface Diff {
   copied: string[];
   modified: FileDiff[];
 }
+
+export type EofNewLine =
+    "oldMissing"
+  | "newMissing"
+  | "bothMissing";
```
cloudhead, line 97: Maybe an enum is a bit nicer?

https://www.typescriptlang.org/docs/handbook/enums.html

sebastinez: Ah yes you are right, forgot about the string enums.. 👍 wanted to handle camelCase serialization that we eventually get from the http-api

cypress.config.cjs
```diff
@@ -4,6 +4,7 @@ const { startDevServer } = require('@cypress/vite-dev-server');
 
 module.exports = defineConfig({
   video: false,
+  defaultCommandTimeout: 10000,
```
sebastinez, line 7: Increased the default timeout for cypress integration tests

src/Avatar.svelte
```diff
@@ -21,9 +23,10 @@
     return avatar.toDataURL();
   }
 
-  $: if (isAddress(source) || isRadicleId(source)) {
+  if (isAddress(source) || isRadicleId(source) || isPeerId(source)) {
```
cloudhead, line 26: Didn't we have this as reactive because we switched from blockies to image avatars on load?

src/utils.ts
```diff
@@ -246,15 +246,47 @@ export function unixTime(): number {
...
+  }
+
+  for (const u in units) {
+    if (elapsed > units[u] || u == 'second') {
+      // We convert the division result to a negative number to get "XX [unit] ago"
+      return rtf.format(Math.round(elapsed / units[u]) * -1, u as Intl.RelativeTimeFormatUnit);
+    }
+  }
+
+  throw Error("Not able to create relative Datetime string");
```
cloudhead, line 277: Maybe instead of blowing up we just return `new Date(timestamp).toUTCString()`?

src/utils.ts
```diff
@@ -355,6 +387,13 @@ export function parseRadicleId(urn: string): string {
   return urn.replace(/^rad:[a-z]+:/, "");
 }
 
+// Get amount of days passed between two dates without including the end date
+export function getDaysPassed(from: Date, to: Date): number {
+  return Math.floor(
+    (to.getTime() - from.getTime()) / (24 * 60 * 60 * 1000)
+  );
+}
```
cloudhead, line 395: Is this used?

sebastinez: Yes, we use this function in the commit grouping by week here
https://github.com/radicle-dev/radicle-interface/blob/c2e52ddbea05496c0f3a13879d73068e2793d4f5/src/commit.ts#L167

Since we are using it only once, we could also move it into `commit.ts`

src/utils.ts
sebastinez: According to the function this error is more like an `Unreachable` but yes the returning of a UTCString is a softer exit.

src/Avatar.svelte
sebastinez: Yeah that was my thinking too, but the initial unresolved value is normally an `Address | URN | Peer ID` and we can check for that and generate the blocky during the `onMount` of `Avatar`.

When e.g. a profile has been resolved we update the `source` prop from the parent component to an image url, and the `Avatar` component updates, without needing reactivity itself.

The only way this could fail is if we go from e.g. an image URL back to a URN, which is not a realistic scenario or possible edge case IMO.

src/issue.ts
```diff
@@ -19,35 +25,15 @@ export type State = {
...
-export interface UnresolvedIdentity {
-  kind: "unresolved";
-  peer: PeerId;
-  urn: string;
-}
-
-export interface CommentWithReplies extends Comment {
-  replies: Comment[];
-}
+export type Discussion = Comment<Comment[]>;
```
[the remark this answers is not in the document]
sebastinez: I thought of `Discussion` as a single thread of one root comment and its replies.

That's the reason I replaced `CommentWithReplies[]` with `Discussion[]` in the `IIssue` interface, and IMO it's clearer to differentiate between `Comment` and `Discussion` than `CommentWithReplies`

Also the use of a generic type for the `replies` property allows us to eventually use a different type than a basic `Comment` there.

src/base/projects/Header.svelte
```diff
@@ -195,18 +198,27 @@
...
       0 issue(s)
     </div>
   {/await}
+  {#await Patch.getPatches(project.urn, seed.api) then patches}
+    <div class="stat patch-count clickable widget" class:active={content == ProjectContent.Patches} on:click={() => toggleContent(ProjectContent.Patches)}>
+      <strong>{groupPatches(patches).proposed.length}</strong> patche(s)
+    </div>
+  {:catch}
+    <div class="stat patch-count not-allowed widget" title="Not supported">
+      0 patche(s)
```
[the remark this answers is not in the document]
sebastinez: IMO, since we won't render other values that aren't `commit | commits`, `patch | patches`, `contributor | contributors` or `issue | issues` in the project header, maybe we could show them without the parenthesis, wdyt?

We wouldn't need to create something complicated to define plurals for unknown words.

src/base/projects/Patch.svelte
```diff
@@ -0,0 +1,139 @@
...
+      <Authorship noAvatar {config}
+        author={patch.author}
+        timestamp={patch.timestamp}
+        caption="opened" />
+    </header>
+    <PatchTabBar
+      {activeTab}
+      bind:revisionNumber={revisionNumber}
+      revisions={patch.revisions}
+      on:switchTab={onSwitch} />
```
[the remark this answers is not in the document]
sebastinez: Move async out of this component, so we can pass the latest revision into this component, and use a dispatch event for switching revision

src/base/projects/Patch/PatchFilter.svelte
```diff
@@ -0,0 +1,77 @@
+<script lang="ts">
+  import { groupPatches, Patch } from "@app/patches";
+import Placeholder from "@app/Placeholder.svelte";
+import { capitalize } from "@app/utils";
```
cloudhead, line 4: Indentation

src/base/projects/Patch/PatchFilter.svelte
```diff
@@ -0,0 +1,77 @@
+<script lang="ts">
+  import { groupPatches, Patch } from "@app/patches";
+import Placeholder from "@app/Placeholder.svelte";
```
sebastinez, line 3: indentation

src/base/projects/Patch/PatchTabBar.svelte
```diff
@@ -0,0 +1,83 @@
...
+    )}`;
+  };
+
+  const revisionList = Object.values(revisions).map((b, i) => ({
+    key: formatRevisionName(b, i),
+    value: i.toString(),
+    badge: null,
+  }));
+
+  revisionNumber = revisions.length - 1;
```
cloudhead, line 25: Set this on line 8

src/base/projects/Patch/PatchTimeline.svelte
```diff
@@ -0,0 +1,74 @@
...
+          {config} />
+      </div>
+    {:else if element.type === TimelineType.Review && element.inner.author.profile?.ens?.name}
+    <div class="margin-left">
+      <Review review={element.inner} {config} {getImage} />
+    </div>
+    {:else if element.type === TimelineType.Comment}
+    <div class="margin-left">
+      <Comment comment={element.inner} {config} {getImage} />
+    </div>
```
cloudhead, line 62: Can we indent all these divs under the branches?

src/base/projects/SourceBrowser/Changeset.svelte
```diff
@@ -22,82 +32,28 @@
...
-        <p>{path}</p>
-        <span class="diff-type created">created</span>
-      </div>
-      <div class="browse" on:click={() => dispatch("browse", path)}>
-        <Icon name="browse" width={20} inline fill />
-      </div>
-    </header>
+<div class="diff-listing">
+  {#each diff.created as file}
+      <FileDiff on:browse {file} mode="created" />
```
cloudhead, line 51: Indentation

src/base/projects/Patch/PatchTabBar.svelte
sebastinez: Will move this to a `Async` component

src/base/projects/Patch/PatchTeaser.svelte
```diff
@@ -0,0 +1,125 @@
...
+      timestamp={patch.timestamp} />
+  </div>
+  {#if commentCount > 0}
+  <div class="column-right">
+    <div class="comment-count">
+      <span class="text-xsmall emoji">💬</span>
+      <span>{commentCount}</span>
+    </div>
+  </div>
+{/if}
```
cloudhead, line 124: Indentation

src/base/projects/Patch/PatchTabBar.svelte
```diff
@@ -0,0 +1,95 @@
...
+        class:tab={revisions.length > 1}
+        class="text-small revision-toggle"
+        disabled={revisions.length <= 1}
+        on:click={() => showSelector = !showSelector}>
+        {formatRevisionName(revisions[revisionNumber], revisionNumber)}
+      </button>
+      <Dropdown
+        items={revisionList} selected={revisionNumber.toString()} visible={showSelector}
+        on:select={onRevisionChange} />
+      </div>
```
cloudhead, line 93: Indentation

PR #285, opened 2022-06-30
Unify Button heights

### LINKED ISSUE #257: Make border white on hover of "connect" button
This looks a bit muddy because the border is a light grey. Would be more crisp if the border was white on hover.
 image

### LINKED ISSUE #272: Unify button heights
There should be two different button heights. Currently there are the following button heights:

59px
image

43px
image

34px
image

32px
image

---

I'd suggest going with 32px and 40px.

brandonhaslegs:
- Added a variable for `--button-regular-height`
- Added a variable for `--button-small-height`
- Added a variable for `--button-tiny-height`
- Changed the badge size to be 1.5rem (height wasn't previously defined, so it was falling between 23 and 24, now it's a more crispy solid 24 height)
- rename "small" to "regular"
- rename "tiny" to "small"
- fix the grey border on the white button hover
- add new size for badges that is "tiny"

Fixes #257 
Fixes #272

Signed-off-by: Brandon Oxendine <brandon.oxendine@gmail.com>

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -74,12 +74,20 @@
 	--font-family-monospace: monospace;
 	--font-weight-medium: 600;
 	--font-weight-bold: 700;
+
 	--border-radius: 50px;
 	--border-radius-medium: 0.25rem;
+
 	--box-shadow-color: var(--color-secondary-2);
+
 	--content-max-width: 1920px;
 	--content-min-width: 480px;
+
 	--scrollbar-width: 0.5rem;
+
+	--button-regular-height: 2.5rem;
+	--button-small-height: 2rem;
+	--button-tiny-height: 1.5rem;
 }
 
 body {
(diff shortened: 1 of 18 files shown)


cloudhead:
There's a number of issues with these style changes, but since this is still a draft I'll wait before commenting :)

cloudhead:
Oh, nothing wrong with the approach, just some rendering issues (only checked Firefox):

alignment:
image

alignment:
image

Button is too small v (switches to big when connected, which is the right size)
image

border color:
image

brandonhaslegs:
Ah I thought I fixed all this. Will do for suuuure. Maybe I just haven't pushed something.

cloudhead:
Ok there's just one last alignment issue on FF I found:

image

cloudhead:
Can we reduce the `regular` button height some? It just looks way too big atm at 48px. Perhaps we can do 2.5rem or 2.75rem?

PR #288, opened 2022-07-12
Improve handling unresolved identities

sebastinez:
This PR improves the `Author` type used in `Issue` and soon in `Patches`, and fixes handling unresolved authors.

### DIFF
--- a/src/base/projects/Issue/IssueComment.svelte
+++ b/src/base/projects/Issue/IssueComment.svelte
@@ -18,10 +18,19 @@
   let profile: Profile | null = null;
 
   onMount(async () => {
-    if (comment.author.identity.ens?.name) {
+    if (comment.author.kind === "resolved" && comment.author.identity.ens?.name) {
       profile = await Profile.get(comment.author.identity.ens.name, ProfileType.Minimal, config);
     }
   });
+
+  $: source = profile?.avatar ||
+    (comment.author.kind === "resolved"
+      ? comment.author.identity.urn
+      : comment.author.urn);
+  $: title = profile?.name ||
+    (comment.author.kind === "resolved"
+      ? comment.author.identity.name
+      : comment.author.urn);
 
   const selectReaction = (event: { detail: string }) => {
     // TODO: Once we allow adding reactions through the http-api, we should call it here.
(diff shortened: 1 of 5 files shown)


src/issue.ts
```diff
@@ -26,14 +26,22 @@ export interface Comment {
   timestamp: number;
 }
 
-export interface Author {
+// This generic types allow us to create a closed union type that allows us to create exclusive or type unions.
+// https://github.com/Microsoft/TypeScript/issues/14094
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+type XOR<T, U> = (T | U) extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
+
+export type Author = XOR<Omit<Identity, "identity">, Omit<Identity, "urn">>;
```
cloudhead, line 34: This seems like a really complex way of creating a sum type. What are you trying to achieve?

src/base/projects/Issue/IssueComment.svelte
```diff
@@ -18,11 +18,14 @@
...
 
   onMount(async () => {
-    if (comment.author.identity.ens?.name) {
+    if (comment.author.identity?.ens?.name) {
       profile = await Profile.get(comment.author.identity.ens.name, ProfileType.Minimal, config);
     }
   });
 
+  $: source = profile?.avatar || comment.author.identity?.urn || comment.author.urn || comment.author.peer;
+  $: title = profile?.name || comment.author.identity?.ens?.name || comment.author.urn || comment.author.peer;
```
cloudhead, line 27: Which part here is reactive? ie. what do we expect to be updated after the component is loaded and why?

src/issue.ts
sebastinez: Yeah I went down a rabbit hole to create an exclusive union type, but adding a type guard with an `Identity` enum similar to how it's handled in the http-api seems nice.

src/base/projects/Issue/IssueComment.svelte
sebastinez: The `Profile.get` function is being called `onMount` so there is eventually profile data we want to use as the `source` and `title` of the user avatar.

src/issue.ts
cloudhead: You can create a sum type by simply tagging it though, eg. `{ kind: "resolved", name: string } | { kind: "unresolved": peer: string }`

cloudhead: This should make it "exclusive" since the tag is required to be different on both sides.

src/base/projects/Issue/IssueComment.svelte
```diff
@@ -18,11 +18,22 @@
...
+    if (comment.author.kind === "resolved" && comment.author.identity.ens?.name) {
       profile = await Profile.get(comment.author.identity.ens.name, ProfileType.Minimal, config);
     }
   });
 
+  $: source =
+    profile?.avatar ||
+    (comment.author.kind === "resolved"
+      ? comment.author.identity.urn
+      : comment.author.urn);
```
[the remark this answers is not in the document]
cloudhead: Ah I see why. I would move the `URN` to the same place for both identity types to avoid this kind of issue.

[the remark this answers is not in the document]
sebastinez: Strictly we don't need it to be resolved but since it could be either "resolved" or "unresolved", we need to check both variants to get always a `string` for `source`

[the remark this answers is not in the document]
sebastinez: > Ah I see why. I would move the `URN` to the same place for both identity types to avoid this kind of issue.

👌  this would also work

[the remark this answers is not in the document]
sebastinez: @cloudhead Should we move the `URN` maybe on the http-api? Seems like it's easier to handle it there..

Maybe we could do something like this
```rs
pub enum Identity {
    /// Only the URN is known.
    Unresolved { urn: Urn },
    /// The full user identity is resolved.
    Resolved { urn: Urn, identity: project::PeerIdentity },
}
```

cloudhead (approved):

PR #291, opened 2022-07-15
Close Floating Modal when clicking Clipboard and style improvements

sebastinez:
### Changeset
- Adds `git clone ...` to the second input field in the Clone popover.
- Closes clone popover when Clipboard icon gets clicked.
-  Fix the copy icon gradient overlay (it overlaps its container)

### DIFF
--- a/src/Clipboard.svelte
+++ b/src/Clipboard.svelte
@@ -1,6 +1,14 @@
 <script lang="ts">
   import Icon from '@app/Icon.svelte';
   import { toClipboard } from '@app/utils';
+  import { createEventDispatcher } from "svelte";
+
+  const dispatch = createEventDispatcher();
+
+  const copy = () => {
+    toClipboard(text);
+    dispatch("copied");
+  };
 
   export let text: string;
   export let small = false;
@@ -30,7 +38,7 @@
   }
 </style>
 
-<span class="clipboard" class:small on:click|stopPropagation={() => toClipboard(text)}>
+<span class="clipboard" class:small on:click|stopPropagation={copy}>
   {#if small}
     <Icon name="clipboard-small" width={12} height={12} />
   {:else}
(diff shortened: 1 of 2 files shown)


brandonhaslegs:
Cool, lookin good! It's kinda weird that the icon flashes green when you click it. Maybe not necessary @cloudhead? Also is there a notification component in the app? Would be good to have a little notification that says "Copied to clipboard" or something.

brandonhaslegs:
"Use Git to clone this repository from the URL above." doesn't make sense anymore btw. Should say something else. Suggestion: "Use Git to clone this repository"

src/Clipboard.svelte
```diff
@@ -1,15 +1,23 @@
...
   export let text: string;
   export let small = false;
 </script>
 
 <style>
   .clipboard {
-    width: 32px;
-    height: 32px;
+    width: 30px;
+    height: 30px;
```
cloudhead, line 20: What's this about? I try to stick to multiples of 1rem (16px)

sebastinez: Yeah, the issue here is that the `input` field has a height of 30px so when put the 32px high span element on top of it, the span element overlaps on both sides of the `input` field.
Bildschirmfoto 2022-08-17 um 07 55 50

By reducing the span height to 30px it matches the height of the input field.

sebastinez: Hmm or we could also increase the input field from 30px to 2rem 🤔

cloudhead: Yeah the input field may aswell be rem-aligned since it's so close anyway. Can we fix this in a different PR and leave the button as-is for now?

sebastinez: > Yeah the input field may aswell be rem-aligned since it's so close anyway. Can we fix this in a different PR and leave the button as-is for now?

Ok yeah no problem

cloudhead (approved):

src/Clipboard.svelte
sebastinez: @cloudhead sizes have been reverted to multiples of 1rem in [904f58a](https://github.com/radicle-dev/radicle-interface/pull/291/commits/904f58adc23d21c93ce3143fddd767208b33856b)

PR #294, opened 2022-07-18
Add project detail loader and `View file` tooltip

### LINKED ISSUE #267: Add tooltip to codeblock icon link
This links to something but I'm not clear on where it goes. Also it's quite unexpected because you can click everywhere else on the header of the codeblock to collapse the codeblock, except on the code icon which I guess links to the file that codeblock is from. Some text next to the icon would be ideal. For example "<> View file"

image

### LINKED ISSUE #268: Add loader to content on project pages
There is sometimes not a loader on the actual project page content (files, commits, etc). So once the header loads, the rest of the page sometimes takes a few seconds longer, but there's no loader so it feels broken.

image

sebastinez:
This PR adds a `<Loading>` indicator to the project view to show the user that something is happening.
Adds `View file` next to the browse icon in the `CommitTeaser`, `Changeset` and `FileDiff` components

Closes #268 
Closes #267

### DIFF
--- a/src/base/projects/SourceBrowser/FileDiff.svelte
+++ b/src/base/projects/SourceBrowser/FileDiff.svelte
@@ -146,7 +146,7 @@
       {/if}
     </div>
     <div class="browse clickable" on:click|stopPropagation={() => dispatch("browse", file.path)}>
-      <Icon name="browse" width={20} inline fill />
+      <Icon name="browse" title="View file" width={20} inline fill />
     </div>
   </header>
   {#if !collapsed}
--- a/src/base/projects/Commit/CommitTeaser.svelte
+++ b/src/base/projects/Commit/CommitTeaser.svelte
@@ -96,7 +96,7 @@
   <div class="column-right">
     <CommitVerifiedBadge {commit} />
     <span class="secondary hash">{formatCommit(commit.header.sha1)}</span>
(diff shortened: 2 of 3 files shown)


rudolfs:
@sebastinez you might want to adjust this PR, now that we have the `Floating` component merged: https://github.com/radicle-dev/radicle-interface/pull/313.

sebastinez:
> @sebastinez you might want to adjust this PR, now that we have the `Floating` component merged: #313.

Congrats on the merge, will remove the handleClickOutside behaviour. 👍

src/base/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -147,6 +155,7 @@
     </div>
     <div class="browse clickable" on:click|stopPropagation={() => dispatch("browse", file.path)}>
       <Icon name="browse" width={20} inline fill />
+      <span class="secondary view text-xsmall">View file</span>
```
cloudhead, line 158: This should be an actual tooltip, right now it always shows up. Perhaps just `title="View file"` would work

sebastinez: Yeah I went with the solution brandon wrote down
> Some text next to the icon would be ideal. For example "<> View file"

But showing the text as tooltip leaves the design more lean.

PR #296, opened 2022-07-28
Update Vite to 3.0

sebastinez:
This PR updates Vite to 3.0

Since Vite now removed more Node Core modules polyfills, we need to replace `sanitize-html` with `xss` which works without `postcss` and it's node dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,28 +23,32 @@
         "@walletconnect/utils": "^1.7.8",
         "buffer": "^6.0.3",
         "caip": "^0.9.2",
+        "connect-history-api-fallback": "^2.0.0",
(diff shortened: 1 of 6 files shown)


cloudhead:
Looks good, but CI needs to pass.

sebastinez:
> Looks good, but CI needs to pass.

Yeah I'm working on that 😊 seems there is an issue with vite dev server.. will ping you if it works

sebastinez:
Okay, CI is passing:
- Had to remove `vite-plugin-rewrite-all` in favor of our own implementation in `vite.config.ts`
- And polyfill Node events package, since readable-streams in the keccak package uses it, and Vite 3 supports no more Node packages.

PR #297, opened 2022-07-29
Improve feedback/accessibility

sebastinez:
**Card**
- Allows navigation by keyboard
- Gives user feedback depending on the current state.


sebastinez:
Time past real quick, will close this PR for now, and take another approach in the near future

Issue #256, opened 2022-06-27
Add themeing (light/dark mode)


brandonhaslegs:
The app is just in dark mode now, but it's pretty standard to have light and dark mode in apps. We buit the design system to handle this really easily, so could just use that.

sebastinez:
I fully agree on the having two themes (dark/light), not sure on the design system part, since we always worked on a very own styling. Maybe something you could talk about with @cloudhead?

brandonhaslegs:
Oh I just meant we could grab the way we did theming from the design system in Upstream maybe.

cloudhead:
Yeah, I think this is very low priority but would be nice some day!

brandonhaslegs:
Of course. I bet it'd be pretty easy though. @efstajas has done it a few times on various web apps.

PR #298, opened 2022-07-30
Add theme switching button

rudolfs:
We add a theme switching button to the top right corner of the app which lets us switch between a dark and a light theme.
For this, we introduce a new color scheme with 6 colors, a foreground scale and 1 background color. Each of the colors and the foreground scale have 6 levels of tints.

Refs: https://github.com/radicle-dev/radicle-interface/issues/256.
Requires: https://github.com/radicle-dev/radicle-client-services/pull/183

Screenshot 2022-09-01 at 14 49 06

Screenshot 2022-09-01 at 14 49 11

Screenshot 2022-09-01 at 14 49 01

Screenshot 2022-09-01 at 14 49 17


rudolfs:
@brandonhaslegs since colors are your expertise, I could use a bit of help here.

Particularly the source browsing screen:
Screenshot 2022-08-02 at 11 16 07

brandonhaslegs:
image

image

brandonhaslegs:
I think everything looks okay. Need to clean up the theming. Not sure what all the color levels really mean, and could probably be organized a bit better.

src/ThemeToggle.svelte
```diff
@@ -0,0 +1,187 @@
+<script lang="ts">
+  import { onMount } from 'svelte';
+
+  let theme: "light" | "dark" | undefined = undefined;
```
sebastinez, line 4:
```suggestion
  let theme: Theme | null = null;
  
  enum Theme {
    Light = "light",
    Dark = "dark"
  }
```

I suggest we keep using string enums here, similar to other places in the web ui, and we prefer to use `null` over `undefined`

src/ThemeToggle.svelte
```diff
@@ -0,0 +1,187 @@
...
+
+<div
+  class="theme-button"
+  role="button"
+  on:click={() => {
+    theme = theme === "dark" ? "light" : "dark";
+  }}
+>
+  {#if theme === "dark"}
+    <svg
```
sebastinez, line 45: I would propose to move the `dark` and `light` svgs to the `Icon` component, to keep all the svgs there when possible

src/ThemeToggle.svelte
```diff
@@ -0,0 +1,187 @@
+<script lang="ts">
+  import { onMount } from 'svelte';
+
+  let theme: "light" | "dark" | undefined = undefined;
+
+  onMount(() => {
+    let storedTheme = window.localStorage.getItem('theme');
```
sebastinez, line 7: IMO we should move the page theme to the `session` store and have some store functions there to get and set the current theme.

src/Header.svelte
```diff
@@ -227,6 +228,7 @@
     <div class="toggle" on:click={toggleNavbar}>
       <Icon name="ellipsis" width={27} height={27} fill />
     </div>
+    <ThemeToggle />
```
sebastinez, line 231: We should be able to pass the current theme to `ThemeToggle` as a prop e.g. `<ThemeToggle theme={$session.theme} />`

sebastinez (commented):
Thanks @rudolfs! really liking the work on a light theme, left some comments.

src/ThemeToggle.svelte
```diff
@@ -0,0 +1,187 @@
...
+        theme = "dark";
+      } else {
+        theme = "light";
+      }
+    } else if (storedTheme === "light" || storedTheme === "dark") {
+      theme = storedTheme;
+    }
+  });
+
+  $: if (theme) {
```
sebastinez, line 21: Also if we can define `theme` in the `session` store from `localStorage` or by `window.matchMedia`, then theme would never be `null`

src/ThemeToggle.svelte
rudolfs: Happy to oblige: ec07fc3fe2e9b0be0f279f65eab5798bafd5fca6.

Just a couple of questions for my own understanding:

1) Is it just a convention or are there actual benefits for using enums? I remember discussing this before in Upstream and I think the outcome was that string literal types had all the same benefits of enums _and_ were simpler and less verbose.

2) What's the thinking behind using `null`? From my point it also makes stuff more verbose. For example, we can't use the optional chaining operator `?.` with component props:

```
// component.svelte
<script>
  export let myFancyProp: string | undefined = undefined;
</script>

// using the component

<Component myFancyProp={someObject?.maybeEmptyKey}>
```

vs

```
// component.svelte
<script>
  export let myFancyProp: string | null = null;
</script>

// using the component

<Component myFancyProp={someObject?.maybeEmptyKey || null}>
```

src/ThemeToggle.svelte
rudolfs: ec07fc3fe2e9b0be0f279f65eab5798bafd5fca6

src/ThemeToggle.svelte
rudolfs: Couldn't move it to `session`, because it's only available when you've connected your wallet.
How about this: 2b1adbfae38338e4f7bd93457e16f40ab2eb8ae5? Could also get a second opinion @cloudhead?

src/Header.svelte
rudolfs: 2b1adbfae38338e4f7bd93457e16f40ab2eb8ae5

src/ThemeToggle.svelte
rudolfs: 2b1adbfae38338e4f7bd93457e16f40ab2eb8ae5

rudolfs:
@brandonhaslegs: I've cleaned up the color system, we now have 8 different colors and 6 different levels of shades for each. There's also a palette component that you can pull up anywhere in the app by pressing the hotkey `d`. You can also toggle a checkered background by clicking anywhere within the palette to better inspect transparencies. Hovering over a color swatch shows you the name as a tooltip. I tried to keep the dark palette as close to the original as possible. Could you have another look at the light palette? It might be easier now to understand what's going on and tweak it.

Screenshot 2022-08-22 at 17 17 37

Screenshot 2022-08-22 at 17 17 29

Screenshot 2022-08-22 at 17 17 51

Screenshot 2022-08-22 at 17 17 55

rudolfs:
> What's up with the dark yellow color in level 2 here?

I think the whole yellow color needs to be rethought in light mode. Otherwise it's too light. Have a look at the `Clone` button as it is now:
Screenshot 2022-08-22 at 19 36 29
 
Screenshot 2022-08-22 at 19 36 25

vs when the yellow colors are all in the proper order:
Screenshot 2022-08-22 at 19 37 26

brandonhaslegs:
The code colors look wrong still because they are hard coded? Or I don't know how they're done really
image

brandonhaslegs:
Also the drop shadows would be nice to be able to control them also on a per-theme basis.

brandonhaslegs:
Also need a hover state (exactly the same as the "Connect" button)

rudolfs:
> The code colors look wrong still because they are hard coded? Or I don't know how they're done really

They're done in the backend. I think that could potentially be a separate PR.

brandonhaslegs:
> They're done in the backend. I think that could potentially be a separate PR.

The code is almost invisible in light mode. Seems important.

rudolfs:
> Also the drop shadows would be nice to be able to control them differently in different themes.

Now it's possible, have a look at the `elevations.css` file.

rudolfs:
@brandonhaslegs 

The copy button looks broken now: 
Screenshot 2022-08-24 at 16 12 37

vs

Screenshot 2022-08-24 at 16 12 46

The hover states of the projects on the index page could use more contrast in light mode:
Screenshot 2022-08-24 at 16 27 31

rudolfs:
@brandonhaslegs could you go through these screens and check that the app looks more or less like the original? I normally open 2 tabs and then switch between them to see the differences:

  - http://127.0.0.1:3000/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/patches/hnrkyr8zhrj81jpiwu3ihs48sdsd3e1hgeipo
  - https://app.radicle.xyz/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/patches/hnrkyr8zhrj81jpiwu3ihs48sdsd3e1hgeipo
-----------------
- http://127.0.0.1:3000/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/patches/hnrkyr8zhrj81jpiwu3ihs48sdsd3e1hgeipo
- https://app.radicle.xyz/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/patches/hnrkyr8zhrj81jpiwu3ihs48sdsd3e1hgeipo

-----------------
- http://127.0.0.1:3000/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/remotes/hyn9diwfnytahjq8u3iw63h9jte1ydcatxax3saymwdxqu1zo645pe/history/a9a377866d6dca352f78c0f3c6e3bd099e089313
- https://app.radicle.xyz/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/remotes/hyn9diwfnytahjq8u3iw63h9jte1ydcatxax3saymwdxqu1zo645pe/history/a9a377866d6dca352f78c0f3c6e3bd099e089313
-----------------

- http://127.0.0.1:3000/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/remotes/hyn9diwfnytahjq8u3iw63h9jte1ydcatxax3saymwdxqu1zo645pe/patches
- https://app.radicle.xyz/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/remotes/hyn9diwfnytahjq8u3iw63h9jte1ydcatxax3saymwdxqu1zo645pe/patches
-----------------

- http://127.0.0.1:3000/cloudhead.radicle.eth
- https://app.radicle.xyz/cloudhead.radicle.eth
-----------------

- http://127.0.0.1:3000/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkjajuucc6zp5eknt3s9xykqsrus44cjimy/commits/0cced197baef052aa2dae7ef5acc3bc5ac27a7d0
- https://app.radicle.xyz/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkjajuucc6zp5eknt3s9xykqsrus44cjimy/commits/0cced197baef052aa2dae7ef5acc3bc5ac27a7d0

rudolfs:
http://127.0.0.1:3000/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkjajuucc6zp5eknt3s9xykqsrus44cjimy/commits/0cced197baef052aa2dae7ef5acc3bc5ac27a7d0

The `created` badge in light mode looks a bit too low-contrast:

Screenshot 2022-08-24 at 16 36 40

rudolfs:
> The code is almost invisible in light mode. Seems important.

@brandonhaslegs was relatively easy https://github.com/radicle-dev/radicle-client-services/pull/183.

brandonhaslegs:
@rudolfs Fixed a bunch of things. Should be good to go, I think!

src/Header.svelte
rudolfs: On a second thought. I think it doesn't make sense to mix the appearance state with the wallet state that we call "session". The two are completely independent and should be in separate modules. In my opinion it's best to keep the appearance state in a module within the theme component, this way it's all in one place and the state can still be imported by other modules if needed.

src/Badge.svelte
```diff
@@ -12,20 +12,21 @@
...
-    color: var(--color-foreground-faded);
-    background: var(--color-foreground-background);
+    color: var(--color-foreground-6);
+    background: var(--color-foreground-2);
   }
   .primary {
     color: var(--color-primary);
-    background: var(--color-primary-background);
+    background: linear-gradient(var(--color-primary-2),var(--color-primary-2)),
+    linear-gradient(var(--color-background), var(--color-background));
```
rudolfs, line 21: @brandonhaslegs wonder if there's another way to get this work without having to mix the colors? I assume you already tried to fiddle with the actual contrasts of the colors to get it right?

brandonhaslegs: I suppose we could flatten all the colors… The only reason I'm doing this here is because in this case, two colors are stacked, and since they're transparencies, they mix and look muddy when stacked. Normally I wouldn't stack colors, but that's just how the design was in this case. Keeping the color levels as transparencies is easy for me, but I guess we could just flatten them… I wish you could stack colors without using a css gradient… it certainly looks dumb…

sebastinez:
Hey I didn't want to mess with the design, since you guys have a much better eye for it.

But one thing that I wanted to mention, that bothered me when using the light theme is that the color contrast in some components between text and background is very low, and I think in particular for people with color blindness or issues to read e.g. gray on a light gray we could do better.

**Component reaction when hovering much less visible than in the dark theme**

https://user-images.githubusercontent.com/7912302/187609104-cb2e2446-37f5-47dd-80be-17922c77a594.mov

**Code blobs is hard on the eyes to read on light theme**

code_blob_dark
code_blob_light

**Patch status, green on light green is hard to read on light theme**

patch_status_dark
patch_status_light

**Project header `0 issue(s)` and `0 patche(s)` reads much better on dark theme** 

project_header_Dark
project_header_light

sebastinez (commented):
Nice work! Looking forward to using the light theme at some point.
I left one comment on improving the color contrast for readability

rudolfs:
@sebastinez thanks for the feedback!! Good points about the contrast, @brandonhaslegs could you have a look at that?

> Code blobs is hard on the eyes to read on light theme
Maybe we could use e.g. #6e7781 which is a much darker gray like GitHub uses for comments in the light theme

I think this will get solved once https://github.com/radicle-dev/radicle-client-services/pull/183 is merged.

src/Badge.svelte
rudolfs: Hm, good point. Yeah, I agree, ideally we shouldn't use transparencies, so we can layer stuff better. I previously tried to convert everything to HSL, but there wasn't an obvious pattern, because we'd have to mix in some of the background into each of the colors. I'll check if there's another way to programmatically achieve this, so that we don't have to flatten them every time by hand.

brandonhaslegs: I can flatten them by hand, but that seems like an overreaction to this one little "problem" of using a css gradient to layer color styles.

brandonhaslegs:
> patch_status_light

How can I test this? I changed the colors to 
```
    color: var(--color-positive-6);
    background-color: var(--color-positive-1);
```

I think that should work

rudolfs:
@brandonhaslegs 

http://localhost:3000/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/patches/hnrkyr8zhrj81jpiwu3ihs48sdsd3e1hgeipo

Screenshot 2022-09-01 at 09 43 29

Screenshot 2022-09-01 at 09 43 25

sebastinez:
> @brandonhaslegs
> 
> http://localhost:3000/cloudhead.radicle.eth/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/patches/hnrkyr8zhrj81jpiwu3ihs48sdsd3e1hgeipo
> 
> Screenshot 2022-09-01 at 09 43 29
> 
> Screenshot 2022-09-01 at 09 43 25

Nice much better 👍

src/Badge.svelte
rudolfs: Yeah, you're probably right. Let's punt on this for now.

rudolfs:
@brandonhaslegs I feel the `--color-tertiary-6` is a bit too intense in the dark theme:

Now:
Screenshot 2022-09-01 at 10 54 57

vs original

Screenshot 2022-09-01 at 10 55 29

brandonhaslegs:
> @brandonhaslegs I feel the `--color-tertiary-6` is a bit too intense in the dark theme:

True, this is a mistake. I mean I did it on purpose, but didn't realize the effect it would have. I'll fix.

rudolfs:
@sebastinez how about this for the hover states?


https://user-images.githubusercontent.com/158411/187885502-cadeec5d-64ca-4c03-9197-9f504d359a3a.mov

sebastinez:
> @sebastinez how about this for the hover states?

Looks much better! 👏

rudolfs:
Btw. we have now the option to use the Solarized source code highlighting theme:

```diff
diff --git a/src/base/projects/Browser.svelte b/src/base/projects/Browser.svelte
index d01acea..6f57836 100644
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -53,7 +53,7 @@
             path,
             isMarkdownPath
               ? { highlight: false }
-              : { highlight: true, theme: `base16-ocean.${theme}` },
+              : { highlight: true, theme: `Solarized (${theme})` },
           );

     state = { status: Status.Loading, path };
diff --git a/src/project.ts b/src/project.ts
index b8f6fab..16dc5bb 100644
--- a/src/project.ts
+++ b/src/project.ts
@@ -403,7 +403,10 @@ export class Project implements ProjectInfo {
     path: string,
     options:
       | { highlight: false }
-      | { highlight: true; theme: "base16-ocean.dark" | "base16-ocean.light" },
+      | {
+          highlight: true;
+          theme: "base16-eighties.dark" | "base16-eighties.light";
+        },
   ): Promise<Blob> {
     return new Request(
       `projects/${this.urn}/blob/${commit}/${path}`,
```

Screenshot 2022-09-02 at 12 19 55
Screenshot 2022-09-02 at 12 19 58

cloudhead:
Going through this again, will post any issues I notice. Here's one, where the yellows are different between the button and the text fields:

image

cloudhead:
This became really dark and hard to see:
image

cloudhead:
* [ ] Header gradient looks gone also.
* [ ] The frontmatter could be bumped up a notch:
image
* [ ] Delegate badge looks weird:
image

cloudhead:
* [ ] This cyan looks off to me: 
image

I'd prefer if we kept close to the old one.

* [ ] Some backgrounds are too subtle, eg. in the commit listing it's barely visible:
image
vs.
image

We could probably also change the backgrounds of the little stats to match the commit listing bg, same deal with the readme.

rudolfs:
> Some backgrounds are too subtle, eg. in the commit listing it's barely visible.

2e51be257a2a397473e15f3065915bf991b99623

> We could probably also change the backgrounds of the little stats to match the commit listing bg, same deal with the readme, and also increase the brightness of the commit history date headers.

706462170821b3ce72e13a810d9a5254fbd71373

Screenshot 2022-09-13 at 17 14 28

Screenshot 2022-09-13 at 17 12 16

PR #306, opened 2022-08-07
Add cache stores and fn

sebastinez:
This PR adds a wrapper function called `cache` to multiple places where lots of fetching and computing is done, to avoid having lots of Loaders and wait times.

**Pending:**
- [x] Expire information after a certain `max-age`
- [x] Limit the amount of objects, contained in a store

### DIFF
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -5,6 +5,9 @@ import { OperationType } from "@gnosis.pm/safe-core-sdk-types";
 
 import { assert } from '@app/error';
 import * as utils from '@app/utils';
+import * as cache from "@app/cache";
+import type { SafeMultisigTransactionListResponse } from '@gnosis.pm/safe-service-client';
+import type SafeServiceClient from '@gnosis.pm/safe-service-client';
 import type { Safe } from '@app/utils';
 import type { Config } from '@app/config';
 import type { PendingAnchor, Anchor } from '@app/project';
@@ -186,7 +189,9 @@ export class Org {
 
   async getProjects(config: Config): Promise<Anchor[]> {
     const result = await utils.querySubgraph(
-      config.orgs.subgraph, GetProjects, { org: this.address }
+      config.orgs.subgraph,
+      GetProjects,
+      { org: this.address }
     );
     const projects: Anchor[] = [];
 
@@ -214,8 +219,8 @@ export class Org {
 
     try {
       const orgAddr = ethers.utils.getAddress(this.address);
-      const response = await config.safe.client.getPendingTransactions(
-        ethers.utils.getAddress(this.owner)
+      const response = await getPendingProjects(
+        ethers.utils.getAddress(this.owner), config.safe.client
       );
       const projects: PendingAnchor[] = [];
 
@@ -298,17 +303,10 @@ export class Org {
     addressOrName: string,
     config: Config,
   ): Promise<Org | null> {
-    const org = new ethers.Contract(
-      addressOrName,
-      config.abi.org,
-      config.provider
-    );
+    const org = await getOrgContract(addressOrName, config);
 
     try {
-      const [owner, resolved] = await Promise.all([
-        org.owner(),
-        org.resolvedAddress,
-      ]);
+      const [owner, resolved] = await resolveOrgOwner(org);
 
       const safe = await utils.getSafe(owner, config);
       // If what is resolved is not the same as the input, it's because we
@@ -391,3 +389,38 @@ export function parseAnchorTx(data: string, config: Config): { id: string; state
   }
   return null;
 }
+
+export const getOrgContract = cache.cached(
+  async (addressOrName: string, config: Config) => {
+    return new ethers.Contract(
+      addressOrName,
+      config.abi.org,
+      config.provider
+    );
+  },
+  (addressOrName) => addressOrName
+);
+
+export const resolveOrgOwner = cache.cached(
+  async (org: ethers.Contract) => {
+    return await Promise.all([
+      org.owner(),
+      org.resolvedAddress,
+    ]);
+  },
+  (org) => org.address,
+);
+
+export const getPendingProjects = cache.cached(
+  async (owner: string, client: SafeServiceClient): Promise<SafeMultisigTransactionListResponse> => {
+    try {
+      return await client.getPendingTransactions(
+        ethers.utils.getAddress(owner)
+      );
+    } catch (e) {
+      return { count: 0, results: [] };
+    }
+  },
+  (owner) => owner,
+  { max: 1000, ttl: 5 * 60 * 1000 } // Cache results for 5 minutes.
+);
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,0 +1,25 @@
+import LruCache from "lru-cache";
+
+// Creates a function that memoizes its result using an LRU cache.
+//
+// The cache key is a string created from the arguments using
+// `makeKey`.
+export function cached<Args extends unknown[], V>(
+  f: (...args: Args) => Promise<V>,
+  makeKey: (...args: Args) => string,
+  options?: LruCache.Options<string, { value: V }>
+): (...args: Args) => Promise<V> {
+  const cache = new LruCache(options || { max: 500 });
+  return async function (...args: Args): Promise<V> {
+    const key = makeKey(...args);
+    const cached = cache.get(key);
+
+    if (cached === undefined) {
+      const value = await f(...args);
+      cache.set(key, { value });
+      return value;
+    } else {
+      return cached.value;
+    }
+  };
+}
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -9,12 +9,13 @@ import type { SafeSignature } from "@gnosis.pm/safe-core-sdk-types";
 import type { Config } from '@app/config';
 import config from "@app/config.json";
 import { assert } from '@app/error';
-import type { EnsProfile } from "@app/base/registrations/registrar";
+import { EnsProfile, getAddress, getResolver } from "@app/base/registrations/registrar";
 import { getAvatar, getSeed, getAnchorsAccount, getRegistration } from '@app/base/registrations/registrar';
 import type { BasicProfile } from '@datamodels/identity-profile-basic';
 import { ProfileType } from '@app/profile';
 import { parseUnits } from "@ethersproject/units";
 import { GetSafe } from "@app/base/orgs/Org";
+import * as cache from "@app/cache";
 
 export enum AddressType {
   Contract,
@@ -58,7 +59,7 @@ export type State =
   | { status: Status.Failed; error: string };
 
 export async function isReverseRecordSet(address: string, domain: string, config: Config): Promise<boolean> {
-  const name = await config.provider.lookupAddress(address);
+  const name = await lookupAddress(address, config);
   return name === domain;
 }
 
@@ -337,7 +338,7 @@ export function safeLink(addr: string, config: Config): string {
 }
 
 // Query a subgraph.
-export async function querySubgraph(
+export async function querySubgraphWithRetry(
   url: string,
   query: string,
   variables: Record<string, any>,
@@ -358,12 +359,18 @@ export async function querySubgraph(
   if (json.errors) {
     console.error("querySubgraph:", json.errors);
 
-    if (retries > 0) querySubgraph(url, query, variables, retries - 1);
+    if (retries > 0) querySubgraphWithRetry(url, query, variables, retries - 1);
     else return null;
   }
 
   return json.data;
 }
+
+export const querySubgraph = cache.cached(
+  querySubgraphWithRetry,
+  (url: string, query: string, variables: Record<string, any>) => JSON.stringify({ url, query, variables }),
+  { max: 500, ttl: 5 * 60 * 1000 } // Cache results for 5 minutes.
+);
 
 // Format a name.
 export function formatName(input: string, config: Config): string {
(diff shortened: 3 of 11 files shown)


src/store.ts
```diff
@@ -0,0 +1,26 @@
...
+export const type = writable<{ [addressOrName: string]: AddressType }>({});
+export const org = writable<{ [addressOrName: string]: Org }>({});
+export const seed = writable<{ [hostName: string]: Seed }>({});
+
+ens.subscribe(s => console.log("ens store", s));
+type.subscribe(s => console.log("type store", s));
+org.subscribe(s => console.log("org store", s));
+seed.subscribe(s => console.log("seed store", s));
+
+export async function cache<T>(key: string, store: Writable<Record<string, T>>, query: () => Promise<T>): Promise<T> {
```
cloudhead, line 17: Would be nice if the key was passed as an argument to the closure

sebastinez: Yeah good idea 👍

src/store.ts
```diff
@@ -0,0 +1,22 @@
...
+//
+// The cache key is a string created from the arguments using
+// `makeKey`.
+export async function cache<V>(
+  k: string,
+  f: (key: string) => Promise<V>,
+): Promise<V> {
+  const cached = lru.get<{ value: V }>(k);
+  if (! cached) {
+    console.log(k);
```
rudolfs, line 15: 👀

sebastinez: Still a work in progress 😄 
And had to make some changes to your implementation.

rudolfs (commented):
Projects don't seem to load on the custom seeds:

Screenshot 2022-08-15 at 09 43 06

They only load if I do a hard refresh on the seed page, but not when I get there from the index page.

src/store.ts
```diff
@@ -0,0 +1,22 @@
...
+export const org = new LruCache<string, Org | null>({ max: 100 });
+export const request = new LruCache<string, any>({ max: 1000 });
+
+export async function cache<T>(key: string, store: LruCache<string, T>, query: (key: string) => Promise<T>): Promise<T> {
+  if (store.has(key)) {
+    const value = store.get(key);
+    if (value !== undefined) {
+      return value;
+    }
+  }
```
rudolfs, line 17: I think the outer `if` is superfluous, if the key isn't in the cache, the stats for it won't be incremented and if it is, we have to get it anyway.
```suggestion
  const value = store.get(key);
  if (value !== undefined) {
    return value;
  }
```

src/store.ts
```diff
@@ -0,0 +1,22 @@
+import LruCache from "lru-cache";
```
rudolfs, line 1: I'd call the file `cache` rather than `store`.

sebastinez: Yeah I agree, will rename it

src/store.ts
sebastinez: Technically I agree, the reason I added the `has` check is:

1. The return value of the `get` method and 2. the eventual update on the recency of the cache entry
> _From the lru-cache documentation_
> Will update the recency of the cache entry found.
>
> If the key is not found, get() will return undefined. This can be confusing when setting values specifically to undefined, as in cache.set(key, undefined). Use cache.has() to determine whether a key is present in the cache at all.

By using `has` as a first `if` I was able to get a boolean to check if the entry really existed, no matter if the entry would be `null`, `undefined` or another value, before trying to get it.

sebastinez:
> Projects don't seem to load on the custom seeds:
> 
> Screenshot 2022-08-15 at 09 43 06
> 
> They only load if I do a hard refresh on the seed page, but not when I get there from the index page.

Good catch @rudolfs 👏 , it seems like there is some race condition while instantiating the `Profile` class and getting the `ens.seed` value.. I'm looking into it

src/store.ts
rudolfs: Hm, in that case I'm confused by the inner `if (value !== undefined)`, because that means that we'll always hit the query branch in case the actual return value of the query is undefined instead of caching it.

sebastinez: i c, yeah probably we should remove the `has` check and don't allow `undefined` cache entries.

sebastinez: Done in 50b15b4fd0bd5a30461342f726bf4b24810118f9

src/store.ts
sebastinez: 50b15b4fd0bd5a30461342f726bf4b24810118f9

rudolfs:
I had another look at the code, a couple of things that I wanted to improve upon:
  - there's a bit too much indirection for my taste
  - currently all API requests are being cached, even PUT and POST requests, which probably shouldn't
  - query requests that return `undefined` are never cached
  - we don't set any TTL values, which means that the results would be cached until the user reloads or the cache max limits get hit
  - i find the use of "memoize" in this context a bit confusing, to me a memoize function should wrap another function, in effect returning a memoized function.

I copied the original `memoizeLru` function from Upstream and applied it to two places in this code base, I think it achieves a similar effect as this PR, but in less code. Have a look: https://github.com/radicle-dev/radicle-interface/tree/rudolfs/memoize

sebastinez:
> I had another look at the code, a couple of things that I wanted to improve upon:

Thanks @rudolfs have added your feedback to dcd9a78658f13c772958ce80aad2f3ea91f4b90d

src/utils.ts
```diff
@@ -440,57 +451,61 @@ export async function resolveIdxProfile(caip10: string, config: Config): Promise
...
-        seed,
-        anchorsAccount,
-      };
     }
-  }
-  return null;
-}
+    return null;
+  },
+  (addressOrName: string, profileType: ProfileType) => JSON.stringify({ addressOrName, profileType }),
```
rudolfs, line 506: I wonder if we shouldn't also key it by `config.network.chainId`? Do the addresses that we query differ across networks?

src/base/orgs/Org.ts
```diff
@@ -391,3 +382,23 @@ export function parseAnchorTx(data: string, config: Config): { id: string; state
   }
   return null;
 }
+
+const memoizeOrg = cache.memoizeLru(
```
rudolfs, line 386:
```suggestion
const memoizedOrg = cache.memoizeLru(
```

src/base/orgs/Org.ts
```diff
@@ -305,19 +308,7 @@ export class Org {
...
-
-      const safe = await utils.getSafe(owner, config);
-      // If what is resolved is not the same as the input, it's because we
-      // were given a name.
-      if (utils.isAddressEqual(addressOrName, resolved)) {
-        return new Org(resolved, owner, null, safe);
-      } else {
-        return new Org(resolved, owner, addressOrName, safe);
-      }
+      return memoizeOrg(addressOrName, org, config);
```
rudolfs, line 311:
```suggestion
      return memoizedOrg(addressOrName, org, config);
```

src/utils.ts
sebastinez: > I guess if it's guaranteed that there is a page refresh upon network switch then it shouldn't be a problem, cause the cache would be yanked anyway, right?

Yes we are indeed doing a reload with `window.location.reload()` everytime we change the `chainId` here
https://github.com/radicle-dev/radicle-interface/blob/65bf2ae38678caf6e7aa444f9a24bf4efd61f7f8/src/session.ts#L307

src/base/orgs/Org.ts
sebastinez: 3b6e7dd ✅

src/base/orgs/Org.ts
sebastinez: 3b6e7dd ✅

sebastinez:
> I spotted another place that could use some caching: `https://safe-transaction.gnosis.io`, didn't apply memoize to that yet.

@rudolfs I added caching for this in 5de59eb I also added caching for IDX profiles in 8f999bb

src/cache.ts
```diff
@@ -0,0 +1,32 @@
...
+  makeKey: (...args: Args) => string,
+  options?: LruCache.Options<string, { value: V }>
+): (...args: Args) => Promise<V> {
+  const cache = new LruCache(options || { max: 500 });
+
+  if (! window.cache) {
+    window.cache = [cache];
+  } else {
+    window.cache.push(cache);
+  }
```
rudolfs, line 18: We should not leave this in here, it was meant only for debugging. Otherwise these objects will take up memory.

src/config.ts
```diff
@@ -13,6 +13,7 @@ declare global {
   interface Window {
     // eslint-disable-next-line @typescript-eslint/naming-convention
     Cypress: any;
+    cache: any;
```
rudolfs, line 16: This should go as well.

src/base/orgs/Org.ts
```diff
@@ -391,3 +384,37 @@ export function parseAnchorTx(data: string, config: Config): { id: string; state
...
+    // If what is resolved is not the same as the input, it's because we
+    // were given a name.
+    if (utils.isAddressEqual(addressOrName, resolved)) {
+      return new Org(resolved, owner, null, safe);
+    } else {
+      return new Org(resolved, owner, addressOrName, safe);
+    }
+  },
+  (addressOrName) => addressOrName,
+  { max: 1000, ttl: 30 * 60 * 1000 } // Cache results for 5 minutes.
```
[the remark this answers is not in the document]
sebastinez: The org information is almost static at the moment and if someone changes the owner of an org, or something similar 30 min cache is more than enough.

I feel like with only doing 2 to 5 minutes, the page ends up refetching a lot of stuff, and it feels a lot less cached.

src/utils.ts
```diff
@@ -410,23 +417,27 @@ export function formatProjectHash(multihash: Uint8Array): string {
...
+      return AddressType.Contract;
     }
-    return AddressType.Contract;
-  }
-  return AddressType.EOA;
-}
+    return AddressType.EOA;
+  },
+  (address: string) => address,
+  { max: 1000 }
```
rudolfs, line 439: We should set a TTL here.

sebastinez: I think thats not necessary, since the type of an address never changes, so having a 1000 entries as max value and that the least recent drops out, is good enough, also we only store `[address]: u8` key pairs

Also I feel we even should consider removing some TTLs since `lru-cache` does not remove them actively from the cache, so they still add to the size of the cache, but they are considered stale after the time passed.
> Max time to live for items before they are considered stale. Note that stale items are NOT preemptively removed by default, and MAY live in the cache, contributing to its LRU max, long after they have expired.

rudolfs: > Also I feel we even should consider removing some TTLs since lru-cache does not remove them actively from the cache, so they still add to the size of the cache, but they are considered stale after the time passed.

I'd keep the other TTLs just for the sake of getting the user more up-to-date data.

cloudhead:
This is looking pretty good, I like the use of the LRU and memoization.. There's three things I'll comment on though:

1. I think it's probably not a good idea to decide how long to cache GET responses in the client: this should be the job of the server, by using the `Cache-Control` http header. This type of caching is also handled by the browser, so it's kind of redundant to do the caching ourselves (and our caching doesn't survive reloads either whereas the browser cache does).
2. Naming: small thing but we don't need to leak the implementation of the cache (LRU + memoization) in the external API (memoizeLru), because that doesn't communicate the intent of the function as well. I'd propose to just call this function `cached`, eg. `const getOrg = cached(doSomeRequest(a, b))`.
3. It seems like the caching has quite a bit of surface area already, since we're wrapping a bunch of functions. To me, caching should be something fairly transparent that the developer doesn't worry about too much because it's implemented at a lower level. I think it's *ok* to do what we're doing as long as we keep the wrapping of functions constrained to one or two modules and not having caching code spread all over the place. One way to approach this is to cache at a slightly lower level, ie. not cache functions that have a lot of logic in them, but rather cache the atomic parts, eg. name resolution, profile fetch, etc.

sebastinez:
> This is looking pretty good, I like the use of the LRU and memoization.. There's three things I'll comment on though:
> 
> 1. I think it's probably not a good idea to decide how long to cache GET responses in the client: this should be the job of the server, by using the `Cache-Control` http header. This type of caching is also handled by the browser, so it's kind of redundant to do the caching ourselves (and our caching doesn't survive reloads either whereas the browser cache does).
> 2. Naming: small thing but we don't need to leak the implementation of the cache (LRU + memoization) in the external API (memoizeLru), because that doesn't communicate the intent of the function as well. I'd propose to just call this function `cached`, eg. `const getOrg = cached(doSomeRequest(a, b))`.
> 3. It seems like the caching has quite a bit of surface area already, since we're wrapping a bunch of functions. To me, caching should be something fairly transparent that the developer doesn't worry about too much because it's implemented at a lower level. I think it's _ok_ to do what we're doing as long as we keep the wrapping of functions constrained to one or two modules and not having caching code spread all over the place. One way to approach this is to cache at a slightly lower level, ie. not cache functions that have a lot of logic in them, but rather cache the atomic parts, eg. name resolution, profile fetch, etc.

Thanks @cloudhead, will try to respond to your comments:

1. I agree with you, eventually all http-api instances should have at least a sensible `max-age` value in their `Cache-Control`.
2. 👍 
3. Here we have IMO to differentiate between `querySubgraph` functions that are easily cached individually, and parts of bigger libraries like ethers.js or gnosis-sdk, where the functions can also get memoized but it can get more verbose quickly. We'll need to fine tune until we find the correct balance.

cloudhead:
Cool, I think this is ready to merge.

I'd like to see in a separate PR though if we can improve the code around for eg. the resolver. Instead of:
```js
getText(resolver, "a");
getText(resolver, "b");
getText(resolver, "c");
```
We should aim for:
```js
let resolver = cachedResolver(resolver);
resolver.getText("a");
resolver.getText("b");
resolver.getText("c");
```

sebastinez:
> I'd like to see in a separate PR though if we can improve the code around for eg. the resolver. Instead of:

Yeah I worked around a few ways of improving this, in the end I opted for this, since the other ways I found were suboptimal..
Will look thought to improve in another PR

Issue #287, opened 2022-07-01
Make it possible to connect a wallet on mobile


brandonhaslegs:
Not possible at the moment but I've definitely seen it done in other web3 apps.

![image](https://user-images.githubusercontent.com/4406983/176916628-cbfbf5c2-c9cf-47e2-bba4-7f66ebc5aa06.png)

cloudhead:
Yeah good point, I think it requires links that open apps on your phone.

brandonhaslegs:
I think @efstajas made this work for rad.lol?

efstajas:
On rad.lol I just used the official WalletConnect SDK & connection modal, which implements all of this out of the box — platform-detects iOS to display the app-specific links, on Android shows a single "Connect" link, and on desktop a QR code.

I did look into it however before deciding implementing it custom wasn't worth it — you can get a list of wallets registered with WalletConnect App Registry through this API: https://registry.walletconnect.com/api/v1/wallets?entries=5&page=1 Then, for the connection modal on iOS, it's a matter of displaying all these apps and triggering the universal link contained in the API response, with `/wc?uri=<connection uri>` at the end, e.g. `https://link.trustwallet.com/wc?uri=wc%3A7c8c54cd-c31e-4d4d-93d7-d0ad7e1fb85b%401%3Fbridge%3Dhttps%253A%252F%252Fx.bridge.walletconnect.org%26key%3Ddb699e6bd6e96ee9c4a8cf53f499b30c9741698a51b3dbcec5cc53c4c40f0952`.

If you think it's worth building a custom UI for this, I'd be happy to give it a go sometime soon after wrapping up the next Workstreams milestone. Another option would obviously be just using the WalletConnect SDK.

sebastinez:
Okay according to what we talked over Discord (https://discord.com/channels/841318878125490186/856904179820920833/1069181291284660286) we'll remove all eth functionality from the web app, so this issue is no longer relevant

PR #308, opened 2022-08-09
Remove WalletConnect compatible link

### LINKED ISSUE #299: Fix broken compatible WalletConnect wallets
And again it seems that the link to a compatible wallet listing for WalletConnect is broken.

The latest working link it seems is: https://explorer.walletconnect.com/registry?type=wallet

This edit should be made here: https://github.com/radicle-dev/radicle-interface/blob/cbbaadd3307bba182e8cc88a3a1279b0a42825f5/src/components/Modal/ConnectWallet.svelte#L56

sebastinez:
- Removes WalletConnect-compatible wallet link, which is not reliable.
- Adds a generic text to let people know, they should use a WalletConnect-compatible wallet.

Closes #299

### DIFF
--- a/src/components/Modal/ConnectWallet.svelte
+++ b/src/components/Modal/ConnectWallet.svelte
@@ -52,10 +52,8 @@
 
     <div slot="subtitle">
       <div class="text-small">
-        WalletConnect<br/>
-        <a href="https://walletconnect.com/registry?type=wallet" class="link">
-          View compatible wallets
-        </a>
+        Scan the QR code with <strong>WalletConnect</strong> or use
+        <strong>Metamask</strong>.
       </div>
     </div>
 


src/components/Modal/ConnectWallet.svelte
```diff
@@ -52,10 +52,7 @@
 
     <div slot="subtitle">
       <div class="text-small">
-        WalletConnect<br/>
-        <a href="https://walletconnect.com/registry?type=wallet" class="link">
-          View compatible wallets
-        </a>
+      Scan QR code with a <br />WalletConnect-compatible wallet
```
cloudhead, line 55: Can we just leave this as `WalletConnect` like before and remove the link?

sebastinez: Yeah sure 👍

sebastinez: It seemed that `WalletConnect` by itself, was standing pretty alone so I thought an explanation was maybe better than to just show `WalletConnect`. 

Or maybe we could change the `Connect your wallet` to something alluding to `WalletConnect`?
Maybe `Connect with WalletConnect`?

sebastinez: Maybe even better we could do a listing of different wallets, similar to Workstreams
Bildschirmfoto 2022-08-12 um 14 33 57

The Metamask button would be like clicking on `Connect with Metamask` and when clicking on WalletConnect, we could implement into our own style, the WalletConnect QR Modal to allow connection with deep linking on mobile ref #287

Would probably need some rework on how we manage the wallet connect state inside the app

wdyt @cloudhead

cloudhead: Two clicks vs. 1

sebastinez: > Two clicks vs. 1

Okay 👌

PR #313, opened 2022-08-09
Refactor project header modals

rudolfs:
Decouple global modal state from the parent element by moving it into a module store of the modal itself.

This makes the floating modal component more composable, we don't have to track the modal state in the parent component anymore. This means we can just add as many floating modals to a page as we like and only one of them will stay open at a time.

This change also implements hiding any open modals when clicking outside of the modal.

https://user-images.githubusercontent.com/158411/183942984-3f6cc541-4290-41cd-8bbf-8e4bb5852049.mov

### DIFF
--- a/cypress/e2e/project.spec.ts
+++ b/cypress/e2e/project.spec.ts
@@ -90,7 +90,7 @@ describe("Project view", () => {
     cy.get("div.stat.contributor-count").should("have.text", "1 contributor(s)");
     cy.get("div.stat.branch").should("have.class", "not-allowed").should("have.text", "main");
     cy.get("div.hash.desktop").should("have.text", "56e4e02");
-    cy.get("div.clone").click();
+    cy.get("div.clone-button").click();
   });
 
   it("Peer selector", () => {
--- a/src/base/projects/BranchSelector.spec.ts
+++ b/src/base/projects/BranchSelector.spec.ts
@@ -18,7 +18,6 @@ const defaultProps = {
   },
   branches: { "master": "e678629cd37c770c640a2cd997fc76303c815772" },
   revision: "e678629cd37c770c640a2cd997fc76303c815772",
-  toggleDropdown: () => "branch"
 };
 
 describe('Logic', () => {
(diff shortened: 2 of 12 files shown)


cypress/e2e/project.spec.ts
```diff
@@ -105,7 +104,7 @@ describe("Project view", () => {
   });
 
   it("Branch selector", () => {
-    cy.get("div.commit div.stat.branch").click();
+    cy.get("div.commit div.stat.branch").click().click();
```
rudolfs, line 107: Cypress uses js click events which behave differently than when a real user clicks. I don't fully understand why this doesn't work in Cypress, hence the workaround.

Would be nice if we could switch to Playwright, which doesn't have issues like this and has overall a better DX.

src/base/projects/BranchSelector.spec.ts
```diff
@@ -141,26 +140,3 @@ describe("Layout", () => {
     cy.get("div.hash.desktop").should("be.visible");
   });
 });
-
```
rudolfs, line 144: Couldn't get the tests to run locally to debug this: https://github.com/radicle-dev/radicle-interface/issues/318.

src/base/projects/PeerSelector.spec.ts
```diff
@@ -57,39 +56,9 @@ describe('Logic', function () {
...
   it("should highlight the current peer", () => {
     render(PeerSelector, {
-      props: { ...defaultProps, peersDropdown: true }
+      props: { ...defaultProps }
     });
+    cy.get("div.selector").click();
     cy.get("div.dropdown-item").should("have.class", "selected");
   });
 });
-
```
rudolfs, line 65: Couldn't get the tests to run locally to debug this: https://github.com/radicle-dev/radicle-interface/issues/318.

cloudhead (approved):
Very nice. I wonder if there's a better name though? When I saw this I thought it was a change to the full-screen modals. Maybe `Popup`?

cloudhead:
Or `Floating`

rudolfs:
Yeah, had the same thought. I like `Popup`.

rudolfs:
Ugh, `popup` is already taken for the verified badge. I'll rename it to `Floating`.

PR #316, opened 2022-08-10
Improve handling of low commit counts in repo activity

sebastinez:
This PR improves the handling of weekly low commit counts.

**Handling:**
- If the commit count is lower than 5 we set the height to 5.
- Else If the `maxValue` is lower than the `viewBoxHeight`, we show the height corresponding to the commit count without normalization
- Else, once the `viewBoxHeight` gets crossed by the `maxValue` we start to do a linear normalization by max/min values.

**Edge cases:**
- If there was no activity in a year we show a straight line

**Known issues:**
- If there is a big delta between commit counts, the lower data points are disappearing, I did some tests with a linear normalization over logarithmic values, but the issue is that the higher data points level out to be very similar..
_Conclusion_: We should monitor if the graphs now shows better the activity on the repo and handle if we see need for improvement.
_Example_: https://svelte.dev/repl/78211cc963754a55968f5c8a35757008?version=3.49.0

### DIFF
--- a/src/Diagram.svelte
+++ b/src/Diagram.svelte
@@ -22,6 +22,9 @@
   // The highest value on the y axis
   let commitCountArray: number[] = [];
 
+  // The minimal amplitude shown e.g. commitCount = 1 => `minimalHeight` points of height in the SVG.
+  let minimalHeight = 5;
+
   let week = 0;
 
   for (const point of points) {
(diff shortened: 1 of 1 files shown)


rudolfs (commented):
The graphs look good to me.

I guess this applies more to #306, but the requests for the `commits` endpoint could use some caching, otherwise it takes half a minute for the graphs to appear. By that time people probably will have already navigated away from that screen. These could even have a longer TTL and eventually persisted to local storage, so they're always available.

Screenshot 2022-08-11 at 17 15 33

sebastinez:
> The graphs look good to me.
> 
> I guess this applies more to #306, but the requests for the `commits` endpoint could use some caching, otherwise it takes half a minute for the graphs to appear. By that time people probably will have already navigated away from that screen. These could even have a longer TTL and eventually persisted to local storage, so they're always available.
> 
> Screenshot 2022-08-11 at 17 15 33

Hmm yeah that's true, and the reason why we display the graphs with the page mostly loaded.
How would you suggest we should load the graphs before the user hits a project listing for the first time?

cloudhead:
Nice, let's try this out for a while and see if we find any outliers.

Re: caching, another solution is to have an endpoint that just gives commit counts, or perhaps an endpoint that batches data for multiple projects?

rudolfs:
> Re: caching, another solution is to have an endpoint that just gives commit counts, or perhaps an endpoint that batches data for multiple projects?

Yeah, I like that idea prolly also easier to cache that in the backend anyway if it takes long to compute.

PR #319, opened 2022-08-12
Don't use outdated node on CI

rudolfs:
ubuntu-latest [already includes](https://github.com/actions/runner-images/blob/ubuntu20/20220807.1/images/linux/Ubuntu2004-Readme.md#language-and-runtime
) a recent node version (16.16.0).

And node 14 is at its EOL.
```
WARNING: node-v14.20.0 is in LTS Maintenance mode and nearing its end of life.
It only receives *critical* security updates, *critical* bug fixes and documentation updates.
```

This should also shave off some time from our builds.

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -6,8 +6,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
-          node-version: '14'
       - run: npm ci
       - run: npm run build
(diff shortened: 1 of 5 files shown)


sebastinez (approved):
lgtm, any reason why we should go with v14? Would it be possible to go for v16?

rudolfs:
> lgtm, any reason why we should go with v14? Would it be possible to go for v16, since it's the current LTS?

I don't understand the question. The OS image already includes node v16. And we _are_ using it now.

sebastinez:
> I don't understand the question. The OS image already includes node v16. And we _are_ using it now.

🤦 you're right. nvm my comment!

PR #320, opened 2022-08-12
Change Node version to v16

sebastinez:
Issues with Cypress component tests forced us to downgrade to Node v16.
Also adding `.tool-versions` file to work with `asdf`


rudolfs (approved):
Nice. Now Cypress component tests work for me locally. This also explains why it worked on CI.

PR #322, opened 2022-08-15
Hide debug logs

rudolfs:
I think debug logs should be logged with the appropriate log level, so they don't mix in with other stuff that's going on in the console.

They're still accessible via:
Screenshot 2022-08-15 at 17 41 55


sebastinez (approved):
I like it, makes the normal console log a bit less verbose

cloudhead (approved):

PR #324, opened 2022-08-16
Get rid of crossorigin warnings in console

rudolfs:
This gets rid of these:

⚠️ A preload for 'https://app.radicle.network/fonts/Inter-Regular.woff' is found, but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.


sebastinez (approved):

cloudhead:
:exploding_head:

PR #325, opened 2022-08-16
Make a missing reverse record error less critical

sebastinez:
This PR adds a new Error type which is being used when a ENS profile has no address set, this means that it is missing a reverse record.

And instead of showing a `No address found` we should show the user that they didn't set the reverse record.


src/error.ts
```diff
@@ -34,6 +34,16 @@ export class NotFoundError extends Error {
   }
 }
 
+export class MissingReverseRecord extends Error {
+  constructor(message?: string) {
+    if (message) {
+      super(`missing reverse record for ${message}`);
```
cloudhead, line 40: Isn't this going to show up as `missing reverse record for No address et for ${address}`?

sebastinez: Currently the way `NotFoundError` is displayed is according to `src/Profile.svelte` line 400.
To differentiate between `NotFoundError` and a missing reverse record I needed to differentiate between those errors.

But we could eventually also let the error message bubble up and show it in the `NotFound` component, which would allow us to avoid creating another `Error` class. wdyt @cloudhead?
Or do you prefer to separate the `Error` instance from the  svelte component?

cloudhead: Ok, it won't show up that way because we're not using the `message` field in that one instance; but the error message is still badly formed, why don't we just fix it?

sebastinez: Done in [6da9bfe](https://github.com/radicle-dev/radicle-interface/pull/325/commits/6da9bfe6173ce1e3726a6418abb5b714bf328137)

cloudhead (approved):

PR #326, opened 2022-08-16
Use ethers defaultProvider when in development mode

sebastinez:
Currently when running the app locally we require the user to have Metamask installed.
We could rely on the Infura Id used by Ethers to have a provider for when `import.meta.env.DEV === true`.
This would improve the UX for users wanting to use the interface on localhost.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -216,13 +216,17 @@ function getProvider(
   config: Record<string, any>,
   metamask: ethers.providers.JsonRpcProvider | null,
 ): ethers.providers.JsonRpcProvider {
-  // Use Alchemy in production. Otherwise use Metamask if installed.
   if (import.meta.env.PROD) {
     return new ethers.providers.AlchemyWebSocketProvider(network.name, config.alchemy.key);
   } else if (metamask) {
     return metamask;
+  } else if (import.meta.env.DEV) {
+    // The ethers defaultProvider doesn't include a `send` method, which breaks the `utils.getTokens` fn.
+    // Since Metamask nor WalletConnect provide an `alchemy_getTokenBalances` nor `alchemy_getTokenMetadata` endpoint,
+    // we can rely on not using `config.provider.send`.
+    return ethers.providers.getDefaultProvider() as ethers.providers.JsonRpcProvider;
   } else {
-    throw new Error("A wallet such as Metamask must be enabled on your browser.");
+    throw new Error("No Web3 provider available.");
   }
 }
 


src/config.ts
```diff
@@ -216,13 +216,16 @@ function getProvider(
...
 ): ethers.providers.JsonRpcProvider {
-  // Use Alchemy in production. Otherwise use Metamask if installed.
   if (import.meta.env.PROD) {
     return new ethers.providers.AlchemyWebSocketProvider(network.name, config.alchemy.key);
   } else if (metamask) {
     return metamask;
+  } else if (import.meta.env.DEV) {
+    // We are using here ethers.js Infura project id, which is used by ethers.providers.getDefaultProvider()
+    // https://github.com/ethers-io/ethers.js/blob/4a8d579dcaf026d0c232e20176605d34cba4767d/packages/providers/src.ts/infura-provider.ts#L17
+    return new ethers.providers.InfuraProvider(network.name, "84842078b09946638c03157f83405213");
```
cloudhead, line 226: I wouldn't do this because it'll break when the key is revoked, which we have no control over..

Can we not call `ethers.getDefaultProvider()` or something?

sebastinez: Yeah that's true, the issue is that `ethers.getDefaultProvider()` returns a `BaseProvider` while we require a `JsonRpcProvider` will look into how we could still make it work.

sebastinez: Ok in the end we should be able to use `getDefaultProvider()` and cast the result to `JsonRpcProvider`

The only caveat there is that we won't have a `send` method available, which we use in the `utils.getTokens` function.
Nevertheless this function is almost exclusively for Alchemy since neither Metamask nor WalletConnect can query `alchemy_*` endpoints.

src/config.ts
```diff
@@ -216,13 +216,17 @@ function getProvider(
...
 ): ethers.providers.JsonRpcProvider {
-  // Use Alchemy in production. Otherwise use Metamask if installed.
   if (import.meta.env.PROD) {
     return new ethers.providers.AlchemyWebSocketProvider(network.name, config.alchemy.key);
   } else if (metamask) {
     return metamask;
+  } else if (import.meta.env.DEV) {
+    // The ethers defaultProvider doesn't include a `send` method, which breaks the `utils.getTokens` fn.
+    // Since Metamask nor WalletConnect provide an `alchemy_getTokenBalances` nor `alchemy_getTokenMetadata` endpoint we can rely on not using `config.provider.send`.
+    // TODO: Improve the provider used for local usage
```
cloudhead, line 226: What's the TODO about? by local usage you mean DEV?

sebastinez: The TODO is about that we eventually should find a way to use a `JsonRpcProvider`, since we eventually we'll find issues with using a `BasicProvider` casted as `JsonRpcProvider`

Issue #250, opened 2022-06-27
Show projects from seed nodes in search results


brandonhaslegs:
Show search results for projects
<img width="592" alt="image" src="https://user-images.githubusercontent.com/4406983/175996780-db31c75f-3820-41e0-98cc-95054d349862.png">

cloudhead:
Yeah this requires a network architecture that we don't have right now, but we will get there eventually.

sebastinez:
Okay, PR #327 does a fetching of all listed seed nodes in `config.seeds.pinned` and shows the projects on all the seed nodes it has been found by name or URN and let's you navigate to it.
If a project has the same name as a registered ENS name it lets the user choose

PR #327, opened 2022-08-18
Expand name search to be more dumb-user-friendly

### LINKED ISSUE #251: Expand name search to be more dumb-user-friendly
Show .eth and .radicle.eth search results (brandonhaslegs.eth and brandonhaslegs.radicle.eth both exist, but searching brandonhaslegs shows no results)
image

sebastinez:
# Better web search

- Adds a possible `.eth` top level domain to the `utils.isENSName` fn
- After typing the search query and clicking enter, the search field gets emptied
- Adds and improves some useful new methods and util functions:
  - `static async getMulti(addressesOrNames: string[], config: Config): Promise<(Profile | null)[]>`
     Allows us to query for multiple profiles and handle  eventual non existant profiles.
  - `static async getMulti(id: string, config: Config): Promise<(Project | null)[]>`
      Retrieve projects through a name or URN on all the seeds that are pinned under `config.seeds.pinned`
  - `async function resolveMultiLabel(labels: string[], config: Config): Promise<Record<string, string | null>>`
      Resolve of multiple ENS labels, and return an object with the following structure `{ <label>: <address> | null }` 
- Rewrites the search algorithm to the following logic:
```
Search resolution
    0. If query is null go to Home
    1. If it's an address, go to the profile page
    2. If it's a Radicle URN, get all matching projects from the pinned seeds
      2.1 If only one project is found, navigate to project page on the matching seed node
      2.2 If there are multiple projects, store in the results object
    3. Try to resolve the query as a ENS name
      3.1 If it resolves to an address and there werent found any projects, go to profile page
      3.2 If unable to resolve the query, lookup names with .radicle.eth and .eth
        3.2.1 If more than one name is found, save in the results object
        3.2.2 If exactly one address, and no projects are found, go to profile page
        3.2.3 If projects have been found, save the name in the results object
    4. If at least one project or one name is in the results object, let the user choose
    4.1 If the results object has neither projects nor names, go for the registration form.
```

Closes #251, #250


brandonhaslegs:
Cool! We can expand the search feature to have a more proper search results as you're typing type thing. But this is already an improvement

src/profile.ts
```diff
@@ -239,12 +239,12 @@ export class Profile {
     throw new NotFoundError(`Not able to resolve profile for ${addressOrName}`);
   }
 
-  static async getMulti(addressesOrNames: string[], config: Config): Promise<Profile[]> {
+  static async getMulti(addressesOrNames: string[], config: Config): Promise<(Profile | null)[]> {
```
cloudhead, line 242: I don't think there's much value in returning `null`s here, is there? Why not filter the results so we only return the projects that resolved?

src/base/resolver/Query.svelte
```diff
@@ -6,36 +6,96 @@
...
+      2.2 If there are multiple projects, store in the results object
+    3. Try to resolve the query as a ENS name
+      3.1 If it resolves to an address and there werent found any projects, go to profile page
+      3.2 If unable to resolve the query, lookup names with .radicle.eth and .eth
+        3.2.1 If more than one name is found, save in the results object
+        3.2.2 If exactly one address, and no projects are found, go to profile page
+        3.2.3 If projects have been found, save the name in the results object
+    4. If at least one project or one name is in the results object, let the user choose
+    4.1 If the results object has neither projects nor names, go for the registration form.
+  */
```
cloudhead, line 31: Can we stick to the `//` comments for consistency?

src/Cards.svelte
```diff
@@ -51,7 +51,9 @@
     {/each}
 
     {#each profiles as profile}
-      <Card {profile} {config} path={`/${profile.nameOrAddress}`} />
+      {#if profile}
+        <Card {profile} {config} path={`/${profile.nameOrAddress}`} />
+      {/if}
     {/each}
```
cloudhead, line 57: This API doesn't make sense anymore: why would a developer pass a list of `null`s instead of profiles? It won't render anything :)

sebastinez: Done in [fc8360d](https://github.com/radicle-dev/radicle-interface/pull/327/commits/fc8360dd71d450efb2ed7e46aa9635e6fa465769)

src/profile.ts
sebastinez: Done in [22c6bbc](https://github.com/radicle-dev/radicle-interface/pull/327/commits/22c6bbc42cae62b40f1f3997a5e271ea02369b9f)

sebastinez: The idea behind the `null` was to not filter at this stage but return project or `null`, and then the component can handle what to do if a `null` returns.
But we can also handle the non existence of a array item as `null` 😄

src/Cards.svelte
sebastinez: Yeah went back with all the `profile could be null` checks

src/base/resolver/Routes.svelte
```diff
@@ -3,10 +3,15 @@
   import Resolve from '@app/base/resolver/Query.svelte';
   import type { Config } from '@app/config';
   import * as utils from '@app/utils';
+  import List from "@app/base/resolver/List.svelte";
 
   export let config: Config;
 </script>
 
+<Route path="/resolver/list">
```
cloudhead, line 11: `/resolver/results` maybe more meaningful?

cloudhead:
Ok I'm trying this out now, pretty cool!

One issue is that the `/results` page doesn't really behave as expected: when you refresh it, it doesn't necessarily show the results again (but sometimes does?). I think we have to include the query string in the results page, and handle refreshes there properly.

sebastinez:
> One issue is that the `/results` page doesn't really behave as expected: when you refresh it, it doesn't necessarily show the results again (but sometimes does?). I think we have to include the query string in the results page, and handle refreshes there properly.

Thanks!
I tried to reproduce your issue, but wasn't able to.. In my case `window.history.state` was very persistant and reliable.
Could you maybe elaborate on how you managed to get no results? Some error maybe?
To fix this maybe we could redirect the user to `resolver/query?p=<query>` if we receive an object with empty arrays..

cloudhead:
> Thanks! I tried to reproduce your issue, but wasn't able to.. In my case `window.history.state` was very persistant and reliable. Could you maybe elaborate on how you managed to get no results? Some error maybe? To fix this maybe we could redirect the user to `resolver/query?p=<query>` if we receive an object with empty arrays..

I type "cloudhead", <enter>, I see the search results.
Then I click on the URL bar, and hit <enter> again, I see a blank page with a console error:

image

cloudhead:
Btw we need to fix the back button here again :D

sebastinez:
Okay here a video on how the search is performing now.. Seems less intrusive and more clear to users
Uploading Bildschirmaufnahme 2022-09-12 um 16.52.48.mov…

PR #328, opened 2022-08-22
Check name for dots at registration

sebastinez:
Since we are not able to create subdomains through the current registration flow, we avoid users to register names with subdomains in them.

Also updating ethers to `v5.7` to allow Wildcard Resolution in the future https://docs.ens.domains/ens-improvement-proposals/ensip-10-wildcard-resolution


cloudhead:
When I land on the page I get an error without having entered anything:
image

sebastinez:
> When I land on the page I get an error without having entered anything: image

Done in [efbd217](https://github.com/radicle-dev/radicle-interface/pull/328/commits/efbd21740daee730af71f5f95b10fcaaa424e424)

src/base/registrations/Index.svelte
```diff
@@ -65,16 +83,22 @@
...
-      <button disabled={!label || label.length < 2 || label.length > 128} class="primary register regular" on:click={register}>
-        Check
-      </button>
+        <button
+          disabled={!label || errors.length !== 0}
+          class="primary register regular"
+          on:click={register}
+        >
+          Check
+        </button>
```
cloudhead, line 101: indent

sebastinez: Done in [c5eddde](https://github.com/radicle-dev/radicle-interface/pull/328/commits/c5eddde23f733ad90bf3dd2f94dad2e722a7f4d6)

PR #330, opened 2022-08-22
Implements activity endpoint for project graph

sebastinez:


cloudhead:
Do we need to deploy new http-apis?

sebastinez:
> Do we need to deploy new http-apis?

To support the project activity graph, I went for breaking current `http-apis`.
We could try to fetch activity and if we receive a 404 go the current way, but I decided to go break things, since there aren't many http-apis in prod besides the community seeds.

We can also wait with this PR, until we have a more stable seed node architecture

PR #333, opened 2022-08-22
Improve UX on issue and patches stats in project header

### LINKED ISSUE #260: Update tooltip text on issues to something more explanatory
Not supported seems like they might be supported if I do something, or in another version of the app. Suggestions:

- Issues • Coming soon!
- or maybe just remove it until the feature launches

image

sebastinez:
This PR closes #260

### DIFF
--- a/src/base/projects/Header.svelte
+++ b/src/base/projects/Header.svelte
@@ -122,8 +122,8 @@
     {#if seed.api.host}
       <div
         class="stat seed clickable widget"
-        on:click={() => navigate(`/seeds/${seed.api.host}`)}
         title="Project data is fetched from this seed">
+        on:click={() => navigate(`/seeds/${seed.api.host}`)}
         <span>{seed.api.host}</span>
       </div>
     {/if}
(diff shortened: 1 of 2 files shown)


src/base/projects/Header.svelte
```diff
@@ -134,29 +134,29 @@
     on:click={() => toggleContent(ProjectContent.History, true)}>
     <strong>{tree.stats.commits}</strong> commit(s)
   </div>
-  {#if project.issues > 0}
+  {#if project.issues !== undefined}
```
cloudhead, line 137: Don't really like checking against `undefined` in application logic; what's the idea here?

cloudhead: Ok I got it, can you do `if "issues" in project`? would that work?

sebastinez: hmm yeah that should work

sebastinez: Ok I think we should be able to resolve it also with `if (project.issues)` since we only have to rule out that it's not `undefined`.

sebastinez: Done in [e7dcd40](https://github.com/radicle-dev/radicle-interface/pull/333/commits/e7dcd408fe68d87b3c9e5020e2ef165f04c2a5b4)

src/base/projects/Header.svelte
```diff
@@ -134,29 +134,29 @@
...
-  {#if project.issues > 0}
+  {#if project.issues}
     <div
       class="stat issue-count clickable widget"
+      class:not-allowed={project.issues === 0}
+      class:clickable={project.issues > 0}
       class:active={content == ProjectContent.Issues}
-      on:click={() => toggleContent(ProjectContent.Issues, false)}>
+      on:click={() => {
+        if (project.issues && project.issues > 0) toggleContent(ProjectContent.Issues, false); }
```
cloudhead, line 144: Seems like these checks should be redundant no? Due to line 137, 141

src/base/projects/Header.svelte
```diff
@@ -134,29 +134,29 @@
...
-  {#if project.patches > 0}
+  {#if project.patches}
     <div
       class="stat patch-count clickable widget"
+      class:not-allowed={project.patches === 0}
+      class:clickable={project.patches > 0}
       class:active={content == ProjectContent.Patches}
-      on:click={() => toggleContent(ProjectContent.Patches, false)}>
+      on:click={() => {
+        if (project.patches && project.patches > 0) toggleContent(ProjectContent.Patches, false); }
```
cloudhead, line 156: Same here

src/base/projects/Header.svelte
sebastinez: Yeah I wanted to be extra explicit, but by hiding or at least not adding a `clickable` class, we can obtain the same result

src/base/projects/Header.svelte
sebastinez: 👍

sebastinez:
Hmm since PR #347 has been merged I'll reduce the scope of this PR and leave just the changes in `src/project.ts`

PR #334, opened 2022-08-23
Remove unused component

rudolfs:
Doesn't seem to be used, can we get rid of it?


sebastinez (approved):
Yeah we can remove that, was before the `Avatar` component unified the avatars and blockies

PR #335, opened 2022-08-23
Refactor icon component

rudolfs:
We simplify the icon component by converting all icons to a 24x24 `viewBox`, flattening and centering them. This way we can rely on flexbox alignment for precise vertical and horizontal positioning. Flattening the icons allows us to only ever have to set the `fill` property to color an icon.

The refactored component is also more type safe, by requiring the `name` prop to be set to only specific values and checking the `if/else` branch for exhaustiveness.

Screenshot 2022-08-23 at 16 15 28
Screenshot 2022-08-23 at 16 15 36
Screenshot 2022-08-23 at 16 15 43
Screenshot 2022-08-23 at 16 15 55
Screenshot 2022-08-23 at 16 16 20
Screenshot 2022-08-23 at 16 16 30
Screenshot 2022-08-23 at 16 17 04


cloudhead (approved):
Nice work!

PR #342, opened 2022-08-25
Add missing `max` attribute to cache call

rudolfs:
This gets rid of the warning we are seeing in the console.

```
TTL caching without ttlAutopurge, max, or maxSize can result in unbounded memory consumption. UnboundedCacheWarning LRU_CACHE_UNBOUNDED
```


sebastinez (approved):
Haha okay yeah makes sense now. 👍

PR #344, opened 2022-08-25
Enable loading more items in commit history and project listing

sebastinez:
Adds a `More` button on the end of the commits history, and when clicked loads more commits

Bildschirmfoto 2022-08-25 um 15 37 18

This PR depends on the merge and implementation of https://github.com/radicle-dev/radicle-client-services/pull/181
Since this would allow us to check if there are more commits to be loaded or if we are already at the end

### DIFF
--- a/src/List.svelte
+++ b/src/List.svelte
@@ -1,0 +1,59 @@
+<script lang="ts">
+  import Button from "@app/Button.svelte";
+  import Loading from "@app/Loading.svelte";
+  import { fade } from "svelte/transition";
+
+  type Item = $$Generic;
+
+  export let items: Item[];
+  export let query: () => Promise<Item[]>;
+
+  // Used to handle the display of the trigger to load more items, according to the current loading state.
+  let loading = false;
+  // Should be changed to true, once no more items are being returned.
+  let complete = false;
+
+  const transitionParams = { duration: 200 };
+
+  const fetchMore = async () => {
+    loading = true;
+    const response = await query();
+
+    if (response.length > 0) {
+      items = [...items, ...response];
+    } else {
+      complete = true;
+    }
+
+    loading = false;
+  };
+</script>
+
+<style>
+  .more {
+    margin-top: 1rem;
+    height: 40px;
+    text-align: center;
+  }
+</style>
+
+<slot name="list" {items} />
+{#if !complete}
+  <slot name="more" {fetchMore} {loading}>
+    {#if loading}
+      <div class="more" transition:fade={transitionParams}>
+        <Loading small />
+      </div>
+    {:else}
+      <div class="more" transition:fade={transitionParams}>
+        <Button
+          variant="secondary"
+          waiting={loading}
+          disabled={loading}
+          on:click={fetchMore}>
+          More
+        </Button>
+      </div>
+    {/if}
+  </slot>
+{/if}
--- a/src/Card.svelte
+++ b/src/Card.svelte
@@ -1,12 +1,10 @@
 <script lang="ts">
-  import { onMount } from "svelte";
   import { Link } from "svelte-routing";
   import Avatar from "@app/Avatar.svelte";
   import type { Profile } from "@app/profile";
   import type { Config } from "@app/config";
   import { formatName, formatAddress } from "@app/utils";
   import type { Seed } from "@app/base/seeds/Seed";
-  import Loading from "@app/Loading.svelte";
 
   export let profile:
     | Profile
@@ -20,15 +18,6 @@
   export let config: Config;
   export let path: string;
   export let members: string[] = [];
-
-  let numberOfProjects: number | null = null;
-
-  onMount(async () => {
-    if (seed) {
-      const projects = await seed.getProjects();
-      numberOfProjects = projects.length;
-    }
-  });
 </script>
 
 <style>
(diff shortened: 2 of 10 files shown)


src/List.svelte
```diff
@@ -0,0 +1,48 @@
...
+    <div class="more">
+      <button class="small" on:click={fetchMore}>
+        {#if loading}
+          <Loading center small />
+        {:else}
+          More
+        {/if}
+      </button>
+    </div>
+    </slot>
```
cloudhead, line 47: indent

sebastinez: Done in [a23c649](https://github.com/radicle-dev/radicle-interface/pull/344/commits/a23c649bf5fd56a7a4bb830a47274adb72f97ba0)

src/List.svelte
```diff
@@ -0,0 +1,48 @@
...
+
+<style>
+  .more {
+    margin-top: 1rem;
+    text-align: center;
+  }
+</style>
+
+<slot name="list" {items} />
+{#if complete === false}
```
cloudhead, line 36: `{#if !complete}`?

src/List.svelte
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
+  import Loading from "@app/Loading.svelte";
+
+  export let items: any[];
+  export let query: any;
+
+  // Being passed to more slot to handle loading state of trigger
```
cloudhead, line 7: Can't parse this sentence

src/List.svelte
```diff
@@ -0,0 +1,48 @@
...
+  let complete = false;
+
+  const fetchMore = async () => {
+    loading = true;
+    const response = await query();
+
+    if (response.length > 0) {
+      items.push(...response);
+      // We explicitly do an assignment here to force Svelte to update the slots props
+      items = items;
```
cloudhead, line 19: Would it be better to split the starting items from the loaded ones?

src/List.svelte
sebastinez: sounded clear to me at the moment 🤣 
Will rephrase it

cloudhead:
Doesn't work after a while:
image

http://127.0.0.1:8080/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/history/cc80d84ea5be6466647777224c1131b2e0ad11c8

src/List.svelte
sebastinez: It would be possible, but in some cases like commit history, we ultimately need one single array of items to group them by date.

sebastinez:
> Doesn't work after a while: image
> 
> http://127.0.0.1:8080/seeds/seed.alt-clients.radicle.xyz/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/history/cc80d84ea5be6466647777224c1131b2e0ad11c8

Hmm ok that's weird, the only way I was able to reproduce the issue is when clicking quickly the more button more than once..
I'll add a disable attribute to the `More` button when in `loading` state, this should fix it

sebastinez:
@cloudhead please let me know if you can try it out again at some point, to rule this error out.

cloudhead:
The project listing gets duplicated when you click on (More), try the alt-clients seed for eg.

src/List.svelte
cloudhead: ok, then perhaps `items = [...items, ...response]`?

sebastinez:
@cloudhead PR needs more thinking, converted it back to a draft, will mark it as ready when finished.

sebastinez:
> The project listing gets duplicated when you click on (More), try the alt-clients seed for eg.

Hmm ok it seems that this is an issue when using a seed node without the recently merged project pagination..
Will look into a way to support older seed nodes.

cc @cloudhead

src/List.svelte
cloudhead: :eyes:

cloudhead:
Also I think you need to use the `<Button>` component for styling now.

sebastinez:
> c9cbda2

Done in [c9cbda2](https://github.com/radicle-dev/radicle-interface/pull/344/commits/c9cbda2d2bbf14a0d4af680f2d5f2ad142782c31)

sebastinez:
> Also I think you need to use the `<Button>` component for styling now.

Done in [c9cbda2](https://github.com/radicle-dev/radicle-interface/pull/344/commits/c9cbda2d2bbf14a0d4af680f2d5f2ad142782c31)

src/List.svelte
sebastinez: Hey thanks for the push, missed this comment, was a tricky one, it seemed that doing a `Array.push()` and then assigning was doing the trick while doing the spread one here not..
In the end I went with binding `items` passed into `List` with the initial list, I think it's a good solution, due to the availability of the data.
The only caveat that I can think of is that the `List` component now is somehow a fetcher and a display component, which initially I wanted to dumb down.

src/List.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
+  import Button from "@app/Button.svelte";
+  import Loading from "@app/Loading.svelte";
+
+  export let items: any[];
+  export let query: any;
```
cloudhead, line 6: Maybe for a future PR, giving this a type.

src/Loading.svelte
```diff
@@ -7,6 +7,7 @@
   export let fadeIn = false;
   export let margins = false;
   export let condensed = false;
+  export let noWait = false; // To avoid confusion when multiple cursor types are defined
```
cloudhead, line 10: Why do we want this to be `true` sometimes?

sebastinez: I removed the `noWait` prop in [92ef5e7](https://github.com/radicle-dev/radicle-interface/pull/344/commits/92ef5e768fbf64f2ae44e6e98ea592b3811ee901) and hide the `Button` for a `Loading` component in a nice transition

PR #347, opened 2022-08-29
Extract ToggleButton component

rudolfs:
It's used in multiple places in the app, and by extracting it into a separate component we can guarantee that it looks and behaves the same everywhere.


src/ToggleButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+  import { capitalize } from "@app/utils";
+
+  export interface ToggleButtonOption<T> {
+    count?: number;
+    title?: string;
+    value: T;
+  }
+</script>
+
+<script lang="ts" strictEvents>
```
sebastinez, line 11: nice `strictEvents`! learning every day something new 😎

src/ToggleButton.svelte
```diff
@@ -0,0 +1,72 @@
+<script lang="ts" context="module">
+  import { capitalize } from "@app/utils";
+
+  export interface ToggleButtonOption<T> {
+    count?: number;
```
sebastinez, line 5: When would `count` be `undefined`? It seems to me that we always define something as count.

src/ToggleButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+    title?: string;
```
sebastinez, line 6: I think we should be fine with using the capitalized `value` almost everywhere, do you have an example when we would use `title`?

src/ToggleButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+  type T = $$Generic;
+  import { createEventDispatcher } from "svelte";
+
+  export let options: ToggleButtonOption<T>[];
+  export let active: T;
+
+  const dispatch = createEventDispatcher<{ select: T }>();
+
+  function disabled(count?: number) {
+    return count === undefined || count === 0;
```
sebastinez, line 21: I think this could be removed and just check against 0 since IMO we wont handle `count === undefined`

src/ToggleButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+  }
+</style>
+
+<div class="wrapper">
+  {#each options as option}
+    <button
+      class="state-toggle"
+      on:click={ () => onSelect(option) }
+      disabled={ disabled(option.count) }
+      class:active={ active === option.value }>
```
sebastinez, line 65: Is there an ESLint rule for spacing in Svelte components?, we don't use spaces in Svelte props in any other component

src/ToggleButton.svelte
```diff
@@ -0,0 +1,72 @@
+<script lang="ts" context="module">
+  import { capitalize } from "@app/utils";
```
sebastinez, line 2: Out of curiosity, is there some benefit to import `capitalize` in the module context?

sebastinez (commented):
Left some comments out of curiosity 😅

src/ToggleButton.svelte
rudolfs: Yeah, me too. We were all waiting for typed events and generics in components, and now they're finally here, albeit still experimental. 🎉

src/ToggleButton.svelte
rudolfs: This is still a draft. I still wanted to refactor the tabs on the patch page, and they don't need a counter.
Screenshot 2022-08-30 at 10 21 07

src/ToggleButton.svelte
rudolfs: Good catch! 6d2574acb69f6723d33d65620d0187c463099579

src/ToggleButton.svelte
rudolfs: See https://github.com/radicle-dev/radicle-interface/pull/347#discussion_r958165344.

src/ToggleButton.svelte
rudolfs: Yeah, could remove it for now.

src/ToggleButton.svelte
sebastinez: Yeah sorry, looked into the PR as said to see your approach. Will leave you to it, surely an improvement over the current way

src/ToggleButton.svelte
rudolfs: Actually, we need it. Because we're using enums in some places and their value doesn't match the tab title. 🤦

src/ToggleButton.svelte
```diff
@@ -0,0 +1,68 @@
+<script lang="ts" context="module">
+  export interface ToggleButtonOption<T> {
+    title?: string;
+    count?: number;
+    value: T;
+  }
+</script>
+
+<script lang="ts" strictEvents>
+  type T = $$Generic;
```
cloudhead, line 10: :eyes: :mage_man:

src/base/projects/Commit.svelte
```diff
@@ -26,6 +26,7 @@
     padding: 1rem;
     background: var(--color-foreground-background-subtle);
     border-radius: var(--border-radius);
+    margin-bottom: 1.5rem;
```
cloudhead, line 29: Is this a visual change?

rudolfs: Only on the patch page, on the commit page the spacing remains the same.

The `<Changeset>` component had the margin inside it originally, I extracted it out and adjusted the parent elements, so visually it looks the same. Now we have more control over the positioning over the component in other places in the code.

Commit view (no visual changes):
Screenshot 2022-08-30 at 14 09 32

Adjusted the spacing in the patch view to look more natural (left is the proposed change):
Screenshot 2022-08-30 at 14 10 07


The spacing difference was more obvious when switching the tabs:

https://user-images.githubusercontent.com/158411/187440138-d999d08c-d6fc-42fd-886c-b833e66b8e88.mov

vs now:

https://user-images.githubusercontent.com/158411/187440243-d7454b04-2243-4080-8966-e9ee56799468.mov

cloudhead:
Great

PR #351, opened 2022-08-31
Reduce amount of global styling

rudolfs:
We get rid of some of the global styling, because it interacts with styling within components and is hard to reason about. There shouldn't be any visual differences.


cloudhead:
image

Is it just me? :thinking:

cloudhead:
actually this is from a different pr.

PR #354, opened 2022-09-01
Improve wallet mock

sebastinez:
This PR replaces the MockExtensionProvider we used for E2E tests with a simpler package `@rsksmart/mock-web3-provider`.
There are still improvements to be made, but we should have a more consistent testing, and less failed `check-integration-tests` in the CI


package.json
```diff
@@ -45,6 +45,7 @@
     "@gnosis.pm/safe-core-sdk-types": "^0.1.1",
     "@gnosis.pm/safe-service-client": "^0.1.1",
     "@radicle/gray-matter": "4.1.0",
+    "@rsksmart/mock-web3-provider": "^1.0.1",
```
rudolfs, line 48: Shouldn't this go in `devDependencies`?

PR #355, opened 2022-09-01
Improve CommitVerifiedBadge UX

rudolfs:
- debounce hover events to prevent flickering when mousing over multiple badges
- keep the card visible while hovering over it
- show verified badge on the Commit page

Left is how it is now, and right is what this PR proposes:

https://user-images.githubusercontent.com/158411/187939910-658ad61a-f33f-4de5-ab8c-9402bddc6c47.mov


src/base/projects/Commit/CommitVerifiedBadge.svelte
```diff
@@ -1,14 +1,25 @@
 <script lang="ts">
   import type { CommitMetadata } from "@app/commit";
-  import CommitAuthorship from "./CommitAuthorship.svelte";
+
+  import { debounce } from "lodash";
+
```
sebastinez, line 5: I like using lodash to avoid implementing it ourselves 😃 but we should definitly only load `/debounce` since `lodash` is huge.
Or maybe even `lodash.debounce`

```suggestion
  import { debounce } from "lodash/debounce";
```

sebastinez (commented):
Nice, just one comment on the `lodash` dependency

src/base/projects/Commit/CommitVerifiedBadge.svelte
rudolfs: Thing is, `loadash` has always been in our dependency tree. Just not explicitly at the top. Grep the lockfile for it and you'll see ; P

rudolfs: There's also other goodies in there that I'd love to start using in other places, e.g. `throttle`, `orderBy`, `isEqual`, etc.

rudolfs: When would we see a performance hit though? Because the bundler should be smart enough to tree-shake anything unneeded, no?

b3b3c06a03c05e1fb53cca93fbf460ee314b5e1c

sebastinez: Here is a good article about this, seems like it still is better if we import it with the dash
https://blog.battlefy.com/tree-shaking-lodash-with-vite

rudolfs: Ugh! Good to know! 🙇

src/base/projects/Commit/CommitVerifiedBadge.svelte
```diff
@@ -1,14 +1,25 @@
...
+  import debounce from "lodash/debounce";
+
   import Badge from "@app/Badge.svelte";
+  import CommitAuthorship from "./CommitAuthorship.svelte";
 
   export let commit: CommitMetadata;
 
-  let hover = false;
+  let visible = false;
+  const showDelay = 50; // ms
```
cloudhead, line 12: A little sluggish, I'd prefer something closer to 1-2 frames on a 60hz monitor, ie. 16-33ms

PR #358, opened 2022-09-02
Show loading indicators after a delay

rudolfs:
This prevents from loading indicators flickering for a split second during a page draw. For requests that take longer than 200ms we show the loading animation.

Before:

https://user-images.githubusercontent.com/158411/188100584-482797a9-ceb2-4a5e-95aa-05a4ac057959.mov

After:

https://user-images.githubusercontent.com/158411/188100630-7d2a2e40-0799-482d-81b1-c976951778e5.mov

### DIFF
--- a/src/Loading.svelte
+++ b/src/Loading.svelte
@@ -1,10 +1,22 @@
 <script lang="ts">
+  import { onDestroy } from "svelte";
+
   export let small = false;
   export let color = "secondary";
   export let center = false;
   export let fadeIn = false;
   export let margins = false;
   export let condensed = false;
+
+  let show: boolean = false;
+
+  const timeout = window.setTimeout(() => {
+    show = true;
+  }, 200);
+
+  onDestroy(() => {
+    window.clearTimeout(timeout);
+  });
 </script>
 
 <style>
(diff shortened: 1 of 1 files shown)


src/Loading.svelte
```diff
@@ -136,18 +148,20 @@
...
+      class:condensed>
+      <div class="bounce1" style="background-color: var(--color-{color})" />
+      {#if !condensed}
+        <div class="bounce2" style="background-color: var(--color-{color})" />
+        <div class="bounce3" style="background-color: var(--color-{color})" />
+      {/if}
+    </div>
   </div>
-</div>
+{/if}
```
cloudhead, line 167: Can we do this with a class instead? `class:show`?

rudolfs: It's doable, but why would we? Wrapping it in `{#if}` prevents the DOM elements to be rendered, making it less performant and also less _idiomatic_ Svelte.

```
diff --git a/src/Loading.svelte b/src/Loading.svelte
index 73957a3..6166af4 100644
--- a/src/Loading.svelte
+++ b/src/Loading.svelte
@@ -21,7 +21,7 @@

 <style>
   .container {
-    display: flex;
+    display: none;
     align-items: center;
     justify-content: center;
     height: 100%;
@@ -146,22 +146,24 @@
       opacity: 1;
     }
   }
+
+  .show {
+    display: flex;
+  }
 </style>

-{#if show}
-  <div class="container">
-    <div
-      class="spinner"
-      class:fade-in={fadeIn}
-      class:small
-      class:center
-      class:margins
-      class:condensed>
-      <div class="bounce1" style="background-color: var(--color-{color})" />
-      {#if !condensed}
-        <div class="bounce2" style="background-color: var(--color-{color})" />
-        <div class="bounce3" style="background-color: var(--color-{color})" />
-      {/if}
-    </div>
+<div class:show class="container">
+  <div
+    class="spinner"
+    class:fade-in={fadeIn}
+    class:small
+    class:center
+    class:margins
+    class:condensed>
+    <div class="bounce1" style="background-color: var(--color-{color})" />
+    {#if !condensed}
+      <div class="bounce2" style="background-color: var(--color-{color})" />
+      <div class="bounce3" style="background-color: var(--color-{color})" />
+    {/if}
   </div>
-{/if}
+</div>
```

cloudhead: Cool let's leave it, I think for a loading indicator it's fine as-is.

PR #359, opened 2022-09-02
Remove deprecated org parts

sebastinez:
- I didn't remove the `Org` class nor the `anchorsAccount` information on `Profile` since there are Orgs using the current implementation who would loose this information.
- I think we should at least inform these Orgs and have a migration plan defined, once the next version of Orgs rolls out, before we remove visible information on current Orgs

If you feel that too much has been axed, feel free to let me know


rudolfs:
Is the `profile.id` ever set?

https://github.com/radicle-dev/radicle-interface/blob/master/src/Profile.svelte#L244

rudolfs:
I wonder if we can also remove the old attestation presentation as well?

Screenshot 2022-09-05 at 16 30 33

sebastinez:
> I wonder if we can also remove the old attestation presentation as well?

I think we should keep the Radicle URN since we also show it for individual users, and if an org defined their Radicle URN in their ENS profile then that's ok too, IMO.

rudolfs:
These seem to be unused now:
```
src/Card.svelte
src/Cards.svelte
```

cloudhead:
Needs rebase.

sebastinez:
> Needs rebase.

Is done ✅

PR #360, opened 2022-09-02
Add PWA support

sebastinez:
In the works


sebastinez:
Closed for now, since we don't know in which direction we wanna go with local hosting

PR #362, opened 2022-09-05
Extract Button component

rudolfs:
We avoid using brittle global styling by extracting buttons into a separate component.

Screenshot 2022-09-05 at 15 38 35

### DIFF
--- a/src/Connect.svelte
+++ b/src/Connect.svelte
@@ -5,11 +5,11 @@
   import Error from "@app/Error.svelte";
   import type { Config } from "@app/config";
   import ConnectWallet from "@app/components/Modal/ConnectWallet.svelte";
+  import Button from "@app/Button.svelte";
 
   export let caption = "Connect";
-  export let className = "";
-  export let style = "";
   export let config: Config;
+  export let buttonVariant: "foreground" | "primary";
 
   let error: Err | null = null;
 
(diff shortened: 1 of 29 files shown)


src/Connect.svelte
```diff
@@ -34,18 +34,17 @@
...
 
-<button
-  on:click|stopPropagation={onConnect}
-  {style}
-  class="connect {className}"
+<Button
+  on:click={onConnect}
+  variant={buttonVariant}
   disabled={connecting}
   data-waiting={connecting || null}>
```
cloudhead, line 41: Data properties won't make it through as-is

rudolfs: Good catch! 4ebbf94d3205f573f786863e9a10c682c38a04d0

cloudhead:
I think this is good-to-go, code-wise, just need to make sure there's no design regressions since a lot of styles moved around!

cloudhead:
Here's one I'm guessing from this PR:
image

rudolfs:
I actually compared most of the screens with production and they lined up.

Screenshot 2022-09-06 at 11 32 13


Screenshot 2022-09-06 at 11 33 04

The only screens I didn't test were Orgs, because we're removing them anyway in https://github.com/radicle-dev/radicle-interface/pull/359.

cloudhead:
Ah damn, then this must come from some other change :thinking:

rudolfs:
> Ah damn, then this must come from some other change 🤔

Very low on my todo list, but I wanted to set up Playwright at some point and make it do screenshots of some of the happy-paths, so that we can spot these regressions more easily with screenshot diffing.

PR #370, opened 2022-09-07
Refactor global typography styles

rudolfs:
This is the first step in refactoring typography, there should not be any visual differences when this is applied.

The following changes have been made:

- Prefix all typography related global CSS classes with a unique prefix `.typo-` -- this allows us to grep the codebase and see exactly where they are applied. Another benefit of this convention is that it helps to avoid accidentally mixing global and local styles. Having global CSS classes that sometimes overlap with component CSS classes makes it very hard to reason about the resulting visuals and making changes unpredictable.

- Introduce CSS variables for standardised typography sizes and apply them to the whole codebase consistently. This way we can adjust the size in one place and it will be applied everywhere.

- Avoid using predefined HTML tags which might have differing styling from browser to browser, e.g. `<h1>`, `<strong>`. We should stick to applying styling either by adding a specific global CSS class or using the predefined CSS variables, that way we can be sure that the styles won't be overriden by browser defaults.

- Have a minimal set of modifiers `.typo-monospace` and `.typo-bold` and apply them consistently. Since `.typo-monospace` is defined via a CSS variable, we'll be able to switch out fonts via a user setting in the future.

- Move all global typography styles to its own file. This makes it easier to find all the typography related definitions for newcomers.

Safari:

https://user-images.githubusercontent.com/158411/188924436-593a4b52-9ac4-46fd-b8ff-0ea873156093.mov

Chrome:

https://user-images.githubusercontent.com/158411/188924658-ccf01586-c798-40bd-af0e-a10222748fd4.mov

Firefox:

https://user-images.githubusercontent.com/158411/188924870-319bea2e-1fe6-42da-b758-34a743c1c4fa.mov


cloudhead:
I like this. I'd prefer if we used `text-` as a prefix though, since `typo` means an error :)

rudolfs:
> I like this. I'd prefer if we used text- as a prefix though, since typo means an error :)

Ha. Yeah I thought about this. The problem with `text-` is that it matches other stuff when grepping:
```
src/Modal.svelte
38:    text-align: center;
63:    text-align: center;
64:    text-overflow: ellipsis;
78:    text-overflow: ellipsis;
84:    text-align: center;
```

Could we use something else that's unique? Was thinking about `typography-`, but it's kinda long.

PR #371, opened 2022-09-07
Fix emoji parser

sebastinez:
Fixes a regex error, since I only used a character set and not a group the return value of `.exec(src)` is not the one we needed..

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -810,7 +810,7 @@ const markdownEmojiTokenizer: marked.TokenizerExtension = {
     return src.indexOf(":");
   },
   tokenizer(src: string) {
-    const rule = /^:[\w+-]+:/;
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -825,6 +825,7 @@ const markdownEmojiTokenizer: marked.TokenizerExtension = {
 const markdownEmojiRenderer: marked.RendererExtension = {
   name: "emoji",
   renderer(token: marked.Tokens.Generic) {
+    console.log(token.emoji);
```
cloudhead, line 828: please my console

sebastinez: Done in [4cd702e](https://github.com/radicle-dev/radicle-interface/pull/371/commits/4cd702e412c295a7a2d65e26403f05cdc32d72d0)

PR #383, opened 2022-09-13
Make modal corners rounded

rudolfs:
All other elements in the app have rounded corners:

Screenshot 2022-09-13 at 12 16 40
Screenshot 2022-09-13 at 12 16 44
Screenshot 2022-09-13 at 12 16 47
Screenshot 2022-09-13 at 12 16 56
Screenshot 2022-09-13 at 12 17 06
Screenshot 2022-09-13 at 12 17 14
Screenshot 2022-09-13 at 12 17 19

It feels weird to me that the modal is the only element that has sharp corners.

Screenshot 2022-09-13 at 12 18 08

Thus I propose to make them consistent with the rest of the design:
Screenshot 2022-09-13 at 12 14 39
Screenshot 2022-09-13 at 12 14 45
Screenshot 2022-09-13 at 12 14 53
Screenshot 2022-09-13 at 12 15 15

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -68,7 +68,7 @@
   --color-glow-error: #ff555522;
   --color-scrollbar: var(--color-foreground-subtler);
 
-  --border-radius: 0.75rem;
+  --border-radius-tiny: 0.75rem;
   --border-radius-small: 0.5rem;
   --border-radius-regular: 1rem;
   --border-radius-round: 10rem;
@@ -129,7 +129,7 @@ body {
 }
 *::-webkit-scrollbar-thumb {
   background: transparent;
-  border-radius: var(--border-radius);
+  border-radius: var(--border-radius-tiny);
 }
 *::-webkit-scrollbar-corner {
   background: transparent;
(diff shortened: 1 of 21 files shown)


brandonhaslegs (approved):
Looks good. Only think maybe we should make the two border radii css variables?

rudolfs:
> Looks good. Only think maybe we should make the two border radii css variables?

Good point: 7916f142f54abcff1224cd4fb540433b0ff14d52  450a8dea13737f64578af7de57479146f154ae24

cloudhead:
Hmm but the window view area itself is not rounded :thinking:

rudolfs:
> Hmm but the window view area itself is not rounded 🤔

On my machine it is 😅 

mymachine

brandonhaslegs:
> Hmm but the window view area itself is not rounded :thinking: 

This is a joke right?

cloudhead:
Though:
image

rudolfs:
> Though: image

lol, u got me

cloudhead:
Not a fan of this change tbh, looks weird to me.

rudolfs:
> Not a fan of this change tbh, looks weird to me.

No biggie 🤷

PR #390, opened 2022-09-15
Don't delay loaders in source browsing file tree

rudolfs:
https://user-images.githubusercontent.com/158411/190404430-1b699e04-9360-48b2-a284-75cabacd719f.mov


rudolfs:
Let's better check if it's possible to do the preloading for folders.

PR #402, opened 2022-09-21
Fix style regression in clone button

rudolfs:
Fixes regression introduced in https://github.com/radicle-dev/radicle-interface/commit/d32bad5b62a14413d4a3511b4c1f48d3465b3293.

Before this fix:
Screenshot 2022-09-21 at 15 10 56
Screenshot 2022-09-21 at 15 11 00

After this fix:
Screenshot 2022-09-21 at 15 11 15
Screenshot 2022-09-21 at 15 11 21

### DIFF
--- a/src/base/projects/CloneButton.svelte
+++ b/src/base/projects/CloneButton.svelte
@@ -53,9 +53,7 @@
     align-items: center;
   }
   .clone-url {
-    background: var(--color-yellow-background-solid);
     border-radius: var(--border-radius-small);
-    color: var(--color-yellow);
     font-family: var(--font-family-monospace);
     font-size: var(--font-size-tiny);
     height: 2rem;
@@ -64,12 +62,16 @@
     text-overflow: ellipsis;
     white-space: nowrap;
     line-height: 1.4;
+    color: var(--color-caution-6);
+    background: linear-gradient(var(--color-caution-2), var(--color-caution-2)),
+      linear-gradient(var(--color-background), var(--color-background));
   }
   .clipboard {
     position: absolute;
     right: 0;
-    color: var(--color-yellow);
-    background: var(--color-yellow-background-solid);
+    color: var(--color-caution-6);
+    background: linear-gradient(var(--color-caution-2), var(--color-caution-2)),
+      linear-gradient(var(--color-background), var(--color-background));
     visibility: hidden;
     width: 4rem;
     height: 2rem;


src/base/projects/CloneButton.svelte
```diff
@@ -64,12 +62,16 @@
...
+    background: linear-gradient(var(--color-caution-2), var(--color-caution-2)),
+      linear-gradient(var(--color-background), var(--color-background));
   }
   .clipboard {
     position: absolute;
     right: 0;
-    color: var(--color-yellow);
-    background: var(--color-yellow-background-solid);
+    color: var(--color-caution-6);
+    background: linear-gradient(var(--color-caution-2), var(--color-caution-2)),
```
sebastinez, line 73: Why do we need two `linear-gradient` here and in `.clone-url`? Couldn't this be solved with `background: var(--color-caution-2);`

sebastinez (approved):
It looks neat, left a comment out of curiosity on the implementation

src/base/projects/CloneButton.svelte
rudolfs: It's a workaround to get the correct color because our color shades are transparencies. You get a different color when you layer multiple transparencies. We could also have flattened all our colors to solid values, but then we'd lose the flexibility to tweak them without introducing a transpile step (SCSS or the likes).

```diff
diff --git a/src/base/projects/CloneButton.svelte b/src/base/projects/CloneButton.svelte
index 107c398..0bd04f3 100644
--- a/src/base/projects/CloneButton.svelte
+++ b/src/base/projects/CloneButton.svelte
@@ -63,15 +63,13 @@
     white-space: nowrap;
     line-height: 1.4;
     color: var(--color-caution-6);
-    background: linear-gradient(var(--color-caution-2), var(--color-caution-2)),
-      linear-gradient(var(--color-background), var(--color-background));
+    background: var(--color-caution-2);
   }
   .clipboard {
     position: absolute;
     right: 0;
     color: var(--color-caution-6);
-    background: linear-gradient(var(--color-caution-2), var(--color-caution-2)),
-      linear-gradient(var(--color-background), var(--color-background));
+    background: var(--color-caution-2);
     visibility: hidden;
     width: 4rem;
     height: 2rem;
```

Screenshot 2022-09-21 at 15 43 16
vs with `linear-gradient` hack:
Screenshot 2022-09-21 at 15 43 27

sebastinez: Ok understood, thanks for the clarification 🙇

PR #406, opened 2022-09-22
Improve pinned projects

rudolfs:
Show "No description" when the project description is missing.

> @joelhans: Also I saw that you added astrolabe to the homepage, which is very exciting because that's my project(!), but I really messed that repo up and ended up moving to the willow seed node instead, so maybe we switch the link to point there instead? https://app.radicle.network/seeds/willow.radicle.garden/rad:git:hnrkrwewct7pb1biwmf59qa3rzod8mauundzy

Screenshot 2022-09-22 at 11 28 13

### DIFF
--- a/src/base/projects/Widget.svelte
+++ b/src/base/projects/Widget.svelte
@@ -114,7 +114,11 @@
     <div class="id">
       <span class="name">{project.name}</span>
     </div>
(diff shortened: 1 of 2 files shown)


cloudhead (approved):
perf

PR #407, opened 2022-09-22
Gracefully handle API quota exhaustion

sebastinez:
We had to rewrite `querySubgraphWithRetry` due to flaws in the return value logic and how the recursion took place.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -377,7 +377,7 @@ export async function querySubgraphWithRetry(
   query: string,
   variables: Record<string, any>,
   retries = 3,
-): Promise<null | any> {
+): Promise<Record<string, any> | null> {
   const response = await fetch(url, {
     method: "POST",
     headers: {
@@ -393,11 +393,14 @@ export async function querySubgraphWithRetry(
   if (json.errors) {
     console.error("querySubgraph:", json.errors);
 
-    if (retries > 0) querySubgraphWithRetry(url, query, variables, retries - 1);
-    else return null;
+    if (retries > 0) {
+      return querySubgraphWithRetry(url, query, variables, retries - 1);
+    } else {
+      return null;
+    }
+  } else {
+    return json.data;
   }
-
-  return json.data;
 }
 
 export const querySubgraph = cache.cached(
@@ -589,7 +592,7 @@ export async function isSafe(
     addr: address.toLowerCase(),
   });
 
-  return query.safe !== null ? true : false;
+  return query?.safe !== null ? true : false;
 }
 
 // Get a Gnosis Safe at an address.
--- a/src/base/orgs/Org.ts
+++ b/src/base/orgs/Org.ts
@@ -203,16 +203,24 @@ export class Org {
       GetSafesByOwners,
       { owners: [owner] },
     );
-    const safes = safesByOwner.safes.reduce(
-      (prev: any, curr: Safe) => prev.concat(curr.id),
-      [],
-    );
+
+    let safes = [];
+    if (safesByOwner && safesByOwner.safes.length > 0) {
+      safes = safesByOwner.safes.reduce(
+        (prev: any, curr: Safe) => prev.concat(curr.id),
+        [],
+      );
+    }
+
     const orgsByOwner = await utils.querySubgraph(
       config.orgs.subgraph,
       GetOrgsByOwners,
       { owners: [...safes, owner] },
     );
-    const orgs: { id: string; owner: string }[] = [...orgsByOwner.orgs];
+    let orgs: { id: string; owner: string }[] = [];
+    if (orgsByOwner && orgsByOwner.orgs) {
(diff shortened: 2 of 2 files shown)


rudolfs:
I think we get the `undefined` back because there was a `return` statement missing where we enter the recursion.
```diff
diff --git a/src/utils.ts b/src/utils.ts
index ae5217e..0e457e1 100644
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -393,11 +393,14 @@ export async function querySubgraphWithRetry(
   if (json.errors) {
     console.error("querySubgraph:", json.errors);

-    if (retries > 0) querySubgraphWithRetry(url, query, variables, retries - 1);
-    else return null;
+    if (retries > 0) {
+      return querySubgraphWithRetry(url, query, variables, retries - 1); // <--- this `return` was missing
+    } else {
+      return null;
+    }
+  } else {
+    return json.data;
   }
-
-  return json.data;
 }

 export const querySubgraph = cache.cached(
```

sebastinez:
> I think we get the `undefined` back because there was a `return` statement missing where we enter the recursion.

Yeah you're right, also we didn't `await` the recursive query, so we returned `undefined` before being able to traverse all the functions..

I also improved the function type signature so `querySubgraph` also returns `Promise<null | any>`

src/utils.ts
```diff
@@ -393,14 +393,24 @@ export async function querySubgraphWithRetry(
   if (json.errors) {
     console.error("querySubgraph:", json.errors);
 
-    if (retries > 0) querySubgraphWithRetry(url, query, variables, retries - 1);
-    else return null;
+    if (retries > 0) {
+      console.debug("Retrying query subgraph, retry countdown:", retries);
```
rudolfs, line 397: 👀

src/utils.ts
```diff
@@ -393,14 +393,24 @@ export async function querySubgraphWithRetry(
...
+      await querySubgraphWithRetry(url, query, variables, retries - 1);
+    } else {
+      return null;
+    }
+  } else {
+    return json.data;
   }
 
-  return json.data;
+  return null;
```
rudolfs, line 406: Isn't this unreachable?

rudolfs:
> Yeah you're right, also we didn't await the recursive query, so we returned undefined before being able to traverse all the functions.

Hm, I'm not sure that `await` is actually necessary, but probably doesn't hurt to have it.

> I also improved the function type signature so querySubgraph also returns Promise<null | any>

Nice. I still wonder why TS wouldn't infer it, it seems to work in the other cases. 🤷

src/utils.ts
```diff
@@ -393,14 +393,24 @@ export async function querySubgraphWithRetry(
   if (json.errors) {
     console.error("querySubgraph:", json.errors);
 
-    if (retries > 0) querySubgraphWithRetry(url, query, variables, retries - 1);
-    else return null;
+    if (retries > 0) {
+      console.debug("Retrying query subgraph, retry countdown:", retries);
+      await querySubgraphWithRetry(url, query, variables, retries - 1);
```
rudolfs, line 398:
```suggestion
      return await querySubgraphWithRetry(url, query, variables, retries - 1);
```

rudolfs: I think if you don't return here, then it would always cache `null` if the function goes into retry.

sebastinez: Is fixed in [b40a9de](https://github.com/radicle-dev/radicle-interface/pull/407/commits/b40a9de845332d499c0e5d8814bf771ad786c448)

src/utils.ts
```diff
@@ -588,8 +596,9 @@ export async function isSafe(
   const query = await querySubgraph(config.orgs.subgraph, GetSafe, {
     addr: address.toLowerCase(),
   });
+  console.log(query);
```
rudolfs, line 599: 👀

sebastinez: fixed in 57a2554

src/utils.ts
```diff
@@ -393,14 +393,22 @@ export async function querySubgraphWithRetry(
...
-
-  return json.data;
 }
 
-export const querySubgraph = cache.cached(
+export const querySubgraph: (
+  url: string,
+  query: string,
+  variables: Record<string, any>,
+) => Promise<any | null> = cache.cached(
```
rudolfs, line 411: Unfortunately this still doesn't give us any type safety for some reason. 🤔 

Screenshot 2022-09-22 at 18 41 14

So when I do:
```diff
diff --git a/src/utils.ts b/src/utils.ts
index 4e08201..f0b9a91 100644
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -597,7 +597,7 @@ export async function isSafe(
     addr: address.toLowerCase(),
   });

-  return query?.safe !== null ? true : false;
+  return query.safe !== null ? true : false;
 }
```
I still don't get a type error.

rudolfs (approved):
Nice work!

PR #410, opened 2022-09-23
Fix `isSafe` util function

sebastinez:
We were calling `return query?.safe !== null ? true : false` but since `query` can become `undefined` through the optional chaining operator, we were returning `true` if the query wasn't able to retrieve anything.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -592,7 +592,7 @@ export async function isSafe(
     addr: address.toLowerCase(),
   });
 
-  return query?.safe !== null ? true : false;
+  return Boolean(query?.safe);
 }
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -592,7 +592,7 @@ export async function isSafe(
     addr: address.toLowerCase(),
   });
 
-  return query?.safe !== null ? true : false;
+  return query?.safe ? true : false;
```
cloudhead, line 595:
```suggestion
  return Boolean(query?.safe)
```

rudolfs (approved):
😬

PR #411, opened 2022-09-23
Show delegate badge in commits

### LINKED ISSUE #404: Show "delegate" badge in commits
It would be nice to see when a commit was signed by a delegate, in for eg. the verified popup:
image

It could say eg. `lorem [delegate] committed`

sebastinez:
- Shows a delegate `Badge` in `CommitAuthorship` if available.
- Changes `width: 14rem` to `min-width: 14rem` to allow space for horizontal text without the need of wrapping it.
  e.g. <Avatar> cloudhead [delegate] committed already was in need of being wrapped which doesn't look nice.

Closes #404

### DIFF
--- a/src/base/projects/Commit/CommitAuthorship.svelte
+++ b/src/base/projects/Commit/CommitAuthorship.svelte
@@ -1,16 +1,20 @@
 <script lang="ts">
   import type { CommitMetadata } from "@app/commit";
+
+  import Badge from "@app/Badge.svelte";
   import { formatTimestamp, gravatarURL } from "@app/utils";
 
   export let commit: CommitMetadata;
-  export let showTime = true;
-  export let showAuthor = true;
+  export let noTime = false;
+  export let noAuthor = false;
+  export let noDelegate = false;
 </script>
 
 <style>
   .authorship {
     display: flex;
     align-items: center;
+    gap: 0.25rem;
     color: var(--color-foreground-5);
     padding: 0.125rem 0;
   }
@@ -23,10 +27,6 @@
     width: 1rem;
     height: 1rem;
     border-radius: var(--border-radius);
-    margin-right: 0.25rem;
-  }
-  .authorship .avatar:not(:first-child) {
-    margin-left: 0.125rem;
   }
 
   @media (max-width: 720px) {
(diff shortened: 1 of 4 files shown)


rudolfs:
I'm not sure the original issue asked for adding the delegate badge to the commit card?
And in the hover card the badge could use some padding.

Screenshot 2022-09-26 at 13 45 44

rudolfs:
It would be helpful if you could add before/after screenshots to PRs that change visuals, so it's easier for me as a reviewer to see what the intent was.

sebastinez:
> I'm not sure the original issue asked for adding the delegate badge to the commit card? And in the hover card the badge could use some padding.

IMO only showing it in the `CommitVerifiedBadge` popup is a bit of sparse and I don't think many users will hover looking for delegate information, maybe we can change the variant to `foreground` to reduce the amount of colors in the listing?

Bildschirmfoto 2022-09-26 um 14 31 17
Bildschirmfoto 2022-09-26 um 14 30 46

Since we are touching the `CommitAuthorship` maybe we should do something wrt, that a bold committer address means that the committer has its identity stored on the seed node, while regular font weight is just a commit author / mail.
I feel like that's not clear to the end user?

Bildschirmfoto 2022-09-26 um 14 40 21

cloudhead:
I think it's sufficient to show it in the popup, and potentially later we can have a user popup as well that shows it.

sebastinez:
Ok did a last work on this PR, and I think it came out pretty nicely, was also able to improve a bit the whitespacing.
Here a link to a diff viewer of the `CommitTeaser` before and after.
https://www.textcompare.org/image/?id=633440c0b479ffb94a9a21f8

Also here the images for your manual comparison
app radicle xyz_seeds_seed alt-clients radicle xyz_rad_git_hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao_history_38704b4fe141d7214dd02ba9717251f1eae2910a
localhost_3000_seeds_seed alt-clients radicle xyz_rad_git_hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao_history_38704b4fe141d7214dd02ba9717251f1eae2910a

And here an before and after image of the `VerifiedBadge` popup
Bildschirmfoto 2022-09-28 um 14 49 35
Bildschirmfoto 2022-09-28 um 14 51 07

cloudhead (approved):
LGTM!

src/base/projects/Commit/CommitAuthorship.svelte
```diff
@@ -1,16 +1,20 @@
...
+
+  import Badge from "@app/Badge.svelte";
   import { formatTimestamp, gravatarURL } from "@app/utils";
 
   export let commit: CommitMetadata;
-  export let showTime = true;
-  export let showAuthor = true;
+  export let noTime = false;
+  export let noAuthor = false;
+  export let noDelegate = false;
```
rudolfs, line 10: Not a huge deal, but my preference would be to leave these the way they were, because it makes the resulting code easier to read by avoiding double negatives.

E.g. `{#if !noAuthor}` I read in my mind as "no no author", is a bit more mental strain than `{#if !showAuthor}` which i read as "don't show author".

rudolfs: Alright. We'll keep it.

PR #416, opened 2022-09-26
Lowercase ENS name search query

sebastinez:
Since ENS names go through a UTS46 normalisation process case-folds labels before hashing them.
So we shouldn't be ENS querying uppercase names.

### DIFF
--- a/src/resolver.ts
+++ b/src/resolver.ts
@@ -53,9 +53,14 @@ export async function resolve(
         results.projects.push(...projects);
 
         // ========= ENS Names =========
+        const normalizedQuery = q.toLowerCase();
         let profile: Profile | null;
         try {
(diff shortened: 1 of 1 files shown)


src/resolver.ts
```diff
@@ -53,6 +53,7 @@ export async function resolve(
         results.projects.push(...projects);
 
         // ========= ENS Names =========
+        q = q.toLowerCase(); // Lowercasing query to conform with ENS name syntax
```
rudolfs, line 56: Nit: could we instead of mutating the query variable just create a new one, and then use that?

```suggestion
     normalizedQuery = q.toLowerCase();
```

sebastinez: Yeah sure, we can avoid mutating the state, makes sense. 👍

PR #417, opened 2022-09-26
Upgrade Cypress dependency

rudolfs:
This gets rid of the `testing-library/*` and `cypress-vite-devserver` dependencies. We also refactor some of the component specs to make them more idiomatic Cypress and extract state from some components to make them easier to test.


sebastinez (approved):
Love it!

Just one question, why do you prefer
`cy.get("button").contains("Back").should("be.visible");`
over
`cy.findByText("Back");`

It seems to me that the latter is shorter and fulfills the same goal.

rudolfs:
> Love it!
> 
> Just one question, why do you prefer `cy.get("button").contains("Back").should("be.visible");` over `cy.findByText("Back");`
> 
> It seems to me that the latter is shorter and fulfills the same goal.

Because I removed the `testing-library/*` deps which it was part of. IMO the shortcut of using `findByText` over `contains + should` wasn't worth the added extra transitive dependencies. I also hope we switch to playwright eventually which has much better DX out of the box.

sebastinez:
> Because I removed the `testing-library/*` deps which it was part of

Ah you're right. 🙇‍♂️

PR #418, opened 2022-09-26
Add shake animation to search component when no results

sebastinez:
Adds a shake animation to the `Search` component, that triggers either when no search result have been found, or when the search failed due to a different error, in which case a `console.error` is being printed

https://user-images.githubusercontent.com/7912302/192304082-d4464e03-1df5-48c0-9f99-bf0e453cf632.mov

Eventually for a different PR we should think about how we want to inform the user about fetching issues on a smaller component like `Search`. Maybe a global error store?


src/Search.svelte
```diff
@@ -1,21 +1,27 @@
...
+  let shake = false;
 
   const dispatch = createEventDispatcher();
   const handleKeydown = async (event: KeyboardEvent) => {
     if (event.key === "Enter") {
       searching = true;
-      results = await resolve(input, config);
+      const results = await resolve(input, config).catch(() => {
+        shake = true;
+        debounce(() => (shake = false), 500)();
```
rudolfs, line 23: 👏

rudolfs (commented):
Nice!

I'd also disable the input while searching:
```diff
diff --git a/src/Search.svelte b/src/Search.svelte
index 0ee5873..66b853d 100644
--- a/src/Search.svelte
+++ b/src/Search.svelte
@@ -58,6 +58,7 @@
 <div class:horizontal-shake={shake}>
   <TextInput
     variant="dashed"
+    disabled={searching}
     bind:value={input}
     on:keydown={handleKeydown}
     placeholder="Search a name or address…">
```

And maybe only clear the input when the search is successful?

rudolfs:
> Hmm interesting, do you mean so the user can try to search the same term again?

Yeah, I was thinking that maybe it's nicer to leave the user input there, in case the user mistyped it and wants to correct it.

PR #421, opened 2022-09-28
Refactor search

rudolfs:
- type events that are handed between components
- simplify the `resolve` function interface
  - before we had return values and exceptions, now just a single union
    type as return value
  - move navigation to the Search component to reduce its
    responsibilities
  - flatten and simplify multi-level branching
- shuffle around code
  - move the `resolve` function to the `Search` component module,
    because it's only used there


src/Header.svelte
```diff
@@ -27,14 +27,9 @@
   export let config: Config;
 
   let query: string;
-  let results: ResolvedSearch;
+  let results: ProjectsAndProfiles | undefined = undefined;
```
sebastinez, line 30: I think this could also be done with using `null` instead of `undefined` as used in other components in the codebase

```suggestion
  let results: ProjectsAndProfiles | null = null;
```

sebastinez (commented):
Looks nice, and seems to work so far 😁 
Not sure if I'm a big fan of moving the big `resolve` fn into the `Search` component, but I understand the reasoning behind.
If you want we can talk tomorrow together over it

src/Header.svelte
rudolfs: 2697b8a7cd213c66eda38f2485c59683b6571cb9

src/Search.svelte
```diff
@@ -1,42 +1,196 @@
...
+      try {
+        const projects = await Project.getMulti(projectOnSeeds);
+        projectsAndProfiles.projects.push(...projects);
+      } catch {
+        // TODO: collect errors and forward to user.
+      }
+
+      try {
+        let params: string[];
+        if (normalizedQuery.includes(".")) {
```
sebastinez, line 72: I suppose you could use here the `isEnsName` util function.
```suggestion
        if (isEnsName(normalizedQuery, config)) {
```

sebastinez (approved):
Looks great to me.
Left one comment where I think we could use the `isEnsName` util function to check if it's already a correctly formatted ENS name

src/Search.svelte
rudolfs: 51853d2100558473d1f6bd62176291c6fd507185

PR #422, opened 2022-09-28
Add seed peer id to `rad clone` in `CloneButton`

sebastinez:
**Important**: Requires `radicle-cli v.0.7.0` to be stable before merging.
Replaces `rad clone` command provided by the `CloneButton` from

```
rad clone rad://clients.radicle.xyz/hnrkbtw9t1of4ykjy6er4qqwxtc54k9943eto
```
to
```
rad clone rad://yyncx743heiqf13ocz4dk373uuf4quiiscwfn75qb7d8rupjnkn7c@clients.radicle.xyz/hnrkbtw9t1of4ykjy6er4qqwxtc54k9943eto
```

### DIFF
--- a/src/base/projects/Header.svelte
+++ b/src/base/projects/Header.svelte
@@ -101,7 +101,7 @@
     on:branchChanged={event => updateRevision(event.detail)} />
 
   {#if seed.git.host}
-    <CloneButton seedHost={seed.git.host} {urn} />
+    <CloneButton seedHost={seed.git.host} seedId={seed.id} {urn} />
   {/if}
   <span>
     {#if seed.api.host}
--- a/src/base/projects/CloneButton.svelte
+++ b/src/base/projects/CloneButton.svelte
@@ -5,9 +5,12 @@
   import { closeFocused } from "@app/Floating.svelte";
 
   export let seedHost: string;
+  export let seedId: string;
   export let urn: string;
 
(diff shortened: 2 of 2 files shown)


src/base/projects/Header.svelte
```diff
@@ -101,7 +101,7 @@
     on:branchChanged={event => updateRevision(event.detail)} />
 
   {#if seed.git.host}
-    <CloneButton seedHost={seed.git.host} {urn} />
+    <CloneButton {seed} {urn} />
```
rudolfs, line 104: We should try to keep the interfaces of components lean.

That is, pass down only the required props and avoid forwarding "chunky" objects. This has the following benefits from my point of view:
  - we can see exactly which data goes into a component:
    - this makes reasoning about the surrounding code easier
    - this also adds friction to making components too complex, because you manually have to pass around props that you don't need, which in turn forces you to think more about what the component should actually do
    - reduces coupling and makes our components more composable
    - is potentially more resource efficient and performant
  - it makes it easier to test the component in isolation, because you don't have to set up complex objects

I think there are also exceptions to this rule, but generally I think we should stick to this. Wdyt @sebastinez?

sebastinez: Yeah I agree with you on that.
In this case, the `CloneButton` was starting to need almost the entire `Seed` class, but in the end I reverted to only use the seed id, so I think I can move back to pass only the required props down.

sebastinez:
Will close this PR for now, since the v0.7 `rad clone` will be probably already look like different

PR #424, opened 2022-09-29
Leave mobile search context menu open if search fails

rudolfs:
It would close the context menu before you can see the shake animation.

https://user-images.githubusercontent.com/158411/193039020-40068070-750e-44ed-b51e-92fa014cb4e2.mov


sebastinez (approved):
👍

PR #425, opened 2022-09-29
Retry a test twice before giving up

rudolfs:
We have some flaky tests in the test suite. This should take care of it.

E.g.: https://github.com/radicle-dev/radicle-interface/actions/runs/3151541498/attempts/1


sebastinez (approved):
LGTM

PR #430, opened 2022-09-30
Show fetch more button only when more projects are available

sebastinez:
- Fetches project count information from `/stats` http-api endpoint
- If less projects are shown than the project count, show a `More` button
- If the amount of projects shown is equal to the total amount stop showing it.
- If the seed does not support the `/stats` endpoint the button will shown until we're not able to fetch more projects.

### DIFF
--- a/src/base/seeds/View/Projects.svelte
+++ b/src/base/seeds/View/Projects.svelte
@@ -4,26 +4,33 @@
   import Widget from "@app/base/projects/Widget.svelte";
   import type { Profile } from "@app/profile";
   import type { ProjectInfo } from "@app/project";
-  import type { Seed } from "@app/base/seeds/Seed";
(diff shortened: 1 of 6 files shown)


cloudhead (approved):

src/base/seeds/View/Projects.svelte
```diff
@@ -49,7 +53,10 @@
 </style>
 
 <div class="projects">
-  <List items={projects} query={fetchMoreProjects}>
+  <List
+    bind:items={projects}
+    complete={projects.length === totalProjectCount}
```
rudolfs, line 58: We should hide the button in case a project gets added while the user is paginating through the list.
```suggestion
    complete={projects.length >= totalProjectCount}
```

rudolfs (approved):
Nice!

PR #432, opened 2022-10-02
Make file tree sticky

### LINKED ISSUE #125: Tree should be anchored to the left while source is scrolled

OsamaElHariri:
When browsing the source in a repo, the file tree is now always visible.

Tested on Linux: Firefox/Chromium, Android: Firefox/Chrome, MacOS: Safari/Brave.

The behaviour on mobile has not been changed.

The video below shows how it looks like with a long file and a long tree, a long file and a short tree, a short file and a short tree, and a short file and a long tree (where "short" means all of it fits on the screen)

[RadicleStickyFileTree.webm](https://user-images.githubusercontent.com/23722804/193448725-6cca7328-3e47-4567-baef-42cf6f7877a4.webm)

Closes #125

### DIFF
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -134,6 +134,11 @@
   nav {
     padding: 0 2rem;
   }
+  .sticky {
+    position: sticky;
+    top: 2rem;
+    max-height: 100vh;
+  }
 
   @media (max-width: 960px) {
     .container {
@@ -164,6 +169,9 @@
     .column-left-visible {
       display: block;
     }
+    .sticky {
+      max-height: initial;
+    }
   }
 </style>
 
@@ -183,7 +191,7 @@
   <div class="container center-content">
     {#if tree.entries.length > 0}
       <div class="column-left" class:column-left-visible={mobileFileTree}>
-        <div class="source-tree">
+        <div class="source-tree sticky">
           <Tree
             {tree}
             {path}


sebastinez (commented):
Hey Osama, thank you very much for your contribution! 👏 

It looks good to me, the only comment I have is that we could eventually remove the redefinition of the `.sticky` class in the media query, and define `.sticky` at the end of the style tag.

OsamaElHariri:
I agree, the two definitions are not ideal :confused: . I'm not sure I understand how to remove the definition in the media query and define it only once at the end of the style tag. I believe that would still have a side effect on the mobile view, correct?

In order to keep the mobile version as is, I had to redefine this class. I could have introduced a media query on `min-width: 720px`, but that's not used in the code anywhere so I figured I'd hold off to keep things consistent.

If a `min-width: 720px` is OK, I can update this so that the `.sticky` class is not introduced twice

sebastinez:
Yeah never mind you can't avoid redeclaring the class.

But you could reduce the amount of properties you have to set back to `initial` since `position: sticky` and `top: 2rem` works in your favor, you only need to set `max-height` back to `initial` to remove the vertical scrolling of the file tree on mobile.

```diff
diff --git a/src/base/projects/Browser.svelte b/src/base/projects/Browser.svelte
index 2693d8a..365158f 100644
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -170,8 +170,6 @@
       display: block;
     }
     .sticky {
-      position: initial;
-      top: initial;
       max-height: initial;
     }
   }
```

OsamaElHariri:
That's interesting, you're right :astonished: 

Looks like it's because the parent component of the sticky element only spans the file tree and not the whole thing. Nice :+1: 

Updated the PR

rudolfs (approved):
Nice! Thanks for the contribution!

rudolfs:
Merged via 9cb3928481f7579b88b7aba629e62b602fdd12ba.

PR #433, opened 2022-10-03
Add katex render to Markdown

### LINKED ISSUE #413: Markdown support for rendering math
Github supports an incredibly powerful and useful syntax for writing math, i.e I can write $a^2 + b^2 = c^2$ with `$a^2 + b^2 = c^2$`.

Since `rad-cli` does an incredible job in making the transition from github to radicle easy, I think it would be great to add support for this kind of rendering also in the radicle interface which currently doesn't support the feature:
(from https://app.radicle.xyz/seeds/pine.radicle.garden/rad:git:hnrkp5jgfxwxrffjsp148fyzdc457pa9ja38y/tree)
tmyxer__A_native_and_open-source_coin_mixer_for_the_Stellar_network_

sebastinez:
This PR:
- adds a marked extension that parses Katex expressions, and renders them using [katex](https://katex.org/).
- to be able to do so it adds also a stylesheet provided by katex.org
- also changes the markdown sanitizer from `xss` to `dompurify` for saner defaults and MathML support.

Here you can see the difference:

**Before**
Bildschirmfoto 2022-10-03 um 17 31 48
**After**
Bildschirmfoto 2022-10-03 um 17 31 25

Closes #413

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,6 +16,7 @@
     <meta property="og:image:height" content="240" />
     <meta property="og:type" content="website" />
     <link rel="stylesheet" type="text/css" href="/typography.css" />
+    <link rel="stylesheet" type="text/css" href="/katex.min.css" />
     <link rel="stylesheet" type="text/css" href="/colors.css" />
     <link rel="stylesheet" type="text/css" href="/elevations.css" />
     <link rel="stylesheet" type="text/css" href="/index.css" />
--- a/scripts/copy-katex-assets
+++ b/scripts/copy-katex-assets
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -Eeuo pipefail
+
+cp -r node_modules/katex/dist/katex.min.css public/katex.min.css
+cp -r node_modules/katex/dist/fonts/* public/fonts/
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,12 @@ build/
 node_modules/
 NOTES
 
+# KaTeX files
+*.min.css
+KaTeX_**.ttf
+KaTeX_**.woff
+KaTeX_**.woff2
+
 # Integration Tests
 cypress/screenshots
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,6 +6,7 @@
   "packages": {
     "": {
       "version": "1.0.0",
+      "hasInstallScript": true,
       "dependencies": {
         "@datamodels/identity-profile-basic": "^0.1.2",
         "@ethersproject/abstract-provider": "^5.4.0",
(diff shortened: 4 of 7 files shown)


index.html
```diff
@@ -16,6 +16,11 @@
     <meta property="og:image:height" content="240" />
     <meta property="og:type" content="website" />
     <link rel="stylesheet" type="text/css" href="/typography.css" />
+    <link
+      rel="stylesheet"
+      type="text/css"
+      href="https://cdn.jsdelivr.net/npm/katex@0.16.2/dist/katex.min.css"
+      crossorigin="anonymous" />
```
rudolfs, line 23: We should bundle this with our app and not rely on a 3rd party cdn. We don't do that for any other of our assets.

sebastinez: Done with 5026dcebe30dc655b3621f0bf4083e619dc97547

rudolfs: We should make pretier ignore it:

```
Checking formatting...
[warn] public/katex.min.css
[warn] Code style issues found in the above file. Forgot to run Prettier?
Error: Process completed with exit code 1.
```

sebastinez:
Did also rebase on latest dependencies changes

index.html
sebastinez: Yeah went with the non minified version. Maybe even better if we want to debug something katex related

public/katex.css
```diff
@@ -0,0 +1,1118 @@
+/* stylelint-disable font-family-no-missing-generic-family-keyword */
+@font-face {
+  font-family: "KaTeX_AMS";
+  src: url(fonts/KaTeX_AMS-Regular.woff2) format("woff2"),
```
rudolfs, line 4: We should probably add all of these fonts as well, no?

sebastinez: These are available at `node_modules/katex/dist/fonts` I don't think that we need to add them twice.

rudolfs: There is nothing that copies them automatically, similar to the CSS:
Screenshot 2022-10-04 at 09 23 20

cloudhead (commented):
LGTM

rudolfs:
Is there any way to resolve these warnings?

Screenshot 2022-10-04 at 10 35 15

public/katex.css
sebastinez: Ok did:
- add a postinstall script, where we copy the font files into the `public` directory
- add all the `KaTeX_*` fonts to `.gitignore` to avoid checking them in
- move back to a minified version of `katex.css` and add it to a new `.prettierignore` file

sebastinez:
> Is there any way to resolve these warnings?

As far as I was able to look into, it derives from unicode characters used in the latex formulas, which katex can parse..
Maybe there is a way to sanitize the input and replace e.g. `Four-Per-Em Space` with a simple whitespace but I would leave that for another PR

scripts/postinstall
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
rudolfs, line 1: Could we call this script `copy-katex-assets` instead of postinstall? We'll have more later on, would rather chain them with `&&` than stick everything into one.

scripts/copy-katex-assets
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+set -e
```
rudolfs, line 2:
```suggestion
set -Eeou pipefail
```

rudolfs (commented):
Nice work!

PR #434, opened 2022-10-04
Don't remove `padding-top` from `pre` code blocks at the beginning

### LINKED ISSUE #428: Small markdown padding bug in code block
image

sebastinez:
This PR excludes `pre` tags from being stripped from their `padding-top` style attribute, but keeps the resetting of `margin-top` to 0.

Closes #428


rudolfs (commented):
I think I'd prefer to reduce the catch-all selectors, because it's hard to reason about them.

```diff
diff --git a/src/Markdown.svelte b/src/Markdown.svelte
index d6f02bd..2d1dc62 100644
--- a/src/Markdown.svelte
+++ b/src/Markdown.svelte
@@ -72,6 +72,16 @@
     padding-left: 0.5rem;
   }

+  .markdown > :global(h1:first-child),
+  .markdown > :global(h2:first-child),
+  .markdown > :global(h3:first-child),
+  .markdown > :global(h4:first-child),
+  .markdown > :global(h5:first-child),
+  .markdown > :global(h6:first-child) {
+    padding-top: 0;
+    margin-top: 0;
+  }
+
   .markdown :global(h1),
   .markdown :global(h2),
   .markdown :global(h3),
@@ -148,6 +158,15 @@
     max-width: 100%;
   }

+  .markdown > :global(pre:only-child) {
+    margin: 0;
+  }
+
+  .markdown :global(pre code) {
+    background: none;
+    padding: 0;
+  }
+
   .markdown :global(code) {
     font-family: var(--font-family-monospace);
     font-size: var(--font-size-regular);
@@ -157,11 +176,6 @@
     padding: 0.125rem 0.25rem;
   }

-  .markdown :global(pre code) {
-    background: none;
-    padding: 0;
-  }
-
   .markdown :global(pre) {
     font-family: var(--font-family-monospace);
     font-size: var(--font-size-regular);
@@ -233,10 +247,6 @@
     padding: 0.5rem 1rem;
   }

-  .markdown > :global(*:first-child) {
-    padding-top: 0 !important;
-    margin-top: 0 !important;
-  }
   .markdown :global(dl dt) {
     font-style: italic;
     margin-top: 1rem;
```
Screenshot 2022-10-04 at 16 33 57

sebastinez:
> I think I'd prefer to reduce the catch-all selectors, because it's hard to reason about them.

Hmm ok yeah, I think if we want to remove the global catch-all selector we would also need to set `padding-top` and `margin-top` to 0 for the `.frontmatter` class, for `hr` and `table` tags?

rudolfs:
> Hmm ok yeah, I think if we want to remove the global catch-all selector we would also need to set padding-top and margin-top to 0 for the .frontmatter class, for hr and table tags? Since we add top paddings and margins to them.

I think `.frontmatter` is fine as it is, because it's outside of the `.markdown` element. `table`, `hr`, yeah, sure.

sebastinez:
> I think I'd prefer to reduce the catch-all selectors, because it's hard to reason about them.

Ok have addressed your feedback, and added `hr` and `table` to the selectors.
**Table as first component**
Bildschirmfoto 2022-10-04 um 21 07 26

**Horizontal line as first element**
Bildschirmfoto 2022-10-04 um 21 07 21

rudolfs:
Now I feel like we have to also add the `only-child` stuff to everything for the sake of completeness.

rudolfs (approved):

PR #435, opened 2022-10-04
Make sure svelte-check checks component specs

rudolfs:
I've previously noticed that sometimes my editor complained about issues that `svelte-check` would ignore. This would only happen in `src/**/*.spec.ts` files. By explicitly telling `svelte-check` to use `tsconfig.json` we check all files.


scripts/check
```diff
@@ -1,4 +1,5 @@
-#!/bin/sh
-set -e
+#!/bin/bash
+set -Eeou pipefail
```
cloudhead, line 2: Are these changes necessary? There's no pipe.

rudolfs: My thinking on this is that I use this shell boilerplate everywhere as it provides a safe default for any case. In my experience scripts usually evolve and then I forget to add this and stuff breaks. So I'd prefer to use `set -Eeou pipefail` in all shell scripts throughout the codebase.

Does that make sense?

scripts/check
```diff
@@ -1,4 +1,5 @@
...
 
-npx tsc --noEmit && npx svelte-check
+npx tsc --noEmit
+npx svelte-check --tsconfig tsconfig.json
```
sebastinez, line 5: Nice find!

After going through the other configuration options I'm thinking maybe we should also enable `--use-new-transformation` ? It catches a few more type issues in `Profile.svelte`

rudolfs: I thought about this and tried enabling it, but couldn't find a quick fix for the problems it found. Would rather leave it for either a follow-up or wait until they turn this feature on by default once it's not experimental anymore.

scripts/check
cloudhead: Yeah, the reason I don't use it unless necessary is that it breaks POSIX shell compatibility, ie. it only works with bash and bash derivatives, which I tend to not install on my systems.

cloudhead (approved):

PR #439, opened 2022-10-06
Sort imports

sebastinez:
Ok got myself into a sorting frenzy 😅 now I feel like it's maybe a bit much to review..
I would also be ok with leaving it to be sorted file by file

Sort imports by lexicalgraphic order.

1. Type imports
2. Third party packages
3. Internal imports

Also sorted props by alfabet

### DIFF
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -1,31 +1,34 @@
 <script lang="ts">
-  import { onMount } from "svelte";
-  import { link } from "svelte-routing";
-  import { ethers } from "ethers";
-  import {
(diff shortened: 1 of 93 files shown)


cloudhead:
My biggest worry is maintaining this sorted..

rudolfs:
> My biggest worry is maintaining this sorted.

I think it's fine to do this once. Whenever I refactor something I sort stuff out of habit anyway and now @sebastinez is also doing it. It improves readability and we don't have to enforce it for PRs, it can be a soft requirement.

PR #444, opened 2022-10-10
Remove deprecated functionality

rudolfs:
- get rid of the Graph queries -- we've run out of funds and the functionality is deprecated
- get rid of the Graph multi-sig write functionality -- orgs in the current form are deprecated and the Gnosis dependencies are holding us back from upgrading our stack
- get rid of remaining anchoring, org and member listing functionality
- get rid of deprecated and unused `multisig` and `multihashes` dependencies


src/config.ts
```diff
@@ -29,13 +23,11 @@ export class Config {
   network: { name: string; chainId: number };
   registrar: { address: string; domain: string };
   radToken: { address: string; faucet: string };
-  orgFactory: { address: string };
   reverseRegistrar: { address: string };
-  orgs: { subgraph: string; contractHash: string; pinned: string[] };
+  orgs: { contractHash: string };
```
cloudhead, line 27: contactHash isn't needed either afaict

src/utils.ts
```diff
@@ -425,21 +328,6 @@ export function getDaysPassed(from: Date, to: Date): number {
...
-  const hash = multihashes.decode(multihash);
-
-  return hash.digest;
-}
-
-// Create a project hash from a hash and format.
-export function formatProjectHash(multihash: Uint8Array): string {
-  const decoded = multihashes.decode(multihash);
-  return ethers.utils.hexlify(decoded.digest).replace(/^0x/, "");
-}
```
cloudhead, line 441: We're probably going to need all these methods again soon.

rudolfs: We should use packages that aren't deprecated when we bring them back then.

```
npm WARN deprecated multibase@4.0.6: This module has been superseded by the multiformats module
```

src/config.ts
rudolfs: It's still used by `utils.identifyAddress` which in turn is used by `Address.svelte`.

src/utils.ts
```diff
@@ -549,47 +410,12 @@ export async function resolveEnsProfile(
...
-    return null;
-  }
-
-  return {
-    address: query.safe.id,
-    owners: query.safe.owners,
-    threshold: query.safe.threshold,
-  };
-}
-
```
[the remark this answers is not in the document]
rudolfs: We'll bring it back when needed.

cloudhead (approved):

PR #447, opened 2022-10-11
Upgrade dependencies

rudolfs:
See individual commit descriptions for details.

```
$ npx npm-check -u
❤️  Your modules look amazing. Keep up the great work. ❤️
```


src/ColorPalette.svelte
```diff
@@ -164,6 +164,7 @@
 <svelte:window on:keydown={onKeydown} on:click={clickOutside} />
 
 {#if show}
+  <!-- svelte-ignore a11y-click-events-have-key-events -->
```
cloudhead, line 167: Can we do a global ignore for this? Seems like it's all over the place..

rudolfs: Yeah, it's tricky. I was looking for a good solution, but didn't find any. Because you have to disable it in your editor _and_ for `svelte-check` separately. We should really fix the a11y warnings instead of ignoring them, but that's a bit out of scope for this PR.

sebastinez (approved):

src/ColorPalette.svelte
rudolfs: Will do it in a follow-up: https://github.com/radicle-dev/radicle-interface/issues/449.

PR #452, opened 2022-10-13
Remove remaining global CSS styles

rudolfs:
Now we're left only with two global CSS classes `layout-` and `txt-`.


sebastinez (commented):
Hey have to still look at the CSS, which makes sense from what I can see from the code, but I would revisit the commit c51384e39fc97ccfe1b4dd3c0bfb8ce30fd81162.
I wouldn't remove the `use:link` since else we refresh the entire app when clicking on those links. But I'm in favor to deprecate the `Link` component and replace it with `a` tags

rudolfs:
> Hey have to still look at the CSS in the browser, from what I can see from the code it makes all sense, but I would revisit the commit [c51384e](https://github.com/radicle-dev/radicle-interface/commit/c51384e39fc97ccfe1b4dd3c0bfb8ce30fd81162). I wouldn't remove the `use:link` since else we refresh the entire app when clicking on those links. But I'm in favor to deprecate the `Link` component and replace it with `a` tags

Good point, didn't know about the refresh at the time I removed the `use:link` stuff. Will address.

rudolfs:
@sebastinez c02eaa8284c05f238d5c9e0c0684892712df60c2

cloudhead (approved):

PR #455, opened 2022-10-17
Upgrade dependencies

rudolfs:
This time we don't upgrade the transitive dependencies. Checked the production build locally, it works.

Seems like the `sive-parser` upgrade caused the production breakage. We should add a [smoke check](https://github.com/radicle-dev/radicle-interface/issues/456).


sebastinez (approved):
LGTM

PR #457, opened 2022-10-18
Disable color palette hotkey in production

rudolfs:
It's for devs only.


sebastinez (approved):
Awesome, yeah I think that works 👍

PR #460, opened 2022-10-20
Inline global gradient

rudolfs:
We only use it in one place anyway. And components should be self-contained, so they're easier to reason about.


sebastinez (approved):
Looks nice 👍

PR #461, opened 2022-10-20
Clean up config

rudolfs:
Motivation: working on replacing cypress with playwright and we need to be able to change the app configuration from within tests. At the moment the config is tightly coupled with wallet state.

- move constants that are tightly coupled to our code and unlikely to change or be user configurable to a central place in `constants.ts`
- decouple configuration from wallet state

Best reviewed commit-by-commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,9 +6,6 @@ simple guidelines.
 * Make sure you run `npx eslint .` to check for code formatting problems.
   You can use the `--fix` flag to automatically fix any issues the linter catches.
 * Before adding any code dependencies, check with the maintainers if this is okay.
-* When using configuration parameters, eg. API URLs, API tokens or contract addresses,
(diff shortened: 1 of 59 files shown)


rudolfs:
> TYPE 👏 SCRIPT 👏 IS 👏 NOT 👏 A 👏 CONFIGURATION 👏 LANGUAGE 👏

I'd argue that most of the stuff that was in `config.json` wasn't really configuration.

cloudhead (approved):
LGTM. You'll want to squash the changes though.

cloudhead:
> > TYPE clap SCRIPT clap IS clap NOT clap A clap CONFIGURATION clap LANGUAGE clap
> 
> I'd argue that most of the stuff that was in `config.json` wasn't really configuration.

Feel free to rename it to `<what-it-actually-is>.json` :wink: -- point is that it's data and doesn't require the full language to express, only a subset.

Issue #449, opened 2022-10-12
Fix all a11y issues


rudolfs:
```
src/views/projects/SourceBrowser/FileDiff.svelte
154:    <!-- svelte-ignore a11y-click-events-have-key-events -->

src/views/projects/Tree/Folder.svelte
67:<!-- svelte-ignore a11y-click-events-have-key-events -->

src/components/SquareButton.svelte
62:<!-- svelte-ignore a11y-click-events-have-key-events -->

src/components/Clipboard.svelte
55:<!-- svelte-ignore a11y-click-events-have-key-events -->

src/components/Button.svelte
116:<!-- svelte-ignore a11y-autofocus -->

src/App/ColorPaletteModal.svelte
134:  <!-- svelte-ignore a11y-click-events-have-key-events -->

src/components/Floating.svelte
45:  <!-- svelte-ignore a11y-click-events-have-key-events -->

src/components/Dropdown.svelte
52:    <!-- svelte-ignore a11y-click-events-have-key-events -->

src/components/Icon.svelte
31:<!-- svelte-ignore a11y-click-events-have-key-events -->

src/App/ModalPortal.svelte
32:    <!-- svelte-ignore a11y-click-events-have-key-events -->

src/App/Header/SettingsDropdown.svelte
94:  <!-- svelte-ignore a11y-click-events-have-key-events -->
107:        <!-- svelte-ignore a11y-click-events-have-key-events -->

src/App/Header/Connect.svelte
181:        <!-- svelte-ignore a11y-click-events-have-key-events -->
199:          <!-- svelte-ignore a11y-click-events-have-key-events -->
216:        <!-- svelte-ignore a11y-click-events-have-key-events -->
```

sebastinez:
PR #471 removes a11y issues from `src/base/projects/Header.svelte`

rudolfs:
We haven't prioritised this in a while. Will open a new one when users start complaining.

PR #463, opened 2022-10-24
Add custom router v2

sebastinez:
Removes `svelte-routing` as the chosen router and adds a custom router, inspired by the [radicle-upstream router](https://github.com/radicle-dev/radicle-upstream/tree/main/ui/src/router).

**Pending:**
- [x] Clean up commit history
- [x] Some refactoring necessary to make the code terser.
- [x] Error handling review
- [x] Add more tests, in particular around hash and querystring parsing. _Moved to another PR_
- [x] Will add more comments at maybe hard to understand places.
- [ ] I think we can add hash routing as an option with some minor changes _Removed from this PR_
- [ ] Will look into this too `would be cool if we could do navigate('/foo', ModuleName, { preload: true }) or something` _Moved to another PR_


src/router/definitions.ts
```diff
@@ -0,0 +1,54 @@
...
+  patch?: string | null;
+  path?: string | null;
+  peer?: string | null;
+  profile?: string | null;
+  revision?: string | null;
+  route?: string | null;
+  search?: string | null;
+  seed?: string | null;
+  urn: string;
+}
```
cloudhead, line 52: Why are we allowing both `null` and `undefined` in all of these?

src/router/definitions.ts
```diff
@@ -0,0 +1,54 @@
...
+              params: { nameOrDomain: string; owner: string | null };
+            }
+          | {
+              type: "view";
+              params: { nameOrDomain: string; retry: boolean };
+            };
+      };
+    }
+  | { type: "seeds"; params: { host: string } }
+  | { type: "vesting" };
```
cloudhead, line 37: I'd rename `type` to `resource`, or `name`, since these routes are all of the same type.
I'm also wondering why `activeView` is not called `view`? Is there an `inactiveView` as well?

src/router/index.ts
```diff
@@ -0,0 +1,422 @@
...
+          return {
+            type: "registration",
+            params: {
+              activeView: {
+                type: activeView,
+                params: { nameOrDomain, owner },
+              },
+            },
+          };
+        } else if (activeView === undefined) {
```
cloudhead, line 138:
```suggestion
        } else if (!activeView) {
```

src/router/index.ts
```diff
@@ -0,0 +1,422 @@
...
+              type: "projects",
+              params: {
+                ...params,
+                search: url.search,
+                seed: host,
+                urn,
+              },
+            };
+          }
+          return { type: "404" };
```
cloudhead, line 201: Can we make `404` be the default, instead of returning it everywhere?

src/router/index.ts
```diff
@@ -0,0 +1,422 @@
...
+  ) {
+    if (route.params.activeView.params.owner) {
+      return `/registration/${route.params.activeView.params.nameOrDomain}/${route.params.activeView.type}?owner=${route.params.activeView.params.owner}`;
+    }
+    return `/registration/${route.params.activeView.params.nameOrDomain}/${route.params.activeView.type}`;
+  } else if (route.type === "profile") {
+    return `/${route.params.addressOrName}`;
+  } else if (route.type === "404") {
+    return null;
+  }
```
cloudhead, line 341: We should be using an exhaustive check here. This `if` statement will still compile if we forgot to handle a route..

src/router/index.ts
```diff
@@ -0,0 +1,422 @@
...
+  const { pathname, search, hash } = window.location;
+  replace(pathToRoute(pathname + search + hash));
+};
+
+export function pathToRoute(path: string | null): Route {
+  if (!path) {
+    return { type: "404" };
+  }
+
+  const url = new URL(path, "https://app.radicle.xyz");
```
cloudhead, line 112: Does this work in local development?

src/App.svelte
```diff
@@ -82,21 +92,65 @@
...
+          seed={$activeRouteStore.params.seed || null}
+          peer={$activeRouteStore.params.peer || null}
+          urn={$activeRouteStore.params.urn} />
+      {:else if $activeRouteStore.type === "profile"}
+        <Profile
+          addressOrName={$activeRouteStore.params.addressOrName}
+          {wallet} />
+      {:else}
+        <NotFound title="404" subtitle="Nothing here" />
+      {/if}
```
cloudhead, line 153: Can we create a `Route` component? This is pretty unreadable..

src/Header.svelte
```diff
@@ -170,7 +170,7 @@
     {:else}
       <span class="network unavailable">No Network</span>
     {/if}
-    <a use:link class="register" href="/registrations">Register</a>
+    <a use:link class="register" href="/registration">Register</a>
```
cloudhead, line 173: Why is this being changed? It's inconsistent with the other resources that are list-like, and it's a breaking change.

src/base/projects/Browser.svelte
```diff
@@ -79,9 +80,15 @@
...
-    }
+    navigate({
+      type: "projects",
+      params: {
+        urn: project.urn,
+        path: newPath,
+        route: null,
+        revision,
+      },
+    });
```
cloudhead, line 91: This function should just take the route name and params: `navigate("projects", { urn: project.urn, ... })`

src/base/registration/CheckNameAvailability.svelte
```diff
@@ -1,6 +1,6 @@
 <script lang="ts">
```
cloudhead, line 1: We name views that let you create a new resource `New`.

src/base/projects/View.svelte
```diff
@@ -37,99 +88,83 @@
...
+        {#await issue.Issue.getIssues(project.urn, project.seed.api)}
+          <Loading center />
+        {:then issues}
+          <Issues {project} state={issueFilter} {wallet} {issues} />
+        {:catch e}
+          <div class="message">
+            <Message error>{e.message}</Message>
+          </div>
+        {/await}
+      {:else if content === "issue" && $activeRoute.params.issue}
```
cloudhead, line 135: Why are there singlular versions of certain resources now?

sebastinez: We started using singular versions of certain resources, to add type safety to the different route types and limit certain route params to `issues` and others to the `issue` route.
Else we would have the `issue` param in one case as `null` and in another as a `string`

src/base/projects/Browser.svelte
sebastinez: The interesting part here is to be able to pass a `Route` type directly into `navigate` for type safety at compile time.
If we separate this into two params we would loose it.

src/router/index.ts
sebastinez: Yeah this works in a local dev environment, since we don't use the base url, but eventually we could use `window.origin` here. 🤔

src/router/definitions.ts
sebastinez: I ended up with both `null` and `undefined`, so we would be able to define some props as `null` in the `pathToRoute` fn but at the same time we could only pass the params we wanted to define, instead of declaring every route param.
Similar to what we had before with the `BrowseTo` interface https://github.com/radicle-dev/radicle-interface/blob/87c1d0f6e2a06b2f6572a1fdcd546a1ba2d47aea/src/project.ts#L121

src/router/index.ts
sebastinez: fixed in 9fe92d1

src/base/registration/CheckNameAvailability.svelte
sebastinez: Reverted in f26098eaa8f4187e99fd28e7be8db02f990d0d5c

src/router/definitions.ts
rudolfs: I think we started using `type` in Upstream, because the route itself is a type union. But yeah, the attribute name also could be something else. My first choice would still be `type`, second `resource`. As for `activeView`, maybe more appropriate would be `subView` or `subRoute`.

src/App.svelte
rudolfs: I'd rather not add another layer of indirection, if possible, because it makes it harder to work with in my opinion. We could however push some of the screens that have sub-routing back down into a component as a compromise to do away with the sub-branching.

src/Header.svelte
rudolfs: This was my doing. My thinking behind it was that most of our routes are singular and we only have one `registration` screen. The other screens being:
```
/faucet
/vesting
```
So I wanted to align that. This also would better align with routes that actually distinguish between singular and plural, like `commit` vs `commits` which are two different sub-routes/screens of the project screen.

src/App.svelte
sebastinez: >  push some of the screens that have sub-routing back down into a component

Will provide a commit for that, so we can see if it makes it more readable

src/base/registration/CheckNameAvailability.svelte
rudolfs: This was also my doing. I didn't know about this convention. My thinking behind it was, that the whole registration flow is broken up in multiple steps and I find it more descriptive if we'd name the steps for what they are supposed to accomplish. So:

`ValidateName.svelte` -> `CheckNameAvailability.svelte` -> `Register.svelte`

And then a separate view for viewing the registration: `View.svelte`.

This is also aligned with our routes, making it (at least from my perspective) easier to reason about.

```typescript
    {
      type: "registration";
      params: {
        activeView:
          | {
              type: "validateName";
            }
          | {
              type: "checkNameAvailability";
              params: { nameOrDomain: string; owner: string | null };
            }
          | {
              type: "register";
              params: { nameOrDomain: string; owner: string | null };
            }
          | {
              type: "view";
              params: { nameOrDomain: string; retry: boolean };
            };
      };
    }
```

src/App.svelte
sebastinez: 7ebc71d6ac21605c09821e3661f4f571b4d040a1

src/router/definitions.ts
sebastinez: fixed by dfe546f49cda95e1dde641a87715950245932de3

src/router/definitions.ts
sebastinez: fixed by 39a604e75653067708bf34b5b1c51f0b2abd8d8d, 72690b7f2496b248c236bfcf7c42cb611942a423 and finally 76c9146346648573fec1ee5584157c85485b9429

src/router/index.ts
sebastinez: Done in 5f3963bd91ebbc552e216dd713136dafa383d1e6

src/Header.svelte
cloudhead: the plural routes are the ones that represent more than one sub-resources, eg. orgs, projects, seeds, registrations

src/base/registration/CheckNameAvailability.svelte
cloudhead: I think there should be a way to have both of those. Still I wouldn't want to have to look through all the components to figure out which one is the entry point for new registrations.

src/base/projects/View.svelte
cloudhead: Ah, so this doesn't affect the URL?

src/base/registration/CheckNameAvailability.svelte
rudolfs: Yeah, I think you're right about the entry point, that's something I'd also want to see from just looking at the code.

src/base/projects/View.svelte
sebastinez: Currently it does affect the URL since we match the resource name used in the `Route` type with the URL.
But I think we could to maintain the same resource between a listing and a detail view by tweaking the parsing functions a bit

src/Header.svelte
sebastinez: fixed in 1232ca3b3595029730d172920de6c48057927a61

src/router/index.ts
sebastinez: dc69cdae858cfdff0c53f7b50ad0b16edeb67afe fixes this, we return `null` as indicator that no route has been found, and make sure in the `initialize` function to return the `404` route if no route has been found

src/router/index.ts
sebastinez: dc69cdae858cfdff0c53f7b50ad0b16edeb67afe moves the `return null` into an else branch, so we handle the case of the forgotten route.

sebastinez: b2df958c17e03ad346900a8ca6491aefd3ef3e58 ok this one is even better as exhaustive check

sebastinez: 00a4bf563db83d86416e145453c79b625f3987e1 and more exhaustivness 😅

src/base/projects/View.svelte
sebastinez: Ok pushed be338c674db2bdc44303ad223c2d47de147e117e which reverts the project resources back to the current state, with minimal changes to the `pathToRoute` and `routeToPath` functions

src/base/registration/CheckNameAvailability.svelte
sebastinez: Ok this commit reverts the renaming of the registration components 0669ddb43f1aa26dadb02faaa248f52a048908a6

src/Header.svelte
```diff
@@ -162,15 +163,30 @@
 
   <div class="right">
     {#if wallet && wallet.network.name === "goerli"}
-      <a use:link href="/faucet">
+      <!-- svelte-ignore a11y-invalid-attribute -->
```
cloudhead, line 166: Can we just disable this lint? It's creating a lot of noise.

src/App.svelte
```diff
@@ -82,21 +86,33 @@
...
+        <Projects {wallet} activeRoute={$activeRouteStore} />
+      {:else if $activeRouteStore.resource === "profile"}
+        <Profile
+          addressOrName={$activeRouteStore.params.addressOrName}
+          {wallet} />
+      {:else if $activeRouteStore.resource === "404"}
+        <NotFound title="404" subtitle="Nothing here" />
+      {:else}
+        {unreachable($activeRouteStore)}
+      {/if}
```
cloudhead, line 115: Could we use a `switch` here to reduce the noise?

src/profile.ts
```diff
@@ -1,4 +1,4 @@
-import type { EnsProfile } from "@app/base/registrations/registrar";
+import type { EnsProfile } from "@app/base/registration/registrar";
```
cloudhead, line 1: :eyes:

src/App.svelte
sebastinez: It seems that a `switch` statement is eventually coming for svelte
https://github.com/sveltejs/svelte/issues/530

But maybe for now we can keep the `ifs` to avoid defining the component in a separate function, and the need to pass around the `wallet` props

src/Header.svelte
rudolfs: I'm not aware of an easy way to [disable these globally](https://github.com/radicle-dev/radicle-interface/issues/449). A potential workaround is to use a different html tag, or extract it into an `<InternalLink>` component.

sebastinez: Ok we have removed all the `invalid-attribute` tags and there is a PR on the way to remove some more wrt to the missing click events #471

cloudhead (approved):
LGTM! I haven't tested it though, so hopefully someone made sure everything works :)

rudolfs:
> LGTM! I haven't tested it though, so hopefully someone made sure everything works :)

I tested it quite a bit. I think it's tight.

rudolfs (approved):
This is awesome @sebastinez, a really nice improvement!! 🎉

sebastinez:
> This is awesome @sebastinez, a really nice improvement!! 🎉

Thanks rudolfs, wouldn't be where it is without your refactor and support

PR #466, opened 2022-10-26
Migrate from Cypress to Playwright

rudolfs:
Playwright is more flexible and has better DX than Cypress.

### Overview

I tried out a couple of ideas and ended up going with a local `http-api` Docker container serving known seed data. Initially I wanted to script all the interactions and build the seed for every test on the fly, but that seemed to be overkill for the current feature-set and turned out to be quite slow. We use Docker to be able to run tests on both Linux and macOS easily as the `radicle-client-services` repository publishes builds for each commit. This also lets us test different versions of backends with the front end.

I created a known seed fixture, which is just an archive of `RAD_HOME` of a seed that contains one project that two peers have interacted with. The fixture is archived to preserve the folder structure exactly, since otherwise git doesn't store empty folders. The project that the fixture is based on contains all kinds of edge cases that test all the source browsing features of the app. When we start the docker container the seed fixture is unpacked and its contents are served over the http-api against which we run the tests.

The script with which I created the fixture as well as the test project repository are part of this PR for later amendments.

I'm on the fence about component tests. For now I didn't try to recreate them in Playwright, because I don't believe they add much value and take time to write and maintain. I tried to cover as much functionality as I could with e2e tests.

### Running e2e tests locally

To run the tests locally, you'll need to install Docker first.

Start the http-api container:
```
$ ./scripts/http-api-test
```

or, if you dislike Docker and have the `radicle-http-api` binary in your `PATH`, you can run it like this:
```
$ ./scripts/http-api-test --binary
```

Then run the tests like this:
```
$ npm run test:e2e -- --project chromium
```
To run tests with visual debugging:
```
$ npm run test:e2e -- --project chromium --debug
```

### Test artifacts

We collect and upload test artifacts whenever a test fails. The zip file contains a combined log file for playwright and browser console logs as well as a Playwright trace.

Screenshot 2022-11-28 at 17 57 07

```
$ tree test-artifacts-Linux
.
└── landingPage-shows-pinned-projects-chromium
    ├── test.log
    └── trace.zip
```

You can replay the trace file locally like this:
```
$ npx playwright show-trace test-artifacts-Linux/landingPage-shows-pinned-projects-chromium/trace.zip
```

Screenshot 2022-11-28 at 18 00 03

### Logging

We add colorized logging for easy debugging. For now we log any requests that the app makes to hosts other than localhost, because these should be either mocked out or disabled in the app during testing. We colorize the logs depending on which source they come from, so far we got `playwright` - originating from the test runner and `console` coming from the app browser console. Later on we could use the same mechanism for logging individual peers interacting with each other.

Screenshot 2022-11-29 at 10 31 31

On CI logs are only collected in a file and uploaded as an artifact if the test fails. We do this to keep the CI log output readable.

Screenshot 2022-11-29 at 10 31 06

### Out of scope

Potential follow-ups:
  - Ethereum feature testing
    - refactor wallet/session so that Ethereum features can be disabled via config
  - visual snapshot tests
    - keep heavy assets in a different history so that the repo can be pruned later
  - test multiple seeds simultaneously
    - search relies on this
    - we need to refactor the app to allow adding multiple local seeds on different ports


tests/e2e/project/commit.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  await expect(page).toHaveURL(modifiedFileFixture);
+});
+
+test("modified file", async ({ page }) => {
+  await page.goto(modifiedFileFixture);
+
+  // Commit header.
+  {
+    const header = page.locator(".commit header");
+    await expect(header.locator("text=bob committed 7 days ago")).toBeVisible();
```
sebastinez, line 23: This seems to have a difference in the `X days ago` while we expect 7 we show 8 days.
Bildschirm­foto 2022-11-29 um 10 58 26

Not sure yet why.

rudolfs: I think I forgot to add time mocking when I moved the tests to a separate file. I wonder why it didn't fail on CI though. 🤔 

Will fix that.

sebastinez:
Wanted to add to any reviewers with `npx playwright test --quiet` you can reduce the verbosity of the testing

tests/e2e/project/commit.spec.ts
rudolfs: Oh, I know why it worked. Because I pushed it yesterday. 🤣

rudolfs: d7df79dca4db0cef4de830563366373a32a01c32

sebastinez: d7df79d fixed this

tests/e2e/project.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+  await sourceTree.getByText("git/").click();
+  await sourceTree.getByText("repositories/").click();
+  await sourceTree.getByText(".gitkeep").click();
+  await expect(
+    page.locator("text=0801ace Add a deeply nested directory tree"),
+  ).toBeVisible();
+
+  // After a page reload the tree browser is still expanded and we're still
+  // showing the .gitkeep file.
+  {
```
sebastinez, line 104: Out of curiosity, I would like to know why we need to change the scope of the `page.reload() and the verification of the reloaded tree`. Couldn't we keep it in the same scope as the initial test?

tests/e2e/project.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+  await sourceTree.getByText(";semicolon;").click();
+  await expect(page.locator(".file-name")).toContainText(";semicolon;");
+
+  await sourceTree.getByText("@at@").click();
+  await expect(page.locator(".file-name")).toContainText("@at@");
+
+  await sourceTree.getByText("_underscore_").click();
+  await expect(page.locator(".file-name")).toContainText("_underscore_");
+
+  // TODO: fix the error in the app for these edge cases.
```
sebastinez, line 150: I would probably change this to `fix the error in the http-api` since the app doesn't have any issues with these values

tests/e2e/project.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+    page.locator(
+      "text=rad clone rad://0.0.0.0/hnrkgd7sjt79k4j59ddh11ooxg18rk7soej8o",
+    ),
+  ).toBeVisible();
+  await expect(
+    page.locator(
+      "text=https://0.0.0.0/hnrkgd7sjt79k4j59ddh11ooxg18rk7soej8o.git",
+    ),
+  ).toBeVisible();
+});
```
sebastinez, line 306: Maybe we can also test here the clipboard functionality?

sebastinez:
I think we should add more tests on the router, specially the hash routing mode

- check that the hash is being positioned at the correct place e.g. `https://ipfs.io/ipfs/#/...`
- the forward and backward navigation in hash and history mode (go forward and backward and check that it still checks out)
- check that the querystring and hashes are correctly created where used and persisted where needed.

playwright.config.ts
```diff
@@ -0,0 +1,38 @@
...
+  globalSetup: "./tests/support/globalSetup",
+  use: {
+    actionTimeout: 0,
+    baseURL: "http://localhost:3000",
+    trace: "retain-on-failure",
+  },
+
+  projects: [
+    {
+      name: "chromium",
```
sebastinez, line 25: It would be nice to add more browser testing, like firefox and safari, specially safari has some things that would be nice to iron out and make it the browser experience the same in all browsers

Something like this should work

```diff
diff --git a/playwright.config.ts b/playwright.config.ts
index 2ca2f157..bfc17586 100644
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -27,6 +27,18 @@ const config: PlaywrightTestConfig = {
         ...devices["Desktop Chrome"],
       },
     },
+    {
+      name: "firefox",
+      use: {
+        ...devices["Desktop Firefox"],
+      },
+    },
+    {
+      name: "webkit",
+      use: {
+        ...devices["Desktop Safari"],
+      },
+    },
   ],

   webServer: {
```

tests/e2e/project.spec.ts
rudolfs: I added scopes is only to make larger tests more readable by dividing them up in visual chunks.

tests/e2e/project.spec.ts
rudolfs: I also thought of testing clipboard functionality, but then decided against it, because there is no isolation for clipboard stuff and since tests run in parallel, there might be race conditions. I'll think on it. Maybe we can add all clipboard tests in one test to work around this issue.

tests/e2e/project.spec.ts
sebastinez: nice, understood and I agree

package.json
```diff
@@ -9,29 +9,33 @@
     "postinstall": "scripts/copy-katex-assets && scripts/install-twemoji-assets",
     "check": "scripts/check",
     "format": "npx prettier '**/*.@(ts|js|svelte|json|css|html)' --ignore-path .gitignore --write",
-    "test:unit": "TZ='UTC' vitest run",
-    "test:components": "cypress run --component",
-    "test:e2e": "cypress run",
```
sebastinez, line 14: Could we leave a
```
"test:e2e": "playwright test --quiet"
```
Seems like something I would do frequently.

rudolfs: What I actually would like is to reduce these logs by refactoring the app. So that we can disable some parts that aren't being tested. For example, we should be able to disable Ethereum features if they're not being tested, which would probably remove all the logs.

The reason I'd like to leave them "on" in dev mode is that we should notice when the app is trying to access resources that aren't stubbed or if there are console logs/warnings/errors, so that we address them right away.

sebastinez: Ah okay got it 👍 
I would still prefer to keep some npm script to execute quickly the e2e test suite. Maybe can do a `"test:e2e": "playwright test"`

tests/e2e/project.spec.ts
rudolfs: f9e5de89c26cc8fc2c60caf2a795f75d14971d64

tests/e2e/project.spec.ts
rudolfs: 695fcdeaae09d24622e138096ba2913bf48140f3

playwright.config.ts
rudolfs: I think we might want to do that only on CI, otherwise it takes too long when run locally. f0850694134845180f8b3a0750bb04020cbc9fcc

Another idea would be to run 3 parallel workflows on CI, one for each browser.

TBD: safari tests, they're all broken for some reason.

package.json
rudolfs: a8efbd7869ec9a96ed35cd1ee268757eda1f205c

tests/e2e/seed.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { test, expect } from "@tests/support/fixtures.js";
+
+test("seed metadata", async ({ page }) => {
+  await page.goto("/seeds/radicle.local");
+
+  await expect(page.locator("header").getByText("radicle.local")).toBeVisible();
+  await expect(page.getByRole("link", { name: "radicle.local" })).toBeVisible();
+  await expect(page.locator("text=hyb6i8…sg6nn4")).toBeVisible();
```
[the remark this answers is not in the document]
rudolfs: a14a6e5e285bafe4420485a09c588cd1cdfde5ed

playwright.config.ts
rudolfs: I'll keep this open and play around with running this in parallel and fix the webkit stuff.

scripts/run-http-api-with-fixtures
```diff
@@ -0,0 +1,132 @@
...
+    echo "Couldn't find the $HTTP_API_BINARY binary in your PATH."
+    echo "You can compile it from source:"
+    echo "  👉 https://github.com/radicle-dev/radicle-client-services/tree/${REV}/http-api"
+    echo
+    exit 1
+  fi
+
+  echo
+  echo "Starting $HTTP_API_BINARY"
+  echo "  $HTTP_API_BINARY -listen 0.0.0.0:8777 --root ${WORKSPACE} --passphrase $PASSPHRASE"
```
cloudhead, line 85:
```suggestion
  echo "  $HTTP_API_BINARY --listen 0.0.0.0:8777 --root ${WORKSPACE} --passphrase $PASSPHRASE"
```

cloudhead (approved):
LGTM, nice work. Make sure you squash to the minimal set of commits :+1:

cloudhead (approved):

playwright.config.ts
rudolfs: Okay, will leave the webkit stuff to a follow-up. For now we run the chromium and firefox tests in parallel on CI.
Locally we can specify which browser we would like to target like this:

```
npm run test:e2e -- --project chromium
```

rudolfs: f0d7f16d13946b0b3e570f27325560ab1fd32068

sebastinez (approved):
Yeah LGTM, would only remove the `cypress/screenshots` folder from `.gitignore`

.gitignore
```diff
@@ -13,6 +13,8 @@ public/twemoji/*.svg
 
 # Integration Tests
 cypress/screenshots
```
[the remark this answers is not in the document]
rudolfs: fcd3e676de73dcf32e0c90cd47aade3f48a47572

PR #467, opened 2022-10-27
Review and improve router rewrite

rudolfs:
PR review in patch form :)


src/router/index.ts
```diff
@@ -141,12 +141,14 @@ export function pathToRoute(path: string | null): Route {
         return {
           type: "faucet",
           params: {
-            activeView: "withdraw",
-            amount: url.searchParams.get("amount"),
+            activeView: {
+              type: "withdraw",
+              params: { amount: url.searchParams.get("amount") ?? "0" },
```
sebastinez, line 146: I understand this change, but I'd prefer to parse it to `null` and handle it in the `FaucetWithdraw` component.
Since a missing search query means that nothing was passed (not even 0)

rudolfs: Before you weren't handling it at all thought:
```
{#if activeView === "form"}
  <Index {wallet} />
{:else if activeView === "withdraw" && amount}
  <Withdraw {wallet} {amount} />
{/if}
```

You didn't render anything.

Screenshot 2022-10-27 at 15 20 57

rudolfs: If we hit the withdraw endpoint manually _with_ an amount, it also doesn't work, because it relies on the form step to do some set-up beforehand. So I think it's out of scope to fix this for the routing PR.

http://127.0.0.1:3000/faucet/withdraw?amount=10

Screenshot 2022-10-27 at 15 26 27

rudolfs: Did it like this: 2f6173fd367783d38dd8d12568cc9ac36a76abe7.

src/NotFound.svelte
```diff
@@ -1,11 +1,11 @@
 <script lang="ts">
+  import { pop } from "@app/router";
```
sebastinez, line 2: Can I suggest to clarify to future devs here maybe to make it clearer use

```suggestion
  import router from "@app/router";
  
  <Button variant="foreground" on:click={router.back}>Back</Button>
```

PR #469, opened 2022-10-28
Router review and improvements

rudolfs:
- [x] get rid of untyped `navigate` function
- [x] encode all project routes as proper types

### DIFF
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { onMount } from "svelte";
-  import * as router from "@app/router";
+  import { link } from "@app/router";
   import { ethers } from "ethers";
   import {
     explorerLink,
@@ -97,7 +97,7 @@
     {#if addressType === AddressType.Org}
       <!-- svelte-ignore a11y-invalid-attribute -->
       <a
-        use:router.link={{
+        use:link={{
           type: "profile",
           params: { addressOrName: nameOrAddress },
         }}
@@ -117,7 +117,7 @@
     {:else if addressType === AddressType.EOA}
       <!-- svelte-ignore a11y-invalid-attribute -->
       <a
-        use:router.link={{
+        use:link={{
           type: "profile",
           params: { addressOrName: nameOrAddress },
         }}
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -28,7 +28,7 @@
 
   let addressType: AddressType | null = null;
 
-  const nameOrAddress = profile?.ens?.name || address;
+  const addressOrName = profile?.ens?.name || address;
 
   onMount(async () => {
     if (!profile) {
@@ -99,7 +99,7 @@
       <a
         use:link={{
           type: "profile",
-          params: { addressOrName: nameOrAddress },
+          params: { addressOrName: addressOrName },
         }}
         href="">
         {addressLabel}
@@ -119,7 +119,7 @@
       <a
         use:link={{
           type: "profile",
-          params: { addressOrName: nameOrAddress },
+          params: { addressOrName: addressOrName },
         }}
         href="">
         {addressLabel}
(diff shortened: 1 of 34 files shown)


src/base/registration/CheckNameAvailability.svelte
```diff
@@ -57,6 +57,13 @@
       error = err.message;
     }
   });
+
+  function goBack() {
```
sebastinez, line 61: Sorry for commenting on your draft PR, but maybe we should find a uniform naming for going back in history.

Maybe we can rename `router.pop` into `router.back` since the router is essentially going back. Not necessary to let users know their popping items from the history stack. wdyt?

src/Address.svelte
```diff
@@ -95,7 +95,7 @@
   {/if}
   <div class="wrapper">
     {#if addressType === AddressType.Org}
-      <a use:link href={`/${nameOrAddress}`}>{addressLabel}</a>
+      <a use:router.link href={`/${nameOrAddress}`}>{addressLabel}</a>
```
sebastinez: This seems weird, I would prefer to use the exported function as previously.. 😕

src/base/projects/View.svelte
```diff
@@ -24,42 +24,45 @@
...
 
     return project;
   };
 
   // Content can be altered in child components.
-  $: revision = $activeRoute.params.revision || null;
-  $: content = $activeRoute.params.content || "tree";
+  $: revision = activeRoute.params.revision || null;
+
+  $: console.log(activeRoute.params.activeView);
```
sebastinez, line 65: This can be eventually removed :eyes: 😅

src/Address.svelte
```diff
@@ -95,7 +95,15 @@
   {/if}
   <div class="wrapper">
     {#if addressType === AddressType.Org}
-      <a use:link href={`/${nameOrAddress}`}>{addressLabel}</a>
+      <!-- svelte-ignore a11y-invalid-attribute -->
+      <a
+        use:router.link={{
```
sebastinez, line 100: This way of defining the use action seems weird.. I would prefer to keep it as `use:link`

sebastinez (commented):
Hey nice work! Left two small comments.
Also tried some things with the `ProjectsParams` to avoid `undefined` and `null` but need more time

src/base/projects/View.svelte
rudolfs: 62b004222c52666f9e13b3c346bda5d8fa980cf3

src/Address.svelte
rudolfs: I kind of want to get rid of the `link` action eventually, it's very awkward to use together with anchor tags, because Svelte doesn't like empty `href`s and the browser styling is off if we omit the `href`. But maybe that's a follow-up thing. Will restore the imports for now.

src/base/projects/ProjectMeta.svelte
```diff
@@ -88,19 +71,32 @@
...
       <span class="divider">/</span>
     {/if}
     <span class="truncate">
-      <a use:link={rootPath} class="project-name" href={routeToPath(rootPath)}>
+      <!-- svelte-ignore a11y-invalid-attribute -->
+      <a
+        href=""
+        use:router.projectLink={{
+          path: "/",
+          peer: null,
```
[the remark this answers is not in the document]
rudolfs: Goog catch!

src/Address.svelte
rudolfs: 3780790d1c8d9e807970c93365287db004ac422c

src/base/registration/CheckNameAvailability.svelte
sebastinez: > we're the users of the router API

True that, ok lets leave it

rudolfs:
> Also tried some things with the ProjectsParams to avoid undefined and null but need more time

I just removed the `null`s, I find using `undefined` more ergonomic in ts codebases. 🤷

PR #471, opened 2022-11-04
Allow switch between rendered markdown and raw content

### LINKED ISSUE #423: Add the ability to switch between rendered markdown blob and source blob
Currently all markdown files are being rendered by default, I think it would be nice to see also the source blob without any rendering if one chooses to.

Not sure yet where this option should be able to be toggled.

_To be continued_

sebastinez:
This PR adds a toggle button on the top right of the file preview shown in the Browser.
This allows to see the markdown file in its raw format and also reference a specific line of the file by URL.

It also refactors the project header stats into their own component `Stat.svelte`

Bildschirm­foto 2022-11-04 um 13 09 05

Bildschirm­foto 2022-11-04 um 14 08 30

Bildschirm­foto 2022-11-04 um 14 10 15

Bildschirm­foto 2022-11-04 um 14 10 48

Closes #423

### DIFF
--- a/src/base/projects/Header.svelte
+++ b/src/base/projects/Header.svelte
@@ -8,6 +8,7 @@
   import CloneButton from "@app/base/projects/CloneButton.svelte";
   import PeerSelector from "@app/base/projects/PeerSelector.svelte";
   import { closeFocused } from "@app/Floating.svelte";
+  import Stat from "@app/base/projects/Stat.svelte";
 
   export let activeRoute: ProjectRoute;
   export let project: Project;
@@ -75,12 +76,6 @@
     border-radius: var(--border-radius-small);
     min-width: max-content;
   }
-  .clickable {
-    cursor: pointer;
-  }
-  .clickable:hover {
-    background-color: var(--color-foreground-2);
-  }
   .not-allowed {
     cursor: not-allowed;
   }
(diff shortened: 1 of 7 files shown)


src/base/projects/Stat.svelte
```diff
@@ -0,0 +1,49 @@
...
+    background-color: var(--color-foreground-2);
+  }
+  .not-allowed {
+    cursor: not-allowed;
+    color: var(--color-foreground-5);
+  }
+</style>
+
+<button
+  {id}
```
rudolfs, line 40: I'd [avoid using `id`s](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) as selectors for cypress.

I'd much rather we use [`aria-label`s](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) for that, this is also the preferred way in Playwright.

src/base/projects/Stat.svelte
```diff
@@ -0,0 +1,49 @@
...
+  {title}
+  {disabled}
+  class:active
+  class:clickable
+  class:not-allowed={disabled}
+  class="stat clickable"
```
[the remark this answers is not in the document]
sebastinez: nope, good catch, we should use it as `class:clickable`

src/base/projects/Stat.svelte
sebastinez: AFAIK we don't want to use `data-*` labels, but I think `aria-label` should be ok. Will change that

sebastinez: e7b75d8c43da5a4e8f908def84108938d9b1eb68

src/base/projects/Stat.svelte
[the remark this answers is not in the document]
sebastinez: 291409451751eee3df74d5ec443f2a03e917eca5

src/base/projects/Header.svelte
```diff
@@ -128,49 +102,47 @@
...
+      active={activeRoute.params.view.resource === "patches"}
+      disabled={project.patches === 0}
       on:click={() => toggleContent("patches", false)}>
       <span class="txt-bold">{project.patches}</span>
       patch(es)
-    </div>
+    </Stat>
   {/if}
-  <div class="stat contributor-count widget">
+  <Stat ariaLabel="contributor-count">
```
rudolfs, line 144: Could we spell out them without the hyphen? E.g. "Contributor count".

src/base/projects/Stat.svelte
```diff
@@ -0,0 +1,49 @@
...
+  export let clickable = false;
+  export let disabled = false;
+</script>
+
+<style>
+  .stat {
+    font-family: var(--font-family-monospace);
+    font-size: var(--font-size-tiny);
+    padding: 0.5rem 0.75rem;
+    padding-bottom: 1rem; /* moving the content a tad higher to match the previous span usage */
```
rudolfs, line 14: This seems a bit odd to me. Is there something off with the line-height?

sebastinez:
waiting for #463 to be merged

rudolfs (commented):
Wonder if we could call the component something like `<HeaderToggleButton />`?

src/base/projects/Stat.svelte
sebastinez: Yeah certainly, I tried to find the difference, but for some magic css reason the text content of the `span` was a few pixels higher than the one from `button`

Bildschirm­foto 2022-11-04 um 16 00 10

If you find the reason why, please feel free to suggest any changes

src/base/projects/Header.svelte
sebastinez: I liked to pass a combined word like `contributor-count` to the test suite since we use it mostly for testing.
But I think it we should be able to spell it out

sebastinez:
> Wonder if we could call the component something like `<HeaderToggleButton />`?

Yeah we certainly could discuss a different name, not sure if `HeaderToggleButton` since not all buttons are toggeable

src/base/projects/Stat.svelte
rudolfs: We'll never know 🤷

cloudhead:
Looks good, I'd maybe just change the label to "Raw", as button actions are usually capitalized

sebastinez:
> Looks good, I'd maybe just change the label to "Raw", as button actions are usually capitalized

Fixed in 2b387b5449e6d116180453cae976c91f6c5e9289

src/base/projects/Header.svelte
rudolfs: I wouldn't have anything against hyphenating it if it wasn't an accessibility attribute which might be read by screen readers to text-to-speech for people who need it.

src/base/projects/Readme.svelte
```diff
@@ -14,9 +14,8 @@
 <style>
   article {
     padding: 2rem;
-    max-width: 64rem;
```
rudolfs, line 17: I thought this was a deliberate change by @cloudhead: https://github.com/radicle-dev/radicle-interface/commit/771ed2aea7b74e7f1108224a40d43e5fed967aa9.

What's the reason for the change?

I think it was so that the RIPs look better:
http://127.0.0.1:3000/seeds/clients.radicle.xyz/rad:git:hnrkx7ruakgb9ozzi7atwb1rh6moq8pjonp6o/tree/8cd1561688462e2fcb234956b6f5916c42bdf4fa/0001-heartwood.md

Your suggested change:
Screenshot 2022-11-07 at 13 31 35

vs production

Screenshot 2022-11-07 at 13 31 44

src/base/projects/Header.svelte
sebastinez: e7c6607e

src/base/projects/Readme.svelte
sebastinez: Good catch didn't know that, the reason was that due to using the `file-header` now from the `Blob` component that uses `width: 100%` we had a header that went until the right side and the markdown stopped at `64rem` will provide a better solution

sebastinez: 217c87628e6eb9de1351c31e2477ad84e64a79a2

rudolfs (approved):
🚀

PR #472, opened 2022-11-07
Avoid the re-rendering of the ProjectMeta component

sebastinez:
By passing a too broad scope into the `getProject` function in the  project view we are re-rendering everything always

This PR reduces the scope to the required `urn`, `seed`, `profile` and `peer` information

https://user-images.githubusercontent.com/7912302/200275163-c3684aa4-0b9a-4307-856c-d268e08a2ee1.mov


rudolfs (commented):
👍

PR #474, opened 2022-11-07
Add smoke check for production builds

### LINKED ISSUE #456: Add a smoke check for production builds
The bundled production app `npm run build && npm run serve` sometimes differs from when we run it locally with `npm start`. We should add a smoke check on CI that runs a simple e2e test with the production build and checks that there are no exceptions in the dev console, to avoid accidentally deploying a broken build.

For context: this happened when we updated the transitive dependencies and didn't check the production build manually before deploying: https://github.com/radicle-dev/radicle-interface/commit/885811eb9ac6698806a9f00727cbf9729cc3547a.

Screenshot 2022-10-17 at 15 41 37

sebastinez:
This PR extends the build CI pipeline, to assert that the production build created by `npm run build`  and served by `npm run serve`, passes through the `home.spec.test` e2e test.

Closes #456

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -12,3 +12,10 @@ jobs:
       - uses: actions/checkout@v2
       - run: npm ci
       - run: npm run build
+      - name: Cypress run
+        uses: cypress-io/github-action@v4
+        with:
+          browser: chrome
+          start: npm run serve
+          wait-on: "http://localhost:4173"
+          command: npx cypress run --config baseUrl="http://localhost:4173" --spec 'cypress/e2e/home.spec.ts'


cypress/e2e/projectCommits.spec.ts
```diff
@@ -131,10 +131,7 @@ describe("project commits", () => {
 
     cy.get(".commit-teaser .badge").last().trigger("mouseenter");
     // Checking that the initial commit has the Verified badge
-    cy.get(".popup .header").should(
```
rudolfs, line 134: Why do we need this change?

rudolfs (approved):
Awesome. Works [as advertised](https://github.com/radicle-dev/radicle-interface/actions/runs/3410520385/jobs/5673575356).

PR #476, opened 2022-11-07
Make all Svelte events type-safe

rudolfs:
Adds more compile-time type checks around our usage of Svelte events everywhere in the codebase.


sebastinez (approved):
LGTM. Typez rulez!

PR #478, opened 2022-11-07
Add hash routing

sebastinez:
This PR adds hash routing as an option at build time, to be used by calling either `npm run build` for the default history navigation method or the new `npm run build:ipfs` for the new hash routing.

It prefixes every URL with a hash symbol e.g.
`https://app.radicle.xyz/#/seeds/clients.radicle.xyz/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/tree`

In case of being used with IPFS it keeps the origin and pathname and pushes our paths on top, you can see a working example here.
https://ipfs.fleek.co/ipfs/QmZTtU4QmATWpvqbRP1puabagezZtcqLNUrMy8wyAKhCkZ/#/

If we need to handle additional hash line numbers, we differentiate between the route hash and the additional line hash
https://ipfs.fleek.co/ipfs/QmZTtU4QmATWpvqbRP1puabagezZtcqLNUrMy8wyAKhCkZ/#/seeds/clients.radicle.xyz/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/tree/04aa53a1711ff9deaf9e9da5e896bc9b7d1b3995/Cargo.toml#L36

Known bugs:
- [x] Clicking on an anchor hash tag, replaces the hash part of the URL with the new hash value.. Fixed by 87ee5554
- [x] This PR isn't based on top of #481 so going back on direct search results can bring you back to a different page


scripts/build-ipfs
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -e
+
+# Generating valid index.ipfs.html file
+sed 's/src\/index.ts/src\/index.ipfs.ts/g' ./index.html > ./index.ipfs.html
+
+vite build -c ./vite.config.ipfs.ts
+cp _redirects build/_redirects
+# TODO: Eventually rollup or vite can do the renaming part from index.ipfs.html to index.html
```
sebastinez, line 9: This could be improved, if `vite build` would output the passed in `index.ipfs.html` directly as `index.html`

src/base/projects/Blob.svelte
```diff
@@ -25,16 +26,23 @@
...
 
-  $: if (line) {
-    scrollIntoView(line);
-  }
+  // Waiting onMount, due to the line numbers still loading.
+  onMount(() => {
+    if (line) {
+      scrollIntoView(line);
+    }
+  });
```
cloudhead, line 34: Strange, shouldn't the reactive statement below be enough?

src/router/definitions.ts
```diff
@@ -19,6 +19,7 @@ export interface ProjectsParams {
     | { resource: "patch"; params: { patch: string } }
     | { resource: "patches" };
   hash?: string;
```
cloudhead, line 21: Remove?

src/router/index.ts
```diff
@@ -8,6 +8,7 @@ import { unreachable } from "@app/utils";
 const documentTitle = "Radicle Interface";
 
 export const historyStore = writable<Route[]>([{ resource: "home" }]);
+let navigationMethod = "history";
```
cloudhead, line 11: This doesn't look right; it shouldn't be allowed to mutate this var at runtime.

src/base/projects/Blob.svelte
sebastinez: If we only keep the reactive statement, we don't take into account the delayed rendering of the line numbers.
But we need the reactive statemtent to handle the scrolling when a line number is clicked.

src/router/definitions.ts
sebastinez: I kept the `hash` property, so we can allow people that come to the web client with a hash param in their URL visiting a markdown document to handle that.
https://github.com/radicle-dev/radicle-interface/blob/f8ddbbf5daf5195c556218d591a1a5a64769175c/src/base/projects/Readme.svelte#L11

cloudhead: Ah ok.

src/router/index.ts
sebastinez: Still have some improvements to do but I think this is a better approach

https://github.com/radicle-dev/radicle-interface/commit/d5e43f3d43a0da86a13f20692717c4fe6f8c6fdd

rudolfs:
> In case of being used with IPFS it keeps the origin and pathname and pushes our paths on top, you can see a working example here.
https://ipfs.fleek.co/ipfs/QmZTtU4QmATWpvqbRP1puabagezZtcqLNUrMy8wyAKhCkZ/#/

This is pretty cool! 😎 

Makes me wonder though, is there a way to verify what code is actually running? Like, is there a way to guarantee that the frontend isn't malicious and doing something shady. Can we get a cryptographic hash of the code that's deployed on IPFS for example?

src/router/index.ts
```diff
@@ -8,6 +8,8 @@ import { unreachable } from "@app/utils";
 const documentTitle = "Radicle Interface";
 
 export const historyStore = writable<Route[]>([{ resource: "home" }]);
+//@ts-expect-error This value is being replaced during build.
+const navigationMethod: "history" | "hash" = "__navigationMethod__";
```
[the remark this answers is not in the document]
rudolfs: Then we wouldn't need the replace plugin and it would align with the runtime changeable config implementation that I need for the playwright tests.

vite.config.ipfs.ts
```diff
@@ -0,0 +1,28 @@
+///<reference types="vitest" />
```
rudolfs, line 1: Could we do it with an env variable instead of a whole new config?

https://vitejs.dev/guide/env-and-mode.html

sebastinez: Yeah the thing is that with @cloudhead we wanted two separate build and eventually start commands, [for context see here](https://discord.com/channels/841318878125490186/856904179820920833/1039548248945332294) 
like `npm build:ipfs `vs. `npm build`

And we wanted to avoid env variables and define the routing mode in the build step. [context](https://discord.com/channels/841318878125490186/856904179820920833/1039547141359669308)

sebastinez: I think we could define them as you mentioned here https://github.com/radicle-dev/radicle-interface/pull/478#discussion_r1021511935 and then keep both configs, since we also have to have a different base URL for hash routing than for history routing

rudolfs: _or_ :) we could still have the env var and just do:

```diff
-    "start:ipfs": "vite -c ./vite.config.ipfs.ts",
+    "start:ipfs": "HASH_ROUTING=true vite",
     "serve": "vite preview",
     "build": "scripts/build",
-    "build:ipfs": "scripts/build-ipfs",
+    "build:ipfs": "HASH_ROUTING=true scripts/build",
```

sebastinez:
> Makes me wonder though, is there a way to verify what code is actually running?

Well that's the beauty of IPFS, once you upload something to IPFS the CID is the cryptographic hash of the filesystem located under that CID.
Here a screenshot of how the CID is computed and the underlying values used including the sha256 hash
Any change to a asset, bundle, etc. has implications in the CID used in the URL
Bildschirm­foto 2022-11-14 um 16 09 38

rudolfs:
> > Makes me wonder though, is there a way to verify what code is actually running?
> 
> Well that's the beauty of IPFS, once you upload something to IPFS the CID is the cryptographic hash of the filesystem located under that CID. Here a screenshot of how the CID is computed and the underlying values used including the sha256 hash Any change to a asset, bundle, etc. has implications in the CID used in the URL

Right, it's content-addressable. 🤦 

So I guess if our pipeline is deterministic (minification, platform differences, stable dependency packaging) it should spit out the same code whose hash we could anchor somewhere for every release.

vite.config.ipfs.ts
cloudhead: It would be better if configuration happens in one place/way though, not spread accross env vars and config files. Perhaps there's a way to share most of the config between the two?

vite.config.ts
```diff
@@ -45,6 +38,8 @@ const config: UserConfig = {
     },
   },
   define: {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    USE_HASH_ROUTING: true,
```
cloudhead, line 42: Shouldn't it be false?

sebastinez: yeah should be false in this config 👍  still polishing

vite.config.ipfs.ts
sebastinez: ok @rudolfs @cloudhead I think I arrived at a way of doing configuration, env vars and config files that I feel confident to die on 😅 

Feel free to challenge me ⚔️

package.json
```diff
@@ -2,8 +2,10 @@
   "version": "1.0.0",
   "scripts": {
     "start": "vite",
+    "start:ipfs": "HASH_ROUTING=true vite --base ./",
     "serve": "vite preview",
-    "build": "scripts/build",
+    "build": "vite build",
+    "build:ipfs": "HASH_ROUTING=true vite build --base ./",
```
rudolfs, line 8: Why do we need to supply `--base ./` here?

cypress/support/e2e.ts
```diff
@@ -5,4 +5,9 @@ declare global {
   }
 }
 
-export {};
+function getPath(location: Location): string {
```
rudolfs, line 8: Wonder if we could make this based on the env var, so it is more obvious what's going on. In case we decide to keep the env var, that is.

vite.config.ipfs.ts
rudolfs: I like what you came up with.

package.json
sebastinez: Because for example with `https://ipfs.fleek.co/ipfs/QmYkhVxzVpidFbXoaGCiu1WxVA6nmzGcFnbnUKuTG7benT/#/` we need to specify the `base` as `./` so we can use a relative path for the assets.
Else we would try to fetch all the assets from `https://ipfs.fleek.co/` which wouldn't find them.

Will try to deploy an example which shows the issue at hand

rudolfs: All good, just wanted to know.

src/index.ts
```diff
@@ -4,4 +4,8 @@ const app = new App({
   target: document.body,
 });
 
+declare global {
+  const HASH_ROUTING: boolean;
```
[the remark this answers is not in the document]
sebastinez: 4446f09c

cypress/support/e2e.ts
sebastinez: 0772047e

vite.config.ts
sebastinez: okay has been resolved by [f03b97b](https://github.com/radicle-dev/radicle-interface/pull/478/commits/f03b97baf22707a4c7a3ac2a36f6a427c54e3b04)

src/router/index.ts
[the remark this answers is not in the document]
sebastinez: Implement with [f03b97b](https://github.com/radicle-dev/radicle-interface/pull/478/commits/f03b97baf22707a4c7a3ac2a36f6a427c54e3b04)

vite.config.ts
```diff
@@ -40,6 +40,8 @@ const config: UserConfig = {
   define: {
     // eslint-disable-next-line @typescript-eslint/naming-convention
     "process.env": { READABLE_STREAM: "disable" },
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    HASH_ROUTING: process.env.HASH_ROUTING !== undefined,
```
cloudhead, line 44:
```suggestion
    HASH_ROUTING: Boolean(process.env.HASH_ROUTING),
```

cloudhead: I wonder if we should just set it inside `process.env` on the line above :thinking:

src/index.ts
```diff
@@ -4,4 +4,11 @@ const app = new App({
   target: document.body,
 });
 
+declare global {
+  interface Window {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    HASH_ROUTING: boolean;
```
cloudhead, line 10: Any reason why we're not calling this `hashRouting`?

package.json
```diff
@@ -2,8 +2,10 @@
   "version": "1.0.0",
   "scripts": {
     "start": "vite",
+    "start:ipfs": "HASH_ROUTING=true vite --base ./",
```
cloudhead, line 5:
```suggestion
    "start:ipfs": "HASH_ROUTING=1 vite --base ./",
```
If we're not reading the value, this is usually the convention

src/index.ts
sebastinez: This originated due to its first use as env variable.. But since we use it here as `Window` property, we can name it `hashRouting`

vite.config.ts
sebastinez: Done with 79463006

src/index.ts
sebastinez: Renamed with 79463006

package.json
sebastinez: 79463006

.eslintrc.json
```diff
@@ -85,6 +85,7 @@
       }
     ],
     "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-namespace": ["error", { "allowDeclarations": true }],
```
rudolfs, line 88: Just for my own understanding. Why do we need this?

scripts/build
```diff
@@ -1,5 +0,0 @@
-#!/bin/sh
-set -e
-
-vite build
-cp _redirects build/_redirects
```
rudolfs, line 5: Is this going to work? The script copied the `_redirects` file to the `build` folder, but you moved it to `public`.

sebastinez: Yes the files that are in the `public` folder are copied to the `build` folder by `vite build`

.eslintrc.json
sebastinez: So we can declare `hashRouting` in the `process.env` types in `src/index.ts`

src/base/projects/Blob.svelte
```diff
@@ -2,16 +2,17 @@
...
   import { isMarkdownPath, scrollIntoView } from "@app/utils";
+  import { onMount } from "svelte";
   import { updateProjectRoute } from "@app/router";
 
+  export let activeRoute: ProjectRoute;
   export let blob: Blob;
-  export let line: string | null;
   export let getImage: (path: string) => Promise<Blob>;
-  export let activeRoute: ProjectRoute;
+  export let line: string | null;
```
rudolfs, line 15: Can we align the types so we don't have both `null` and `undefined`?

```diff
diff --git a/src/base/projects/Blob.svelte b/src/base/projects/Blob.svelte
index 9f427f3..cd7709f 100644
--- a/src/base/projects/Blob.svelte
+++ b/src/base/projects/Blob.svelte
@@ -12,9 +12,9 @@
   export let activeRoute: ProjectRoute;
   export let blob: Blob;
   export let getImage: (path: string) => Promise<Blob>;
-  export let line: string | null;
+  export let line: string | undefined = undefined;

-  $: lineNumber = line ? parseInt(line.substring(1)) : null;
+  $: lineNumber = line ? parseInt(line.substring(1)) : undefined;

   const lastCommit = blob.info.lastCommit;
   const lines = blob.binary ? 0 : (blob.content.match(/\n/g) || []).length;
diff --git a/src/base/projects/Browser.svelte b/src/base/projects/Browser.svelte
index 43bd4db..f6b9509 100644
--- a/src/base/projects/Browser.svelte
+++ b/src/base/projects/Browser.svelte
@@ -28,7 +28,7 @@
   export let activeRoute: ProjectRoute;

   $: path = activeRoute.params.path || "/";
-  $: line = activeRoute.params.line || null;
+  $: line = activeRoute.params.line;

   // When the component is loaded the first time, the blob is yet to be loaded.
   let state: State = { status: Status.Loading, path };
```

sebastinez: 54871e3d

rudolfs (approved):
Awesome work!! 🚀

PR #482, opened 2022-11-09
Fix markup being wrapped by self-closing anchor tags

### LINKED ISSUE #405: Anchor tags mess up Markdown
Self-closing anchor tags (`<a />` ) wrap markup that follows them. This might be `marked` bug.

https://app.radicle.xyz/seeds/maple.radicle.garden/rad:git:hnrkfgsbjgkqggsfoty5noz5bxr6eqqsbzsmy/tree/142d87f82ff5ca3bf0a5aa4ce25979298c67c1d1/markdown-cheatsheet.md#code

Screenshot 2022-09-08 at 12 01 25
Screenshot 2022-09-08 at 12 02 43

sebastinez:
This PR adds a new marked custom extension, that converts self-closing anchor tags into empty anchor tags, to avoid wrapping of those anchor tags on adjacent markup.

Bildschirm­foto 2022-11-09 um 22 56 40

----------------------------------------------

Bildschirm­foto 2022-11-09 um 22 57 51

Closes #405


rudolfs:
Seems like there's a bug somewhere. Also wonder if we should maybe submit a bug report to marked, because it should really be fixed on their end.

Screenshot 2022-11-14 at 08 52 30

Screenshot 2022-11-14 at 08 53 24
Screenshot 2022-11-14 at 08 53 27
Screenshot 2022-11-14 at 08 53 32
Screenshot 2022-11-14 at 08 53 35
Screenshot 2022-11-14 at 08 53 39
Screenshot 2022-11-14 at 08 53 42


http://127.0.0.1:3000/seeds/maple.radicle.garden/markdown/tree/main/markdown-cheatsheet.md

sebastinez:
Thanks @rudolfs , yeah the bracket closing showing up is a "known" bug 😅 

I found this interesting issue that shows that GitHub also doesn't really standardize how to handle those self closing anchor tags.
https://github.com/Microsoft/vscode/issues/7675

Not certain this is a bug on marked.js side, since its not defined by the [original markdown](https://daringfireball.net/projects/markdown/syntax), the [gfm spec](https://github.github.com/gfm), nor the [commonmark spec](https://spec.commonmark.org/0.30/) how to handle header ids at all.
So since this comes from a custom way of injecting anchor hashes by GitHub, IMO either we roll our custom (tokenizer/renderer) or we don't support that non-standard case by GitHub.

The easiest way of making anchor hashes work in markdown is e.g. 

```markdown
[#first-header](First header)
[#second-header](Second header)
[#third-header](Third header)

# First header
# Second header
# Third header
```

Which would output something like
```html
<a href="#first-header">First header</a>
<a href="#second-header">Second header</a>
<a href="#third-header">Third header</a>

<h1 id="first-header">First header</h1>
<h1 id="second-header">Second header</h1>
<h1 id="third-header">Third header</h1>
```

Which IMO is nicer to write and doesn't require to inject custom HTML.
The downside is that if the author uses e.g. emojis in the headers we need to strip them and the author has to take that into account when writing their TOC

sebastinez:
> Seems like there's a bug somewhere. 

Fixed by faf6d88146da4dbb3235628da0528bf230651428

sebastinez:
I would also suggest we take a good look at `dompurify` which seems to also have some issue with self closing tags in html
If you try to pass

```html
<a name="header"></a>

<h2 id="header">Hello</h2>
```

in https://cure53.de/purify you get the wrapping behaviour we are trying to remove, am looking if we can fix it there.

sebastinez:
Ok in the end I think, it's a combined issue between `marked` that doesn't handle self closing html anchor hash tags well and `dompurify` which is very strict on `anchor` tags and strips the attributes too eagerly. 🤷

rudolfs:
> Ok in the end I think, it's a combined issue between marked that doesn't handle self closing html anchor hash tags well and dompurify which is very strict on anchor tags and strips the attributes too eagerly. 🤷

Yeah, that rings a bell, I think I arrived at the same conclusion back when I looked into it.

rudolfs:
Seeing a lot more warnings after the change to `sanitize-html`, it might not be a problem, but we should look into it.

Screenshot 2022-11-14 at 13 05 37

rudolfs:
This seems like a regression:

Screenshot 2022-11-14 at 12 56 30
vs production
Screenshot 2022-11-14 at 12 56 27

rudolfs:
The links in the RIPs TOC stopped working:
http://127.0.0.1:3000/seeds/clients.radicle.xyz/rad:git:hnrkx7ruakgb9ozzi7atwb1rh6moq8pjonp6o/tree/8cd1561688462e2fcb234956b6f5916c42bdf4fa/0001-heartwood.md

Screenshot 2022-11-14 at 13 16 41

sebastinez:
Thanks @rudolfs for the catches, still working on ironing out some of these things

rudolfs (approved):

PR #483, opened 2022-11-10
Convert emojis to twemojis

### LINKED ISSUE #356: Emojis differ depending on which OS you're on
It would be nice if the emoji assets would be the same regardless of which OS you're looking at the app from. A possible workaround is to use [Twemoji assets](https://twemoji.twitter.com/). We already implemented this in Upstream [[1]](https://github.com/radicle-dev/radicle-design-system/blob/main/Emoji.svelte)[[2]](https://github.com/radicle-dev/radicle-design-system/blob/main/scripts/install-twemoji-assets.sh), it would be quite easy to port it over.

macOS:

Screenshot 2022-09-01 at 16 50 27

Windows:

Screenshot 2022-09-01 at 16 45 27

Linux:

Screenshot_from_2022-09-01_16-47-10

sebastinez:
To provide the same emoji experience on all browsers and OS we use Twemoji and replace all emojis where possible.

Okay I think I did a pretty good work with adding the `twemoji` action were necessary.
I didn't want to do a general twemoji parsing over the entire page, to allow us to be specific where we want to have twemoji involved..
And I think a svelte action is something very idiomatic to svelte, with a small footprint.

I'm open for any suggestions.

Closes #356

### DIFF
--- a/public/typography.css
+++ b/public/typography.css
@@ -103,3 +103,9 @@ p {
   color: var(--color-foreground);
   border-bottom-color: var(--color-foreground);
 }
+.txt-emoji {
+  height: 1em;
+  width: 1em;
+  margin: 0 0.05em 0 0.1em;
+  vertical-align: -0.1em;
+}
--- a/src/Dropdown.svelte
+++ b/src/Dropdown.svelte
@@ -1,6 +1,7 @@
 <script lang="ts" strictEvents>
   import { createEventDispatcher } from "svelte";
   import Badge from "@app/Badge.svelte";
+  import { twemoji } from "@app/utils";
 
   export let items: {
     key: string;
@@ -54,6 +55,7 @@
       <div
         class="dropdown-item"
         class:selected={value === selected}
+        use:twemoji
         on:click={() => onSelect(value)}
         {title}>
         {@html key}
--- a/src/ErrorModal.svelte
+++ b/src/ErrorModal.svelte
@@ -1,9 +1,10 @@
 <script lang="ts" strictEvents>
   import type { Err } from "@app/error";
 
-  import { createEventDispatcher } from "svelte";
-  import Modal from "@app/Modal.svelte";
   import Button from "@app/Button.svelte";
+  import Modal from "@app/Modal.svelte";
+  import { createEventDispatcher } from "svelte";
+  import { twemoji } from "@app/utils";
 
   const dispatch = createEventDispatcher<{ close: never }>();
 
@@ -20,7 +21,7 @@
 </script>
 
 <Modal on:close error {floating} {subtle}>
-  <span slot="title">
+  <span slot="title" use:twemoji>
     {#if emoji}
       <div>{emoji}</div>
     {/if}
--- a/src/Markdown.svelte
+++ b/src/Markdown.svelte
@@ -5,8 +5,10 @@
   import type * as proj from "@app/project";
   import {
     markdownExtensions as extensions,
+    renderer,
     getImageMime,
     isUrl,
+    twemoji,
     scrollIntoView,
   } from "@app/utils";
   import dompurify from "dompurify";
@@ -17,12 +19,12 @@
   export let doc = matter(content);
 
   const frontMatter = Object.entries(doc.data);
-  marked.use({ extensions });
+  marked.use({ extensions, renderer });
 
   let container: HTMLElement;
 
   const render = (content: string): string =>
-    dompurify.sanitize(marked.parse(content));
+    dompurify.sanitize(marked.parse(content, { headerIds: false }));
 
   onMount(() => {
     // Don't underline <a> tags that contain images.
@@ -43,7 +45,7 @@
       const path = i.getAttribute("src");
 
       // Make sure the source isn't a URL before trying to fetch it from the repo
-      if (path && !isUrl(path)) {
+      if (path && !isUrl(path) && !path.startsWith("/twemoji")) {
         getImage(path).then(blob => {
           if (blob.content) {
             const mime = getImageMime(path);
@@ -267,6 +269,6 @@
   </div>
 {/if}
 
-<div class="markdown" bind:this={container}>
+<div class="markdown" bind:this={container} use:twemoji>
   {@html render(doc.content)}
 </div>
--- a/src/Placeholder.svelte
+++ b/src/Placeholder.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
-  export let icon: string;
+  import { twemoji } from "@app/utils";
+
+  export let emoji: string;
 </script>
 
 <style>
@@ -17,14 +19,14 @@
     padding: 1rem 0;
     font-weight: var(--font-weight-bold);
   }
-  .placeholder .icon {
+  .placeholder .emoji {
     margin-bottom: 1rem;
   }
 </style>
 
 <div class="placeholder">
   <header>
-    <div class="icon txt-large">{icon}</div>
+    <div class="emoji txt-large" use:twemoji>{emoji}</div>
     <slot name="title" />
   </header>
   <slot name="body" />
--- a/src/SeedAddress.spec.ts
+++ b/src/SeedAddress.spec.ts
@@ -15,7 +15,12 @@ describe("SeedAddress", () => {
         port: 8776,
       },
     });
-    cy.get("span.seed-icon").should("have.text", "🌱");
+    cy.get("span.seed-icon img").should("have.attr", "alt", "🌱");
+    cy.get("span.seed-icon img").should(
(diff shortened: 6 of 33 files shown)


src/utils.ts
```diff
@@ -520,4 +522,30 @@ const katexMarkedExtension = {
...
+  // Add escaped header ids for anchor hash browsing
+  heading(text: string, level: 1 | 2 | 3 | 4 | 5 | 6) {
+    const escapedText = text
+      .toLowerCase()
+      .replace(/[^\w]+/g, "-")
+      .replace(/^-|-$/g, "");
+
+    return `<h${level} id="${escapedText}">${text}</h${level}>`;
+  },
+};
```
cloudhead, line 535: Can you document the purpose of this function?

cloudhead:
I like the approach :+1:

src/utils.ts
sebastinez: yeah sure. np

src/SeedAddress.spec.ts
```diff
@@ -15,7 +15,12 @@ describe("SeedAddress", () => {
         port: 8776,
       },
     });
-    cy.get("span.seed-icon").should("have.text", "🌱");
+    cy.get("span.seed-icon img").should("have.attr", "alt", "🌱");
+    cy.get("span.seed-icon img").should(
+      "have.attr",
+      "src",
+      "https://twemoji.maxcdn.com/v/14.0.2/72x72/1f331.png",
```
rudolfs, line 22: We should bundle the twemoji assets and serve them ourselves. Have a look at 
  - https://github.com/radicle-dev/radicle-upstream/blob/main/scripts/install-twemoji-assets.sh
  - https://github.com/radicle-dev/radicle-upstream/blob/main/package.json#L35
for inspiration.

sebastinez: hey, yeah thats nice! Did remember that you did something like this on Upstream, this would allow us to not depend on the tweemoji cdn 👍

rudolfs:
> And I think a svelte action is something very idiomatic to svelte, with a small footprint.
> I'm open for any suggestions.

An alternative could be just using a component:

https://github.com/radicle-dev/radicle-upstream/blob/main/design-system/Emoji.svelte

sebastinez:
> > And I think a svelte action is something very idiomatic to svelte, with a small footprint.
> > I'm open for any suggestions.
> 
> An alternative could be just using a component:
> 
> https://github.com/radicle-dev/radicle-upstream/blob/main/design-system/Emoji.svelte

Yeah that's something I pursued at some point, but I felt like writing simple text and having to inline an Emoji component, felt a bit hacky, vs. a use action on the wrapping element.
But the end output would be the same so 🤷‍♂️ depends on what we prefer for DX I guess

src/SeedAddress.spec.ts
sebastinez: 7740f4519b138ed04d855e36abcf2cfea547e8c5

src/utils.ts
sebastinez: 6cef9f10

src/Form.svelte
```diff
@@ -267,7 +268,7 @@
           {/if}
         </div>
       {:else}
-        <div class="description txt-faded txt-small">
+        <div class="description txt-faded txt-small" use:twemoji>
```
rudolfs, line 271: Do we have emojis in any of the descriptions?

src/ErrorModal.svelte
```diff
@@ -20,7 +21,7 @@
 </script>
 
 <Modal on:close error {floating} {subtle}>
-  <span slot="title">
+  <span slot="title" use:twemoji>
```
rudolfs, line 24: This could be moved to the inner `div`.

src/Dropdown.svelte
```diff
@@ -54,6 +55,7 @@
       <div
         class="dropdown-item"
         class:selected={value === selected}
+        use:twemoji
```
rudolfs, line 58: Where do we use emojis in dropdowns?

src/Placeholder.svelte
```diff
@@ -1,4 +1,6 @@
 <script lang="ts">
+  import { twemoji } from "@app/utils";
+
   export let icon: string;
```
rudolfs, line 4: Wonder if we should rename this prop to `emoji`?

src/utils.ts
```diff
@@ -520,4 +522,40 @@ const katexMarkedExtension = {
     }),
 };
 
+// Overwrites the rendering of heading tokens.
+// Since there are possible non ASCII characters in headings,
+// we escape them by replacing them with dashes and,
+// trim eventual dashes on each side of the string.
```
rudolfs, line 528: Do you have a concrete example of where this happens?

src/utils.ts
```diff
@@ -520,4 +522,40 @@ const katexMarkedExtension = {
...
+  };
+
+  twemojiModule.parse(node, options);
+  afterUpdate(() => {
+    twemojiModule.parse(node, options);
+  });
+  return {
+    update() {
+      console.warn(
+        "Changing twemoji options after the action was mounted is not possible.",
```
rudolfs, line 557: How could this even happen?

rudolfs:
Seeing emoji related errors in source browsing.

Screenshot 2022-11-14 at 13 44 52

rudolfs:
> Yeah that's something I pursued at some point, but I felt like writing simple text and having to inline an Emoji component, felt a bit hacky, vs. a use action on the wrapping element.

Yeah, I see your point. It also doesn't work on markdown or where the input text isn't ours.

> But the end output would be the same so 🤷‍♂️ depends on what we prefer for DX I guess

True. My only worry with svelte actions is that they're easier to misuse than components, e.g. by applying them too far up the component tree or applying them in places where it doesn't make sense.

I guess I don't have a clear favourite here, so let's go with actions for now. 🤷

src/Markdown.svelte
```diff
@@ -17,12 +19,12 @@
...
 
   const frontMatter = Object.entries(doc.data);
-  marked.use({ extensions });
+  marked.use({ extensions, renderer });
 
   let container: HTMLElement;
 
   const render = (content: string): string =>
-    dompurify.sanitize(marked.parse(content));
+    dompurify.sanitize(marked.parse(content, { headerIds: false }));
```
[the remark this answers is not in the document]
sebastinez: If we let `marked` generate the header ids automatically it takes emojis into account e.g. 
`<h2 id="⛺️-where-to-find-us">⛺️ Where to find us</h2>` so if we pass this option and then do the escaping ourselves in the `renderer` function we can control that.

src/Form.svelte
sebastinez: Can be removed 👍

src/Dropdown.svelte
sebastinez: For example the `SeedDropdown` can use emojis, once used
`const key = `${seed.emoji} ${seed.host}`;`

src/ErrorModal.svelte
sebastinez: Yep we could, the thing is that sometimes `title` can be an emoji too..

src/Placeholder.svelte
sebastinez: fa2ea291

src/Form.svelte
sebastinez: b7e034ac

src/utils.ts
sebastinez: http://127.0.0.1:3000/seeds/clients.radicle.xyz/rad:git:hnrkmg77m8tfzj4gi4pa4mbhgysfgzwntjpao/tree/04aa53a1711ff9deaf9e9da5e896bc9b7d1b3995

Here are a few headings that take emojis, and then are a PITA

sebastinez:
> Screenshot 2022-11-14 at 13 44 52

3b1517c4e745c5aed00b9702c99e67225d15a6bc

src/utils.ts
sebastinez: 8f9cf0d0989e07edaf2e9452977b6e1395934400

public/typography.css
```diff
@@ -103,3 +103,10 @@ p {
   color: var(--color-foreground);
   border-bottom-color: var(--color-foreground);
 }
+
+img.emoji {
```
rudolfs, line 107: Could we change this to `.txt-emoji` and also change the class name in the twemoji options, so it aligns with the rest of the global `.txt-*` classes?

sebastinez: e855171e

rudolfs (approved):
⭐ 💯 🚀 ✨

PR #484, opened 2022-11-10
Change monospace font to JetBrains Mono

### LINKED ISSUE #282: Allow user to customize monospace font
The default monospace font is … jarring. Would be nice to be able to set something more thoughtful and also to allow users to customize it to whatever they are most comfortable with.

sebastinez:
This PR changes the default monospace font used in the web client to [JetBrains Mono](https://www.jetbrains.com/lp/mono/)
If JetBrainsMono isn't installed locally it falls back to the default monospace font, in MacOS it's Courier

Closes #282

### DIFF
--- a/public/typography.css
+++ b/public/typography.css
@@ -22,7 +22,7 @@
 
 :root {
   --font-family-sans-serif: Inter, sans-serif;
-  --font-family-monospace: monospace;
+  --font-family-monospace: "JetBrains Mono", monospace;
   --font-weight-normal: 400;
   --font-weight-medium: 600;
   --font-weight-bold: 700;


brandonhaslegs:
You need to import the font for this to work right?

sebastinez:
> You need to import the font for this to work right?

Yes, the user would need to install the `JetBrains Mono` font on their machine, and if not available it would go fall back to the default OS font.

brandonhaslegs:
Right so the font file needs to be added so that everyone sees it.

sebastinez:
> Right so the font file needs to be added so that everyone sees it.

Yes the thing is that if we add the font file, we don't allow users to use their own fonts, as mentioned in the originating issue.
>  For linux users it's still useful to be able to set it to monospace as that allows you to have the same font in your editor as well as on the web without requiring the font to be available as an option. - cloudhead

So we end up forcing everybody to use `JetBrains Mono`.

Would like your opinion on that @cloudhead

brandonhaslegs:
So the feature should be designed first. It should be a setting similar to dark/light mode where we give the user a few options we select, and there's also an option to use the system monospace font. As is, this PR doesn't really do much, just shows users JetBrains if they have it installed (which means everyone on macOS still sees the default monospace font, which is the weird thing. I designed something similar in Upstream: 
image

sebastinez:
> So the feature should be designed first. It should be a setting similar to dark/light mode where we give the user a few options we select, and there's also an option to use the system monospace font. As is, this PR doesn't really do much, just shows users JetBrains if they have it installed, which means almost everyone on macOS still sees the default monospace font, which is the weird thing. I designed something similar in Upstream: image

Ok understood in the end 😅 
IMO the upstream design could fit good inside the web client, will build something on top of it.

brandonhaslegs:
Would be cool to have a settings page for sure! We could move the light/dark mode setting there too.

sebastinez:
Still some persistance missing, probably in `localStorage` but already coming nicely.
Would like help with a nice svg path for a gear icon for settings, don't really like the current one.

https://user-images.githubusercontent.com/7912302/201401070-c0b6dc72-7628-4240-a3e9-b05bd230efdf.mov

sebastinez:
Superseded by #486

PR #485, opened 2022-11-11
Add a first exploration of the component documentation

sebastinez:
So this is a first work on documenting components.

I think it doesn't make much sense to do so for every component, but the ones that are more often used could benefit from it.

So far I think a good structure would be:
```
## Usage
`one or multiple lines on how this component should be used`

## Best practices
`a list of things that should be taken into account, for code style, perf, etc.`

## Props
`the props that the component expects, eventually also the defaults.
```

Feel free to add your opinion on it

### DIFF
--- a/src/Address.svelte
+++ b/src/Address.svelte
@@ -1,8 +1,31 @@
+<!--
+ @component
+ ## Usage
+ An inline element to show a ETH address or profile information (name and avatar).  
+ Displays a blockie avatar if no profile or avatar is found  
+ Allows the navigate to profiles or external sites, e.g. multisig safes
+ 
+ ## Best Practices
+ - Where possible we should pass in a resolved `profile`.
+ 
+ ## Props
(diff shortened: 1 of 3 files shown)


rudolfs:
I think the intention of making the code base more accessible is a good one. However, I'm not convinced that adding doc-strings and comments is the most effective way. In my experience comments usually require a lot of work to maintain, otherwise they go out of date very quickly.

I think components should be rather small and not do too much, so that we get the gist by just looking at the component. Since we're using typescript, the interface of a component should be self explanatory if we use descriptive variable names and descriptive string literal types for configuration options.

So components and their code should be simple and self-descriptive, otherwise we're doing it wrong and should take a step back and refactor by breaking them down into simpler or more composable ones.

sebastinez:
I think I'll look into some way to generate documentation in a more programatic way, since I agree with @rudolfs this is too much work to maintain

PR #491, opened 2022-11-15
Fix twemoji base

sebastinez:
Twemojis were imported with an absolute base, we need to listen to `window.hashRouting` and switch to a relative base if we use hash routing


rudolfs (approved):
🍒

PR #493, opened 2022-11-16
Removes `compilerOptions.dev` and chunk size warnings on build

sebastinez:
We split some third party packages into their own bundles to reduce the size of the index.js bundle.
We also don't need hmr ever in production.

Also removes the sourcemap output in build to reduce the size of the total bundle.
In dev environment we still get an inlined sourcemap, to help debugging.

Improves the speed also a bit, "about 200ms locally hosted" since we don't have to wait for e.g. markdown, siwe, walletconnect packages for a first contentful paint, since they have a considerable size

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -19,7 +19,6 @@ const config: UserConfig = {
   },
   plugins: [
     svelte({
-      hot: !process.env.VITEST,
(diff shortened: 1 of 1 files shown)


cloudhead (approved):
Love it

PR #496, opened 2022-11-16
Move plausible.js to npm package and improve base handling between hash and history routing

sebastinez:
More optimization, and some refactoring

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -3,6 +3,7 @@
   import { getWallet } from "@app/wallet";
   import { initialize, activeRouteStore } from "@app/router";
   import { twemoji, unreachable } from "@app/utils";
+  import Plausible from "plausible-tracker";
 
   import ColorPalette from "@app/ColorPalette.svelte";
   import Faucet from "@app/base/faucet/Routes.svelte";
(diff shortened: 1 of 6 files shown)


src/App.svelte
```diff
@@ -72,7 +72,7 @@
     <script
       defer
       data-domain="app.radicle.xyz"
-      src="https://plausible.io/js/plausible.js"></script>
+      src={`${base}plausible.js`}></script>
```
rudolfs, line 75: Will this even work if we serve it from our own domain? I'm not familiar with how plausible works, but I wonder if it's not a requirement that it gets pulled from their servers.

Their docs even mention proxying instead of bundling their script: https://plausible.io/docs/proxy/introduction.

sebastinez: Hmm good catch 🤔 
My idea was to avoid fetching the js file, which delayed the load completion of the site.
Maybe we could remove the script altogether and use the npm package that doesn't rely on downloading the script?
https://github.com/plausible/plausible-tracker wdyt?

rudolfs: Yeah, the npm package sounds good.

sebastinez: b1cb59a

src/router/index.ts
```diff
@@ -16,6 +16,8 @@ export const activeRouteStore: Readable<Route> = derived(
   },
 );
 
+export const base = process.env.hashRouting ? "./" : "/";
```
rudolfs, line 19: 😍

rudolfs (approved):
👍 

Since there's no way to test this before deploying, could you keep an eye on the stats for the next day or so to make sure it actually works?

https://plausible.io/app.radicle.xyz

PR #498, opened 2022-11-16
Dynamic imports

sebastinez:
Ok this is a test, which I think will give the page even more snappiness, instead of loading all components on start, we split all the components in separate bundles, and load/import them dynamically when needed.

This saves us around 600 KB in data resources to be transferred when landing for the first time on the page.
Also I get around 300 ms quicker `DOMContentLoaded` when visiting `base/home/Index.svelte`

I deployed the latest commit here, so you can see how it behaves
https://tubular-taiyaki-bf2c09.netlify.app

Also I moved some bigger packages like `katex` into a separate `markdown` module

For desktop computer with unthrottled bandwith the difference is not significant, but for lower connections, to download 600 KB less trims the load time a good amount.
These tests were run with the following network throttle
`150 ms TCP RTT, 1.638,4 Kbps throughput (Simulated)` similar to a slow 4G mobile connection
## Current
app radicle xyz

## Dynamic Imports
tubular-taiyaki

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -7,7 +7,7 @@
   import Header from "@app/Header.svelte";
   import Loading from "@app/Loading.svelte";
   import Modal from "@app/Modal.svelte";
-  import Preload from "@app/router/Preload.svelte";
+  import DynamicComponentImport from "@app/router/DynamicComponentImport.svelte";
 
   initialize();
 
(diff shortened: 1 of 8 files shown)


src/App.svelte
```diff
@@ -83,33 +75,47 @@
       <Loading center />
     </div>
   {:then wallet}
-    <ColorPalette />
     <Header session={$session} {wallet} />
     <div class="wrapper">
       {#if $activeRouteStore.resource === "home"}
-        <Home />
+        <DynamicComponentImport path="../base/home/Index.svelte" />
```
rudolfs, line 81: We lose all type safety here. I think I prefer to keep the type safety over the performance gain.

PR #499, opened 2022-11-18
Add middle click behaviour

sebastinez:

### DIFF
--- a/src/components/Button.svelte
+++ b/src/components/Button.svelte
@@ -117,6 +117,7 @@
   {disabled}
   {style}
   on:click|stopPropagation
+  on:auxclick
   on:focus
   on:blur
   on:mouseout
--- a/src/components/Connect/ConnectWallet.svelte
+++ b/src/components/Connect/ConnectWallet.svelte
@@ -84,6 +84,7 @@
         variant="secondary"
         size="small"
         on:click={onClickConnect}
+        on:auxclick={onClickConnect}
         disabled={!wallet.metamask.signer}>
         Connect with Metamask
       </Button>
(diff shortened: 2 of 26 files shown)


src/router/index.ts
```diff
@@ -62,31 +62,41 @@ export function projectLinkHref(
...
 }
 
-export const push = (newRoute: Route): void => {
+export const push = (newRoute: Route, e?: MouseEvent): void => {
   const history = get(historyStore);
 
-  // Limit history to a maximum of 10 steps. We shouldn't be doing more than
-  // one subsequent pop() anyway.
-  historyStore.set([...history, newRoute].slice(-10));
+  if (e?.button === 1) {
```
sebastinez, line 87: Hey @rudolfs this PR is working great for adding middle click behaviour and I'm happy to be able to remove a lot of event dispatches.

The thing I'm not 100% certain is if the checking here against `e.button` is the best one.. since users are able to change their pointer devices configuration so 1 could be left click or right click or whatever else click..
The other way would be to start with `onauxclick` and later see if some better solution comes up for Safari users, wdyt?

rudolfs: I guess we can try this and see if people complain? 🤷
Don't know what the best solution is here.

sebastinez: Ok cool, I can live with that

sebastinez:
Ok passing this PR to ready for review, there is still some weird things happening on Firefox, hopefully I can fix those but would like to get this PR out before an eventual feature freeze

sebastinez:
Will close this for now, due to too much conflicts, and low priority

PR #514, opened 2022-11-25
Handle errors in Browser.svelte

sebastinez:
This PR improves the error messages given in the `Browser` component

We can differentiate in the same Blob window between:
- Not able to load a blob on initial mount

https://user-images.githubusercontent.com/7912302/208771728-30d24da6-2c51-4e6b-8b07-c846adef87fa.mp4

- Not able to fetch and expand tree

https://user-images.githubusercontent.com/7912302/208771832-ab334f32-ee70-4952-a808-c89fc9966891.mp4

- Not able to load a selected file

https://user-images.githubusercontent.com/7912302/208771891-2e2e9222-fba1-433e-ab59-cc15694c9572.mp4

- Handle and give some console.warn if a local image to be shown in a markdown document can't fetched

https://user-images.githubusercontent.com/7912302/208771631-73a8ab69-2f94-45fa-8cd7-e8ecce8b1c62.mp4


rudolfs:
I don't see a placeholder if I update the URL to point to a non-existent file. Looking at your video, was this PR supposed to fix that?

Screenshot 2022-12-01 at 17 13 53

sebastinez:
Yeah I'm working on some improvements, and writing tests to support this PR 😎 
Will convert it back to draft for now

tests/e2e/project/browser.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { test, expect } from "@tests/support/fixtures.js";
+
+const sourceBrowsingFixture =
+  "/seeds/0.0.0.0/rad:git:hnrkgd7sjt79k4j59ddh11ooxg18rk7soej8o";
+
+test.describe("browser error handling", () => {
+  test("check that error is shown when not able to load folder", async ({
```
rudolfs, line 7: It would be nice if we'd write the test subject so that it reads like a sentence.

Now:
`test("check that error is shown when not able to load folder"`

Test check that error is shown...

Suggestion:
`test("show error when folder can't be loaded"`

Test show error when folder can't be loaded.

tests/e2e/project/browser.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { test, expect } from "@tests/support/fixtures.js";
```
rudolfs, line 1: I think we have most source browsing tests in `project.spec.ts`. I'd move these tests there.

src/base/projects/Browser.svelte
```diff
@@ -60,26 +69,45 @@
...
   // Get an image blob based on a relative path.
-  const getImage = async (imagePath: string): Promise<proj.Blob> => {
+  const getImage = async (
+    imagePath: string,
+  ): Promise<proj.Blob | undefined> => {
     const finalPath = utils.canonicalize(imagePath, path);
-    return project.getBlob(commit, finalPath, { highlight: false });
+    try {
+      return project.getBlob(commit, finalPath, { highlight: false });
+    } catch (e: any) {
```
rudolfs, line 87: I think it's more correct to do `catch (e: unknown)`. But maybe we should remove it, since we don't use the error as far as I can see?

sebastinez: Yeah removing it, doesn't work with prettier svelte plugin. But we can remove the type so I think it defaults to `unknown`

sebastinez: d246e3c4

tests/e2e/project/browser.spec.ts
sebastinez: 1ea4b49d

tests/e2e/project/browser.spec.ts
sebastinez: 1ea4b49d

sebastinez (commented):
closing my own review 😅

src/base/projects/Browser.svelte
rudolfs: This should work:
```diff
-    } catch (e: any) {
+    } catch {
```

sebastinez: TIL 😅

sebastinez: 9f866d72

src/base/projects/Tree/Folder.svelte
```diff
@@ -1,20 +1,28 @@
...
-  export let fetchTree: (path: string) => Promise<Tree>;
+  export let fetchTree: (path: string) => Promise<Tree | undefined>;
   export let name: string;
   export let prefix: string;
   export let currentPath: string;
   export let loadingPath: string | null = null;
 
   let expanded = currentPath.indexOf(prefix) === 0;
-  let tree: Promise<Tree> | null = expanded ? fetchTree(prefix) : null;
+  let tree: Promise<Tree | undefined> | null = expanded
```
rudolfs, line 17: Why do we mix undefined and null here?

tests/e2e/project.spec.ts
```diff
@@ -337,3 +337,58 @@ test("only one modal can be open at a time", async ({ page }) => {
...
+    await page.route(
+      "**/v1/projects/rad:git:hnrkgd7sjt79k4j59ddh11ooxg18rk7soej8o/blob/530aabdcc80397af254bc488b767169b92496e81/.hidden?highlight=true&theme=base16-ocean.dark",
+      route => route.fulfill({ status: 500 }),
+    );
+
+    await page.goto(`${sourceBrowsingFixture}/tree/master/.hidden`);
+
+    const error = page.locator("text=Not able to load file");
+    await expect(error).toBeVisible();
+  });
```
rudolfs, line 376: Aren't these two tests testing the same thing?

tests/e2e/project.spec.ts
```diff
@@ -337,3 +337,58 @@ test("only one modal can be open at a time", async ({ page }) => {
...
+  test("error appears when navigating to nested folder that can't be loaded", async ({
```
rudolfs, line 377: And this test seems to also test the same thing as `error appears when folder can't be loaded`, no? Or are we hitting different code paths with it?

tests/e2e/project.spec.ts
sebastinez: One is checking that the error shows on internal navigation and the other when browsing to a path from the browser address bar

src/base/projects/Browser.svelte
```diff
@@ -60,26 +69,45 @@
...
+  const getImage = async (
+    imagePath: string,
+  ): Promise<proj.Blob | undefined> => {
     const finalPath = utils.canonicalize(imagePath, path);
-    return project.getBlob(commit, finalPath, { highlight: false });
+    try {
+      return project.getBlob(commit, finalPath, { highlight: false });
+    } catch {
+      browserErrorStore.set({
+        message: "Not able to load image blob",
```
rudolfs, line 89: Could we test this path?

src/base/projects/Browser.svelte
```diff
@@ -91,15 +119,27 @@
   };
 
   const fetchTree = async (path: string) => {
-    return project.getTree(commit, path);
+    try {
+      return project.getTree(commit, path);
+    } catch {
+      browserErrorStore.set({
+        message: "Not able to fetch directory structure",
```
rudolfs, line 126: Could we test this path?

src/base/projects/Browser.svelte
```diff
@@ -209,28 +249,30 @@
...
             </span>
             <span slot="body">
-              {#if path === "/"}
-                The README could not be loaded.
-              {:else}
-                This path could not be loaded.
-              {/if}
+              <span>
+                {#if $browserErrorStore.path === "/"}
+                  The README could not be loaded.
```
rudolfs, line 260: And this one? : )

sebastinez: Is done in latest commits

src/base/projects/Browser.svelte
sebastinez: Is done in latest commits

src/base/projects/Browser.svelte
sebastinez: Is done in latest commits

tests/e2e/project.spec.ts
sebastinez: Yeah I removed it

src/base/projects/Tree/Folder.svelte
sebastinez: Has been rewritten to only use `undefined`

tests/e2e/project.spec.ts
```diff
@@ -335,29 +335,100 @@ test("only one modal can be open at a time", async ({ page }) => {
...
+      "**/v1/projects/rad:git:hnrkdi8be7n4hhqoz9rpzrgd68u9dr3zsxgmy/blob/fcc929424b82984b7cbff9c01d2e20d9b1249842/.hidden",
+      route => route.fulfill({ status: 500 }),
+    );
+
+    await page.goto(projectFixtureUrl);
+    await page.locator("text=.hidden").click();
+
+    await expect(page.locator("text=Not able to load file")).toBeVisible();
+  });
+  test.only("error appears when README can't be loaded", async ({ page }) => {
```
rudolfs, line 392: The `.only` makes the CI test suite fail.

src/base/projects/Tree/Folder.svelte
```diff
@@ -23,9 +25,9 @@
   const onClick = () => {
     expanded = !expanded;
 
-    if (expanded) {
-      tree = fetchTree(prefix);
-    }
+    tree = fetchTree(prefix).then(tree => {
```
rudolfs, line 28: Could we convert this to the `await` syntax?

PR #522, opened 2022-12-01
Cache playwright binaries

sebastinez:
## First commit that installs Playwright browsers
Running `npx playwright install --with-deps`
Bildschirm­foto 2022-12-01 um 21 03 41

## With the Playwright browsers installed it skips the installment and only install OS dependencies
Running `npx playwright install-deps`
Bildschirm­foto 2022-12-01 um 21 09 22


rudolfs (approved):
⚡

PR #532, opened 2022-12-12
Update dependencies

sebastinez:
Includes the move from the `svelte-preprocess` package for preprocessing to `@sveltejs/vite-plugin-svelte` which allows us to use Vite also for preprocessing.

Also had to remove the `vite-plugin-rewrite-all` again since they don't support vite@4. So I wrote the plugin myself and without the dependecies used by `vite-plugin-rewrite-all`.


rudolfs (approved):
✨ 🚀

PR #536, opened 2022-12-12
Rewrites url for `vite preview` server paths

sebastinez:
This PR fixes an issue when reloading a page served through `npm run serve` and then trying to reload it.
Since the URL was being passed literally as e.g. `/seeds/clients.radicle.xyz` the server wasn't able to find it.

So the PR checks that the request is not related to a public asset or bundle and if it's a url path it rewrites it to `/index.html`


rudolfs (approved):
🚀

PR #539, opened 2022-12-14
Add visual snapshot tests

rudolfs:
We abuse the Github cache action to store generated snapshots when pushing to master and then load them in pull requests and run the visual tests against it. This way we don't have to commit the images to the repository and always get visual diffs against the latest master.

Here's how the proposed workflow could look like:

- When this check fails, it means that there are visual differences in the proposed pull request.

  Screenshot 2022-12-15 at 21 01 59

- In this case the reviewer has to download the artifacts and check that the visual changes are not regressions, but intentional changes.

  Screenshot 2022-12-15 at 21 03 37

- In this case it seems like there is a visual regression that was not intended:

  landing-page-1-actual
  landing-page-1-diff
  landing-page-1-expected

- We ask the PR author to fix the problem
- Author fixes the issue, pushes the changes and the check passes
- We merge to master and the snapshots automatically get updated in case the tests got adjusted

A potential follow up could be to push the failed test images to GCP and link them in a PR comment automatically, so we don't have to download the zip file and inspect it.

An example run: https://github.com/rudolfs/radicle-interface/pull/5.

### DIFF
--- a/tests/support/fixtures.ts
+++ b/tests/support/fixtures.ts
@@ -139,7 +139,10 @@ export const test = base.extend<{
     await Fs.mkdir(stateDir, { recursive: true });
 
     await use(stateDir);
-    if (process.env.CI && testInfo.status === "passed") {
+    if (
+      process.env.CI &&
+      (testInfo.status === "passed" || testInfo.status === "skipped")
+    ) {
       await Fs.rm(stateDir, { recursive: true });
     }
   },
--- a/tests/support/fixtures.ts
+++ b/tests/support/fixtures.ts
@@ -156,3 +156,24 @@ function log(text: string, label: string, outputLog: Stream.Writable) {
     console.log(output);
   }
 }
+
+export function appConfigWithFixture() {
+  window.APP_CONFIG = {
+    walletConnect: {
+      bridge: "https://radicle.bridge.walletconnect.org",
+    },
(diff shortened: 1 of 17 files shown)


tests/support/fixtures.ts
```diff
@@ -139,7 +139,10 @@ export const test = base.extend<{
     await Fs.mkdir(stateDir, { recursive: true });
 
     await use(stateDir);
-    if (process.env.CI && testInfo.status === "passed") {
+    if (
+      process.env.CI &&
+      (testInfo.status === "passed" || testInfo.status === "skipped")
```
sebastinez, line 144: When does this happens? Are we at some point skipping visual tests?
`testInfo.status === "skipped"`

rudolfs: No, this is to eliminate an empty `test.log` being uploaded as an artifact, if we skip a test that's only supposed to run on one browser. See `clipboard.spec.ts`.

sebastinez (approved):
I think it's fine, there are some edge cases that we already discussed, like
- what happens if a build fails (there won't be any new snapshots) = rerunning or manually uploading snapshots
- how to handle intentional visual changes, where visual optional tests are gonna fail.

But I think while the visual tests are optional we can handle these cases

PR #541, opened 2022-12-14
Improvements to the route and codefont types

sebastinez:
I'm currently diving into more advanced Typescript, and I think these changes, can make the codebase leaner

### DIFF
--- a/src/appearance.ts
+++ b/src/appearance.ts
@@ -1,6 +1,6 @@
 import { writable } from "svelte/store";
 
-export type CodeFont = "jetbrains" | "system";
+export type CodeFont = typeof codeFonts[number]["storedName"];
 export const codeFont = writable<CodeFont>(loadCodeFont());
 
 export function storeCodeFont(codeFont: CodeFont): void {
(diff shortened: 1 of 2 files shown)


rudolfs:
I find the original easier to read tbh. Also it doesn't seem to be any leaner:
Screenshot 2022-12-14 at 17 23 22

sebastinez:
> I find the original easier to read tbh. Also it doesn't seem to be any leaner: Screenshot 2022-12-14 at 17 23 22

Hmm ok I agree on the route definitions, but the codeFont types are nicer by having more type inference don't you think? @rudolfs 
I would generally suggest to use in the future `as const` on js objects where it makes sense.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

rudolfs:
> Hmm ok I agree on the route definitions, but the codeFont types are nicer by having more type inference don't you think? @rudolfs I would generally suggest to use in the future `as const` on js objects where it makes sense. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

In terms of type safety there is really no difference. The original version has the same type safety:

Screenshot 2022-12-15 at 09 12 09


What changes is the site of where we define which types are acceptable: in the original version it's at the `CodeFont` definition up at the top, in your suggestion it's at the bottom in the `codeFonts` const.

Screenshot 2022-12-15 at 09 14 10


I think in the end it's a matter of taste. I still prefer the original version, because it's more explicit. I'd argue also that it's easier to read.

sebastinez:
Okay I surrender 🏳️ 😅 
Thanks for your feedback @rudolfs

rudolfs:
Was worth a try. 😅  I'll think more about it, maybe I'll warm up to it later on ; P

PR #547, opened 2022-12-16
Small CI improvements

rudolfs:
CI seems to take longer for the big markdown screenshot diff (usually around 22s).
Ignore warnings about there being no artifacts to upload.


sebastinez (approved):
LGTM

PR #549, opened 2022-12-19
Fix flaky visual tests

rudolfs:
Forgot to mock time in some places. Some tests would run too fast locally and the snapshots would end up showing a loading state, hence the added `{ waitUntil: "networkidle" }`.


sebastinez (approved):
LGTM

Issue #526, opened 2022-12-05
Feature Request - Make vesting contracts more discoverable 


bordumb:
## Overview

Below is an existing vesting contract:

https://app.radicle.xyz/vesting/0xed99ca7e08f89eA2503fF3b59C5B35980FdDe3E3

That is all well and good if we happen to know that the URL is:
```
https://app.radicle.xyz/vesting/{contract_address}
```

## Issue 

If I have a contract ID, but don't know the URL `https://app.radicle.xyz/vesting/{contract_address}`:
I would likely use the Search function at the top left of the site:

<img width="1244" alt="image" src="https://user-images.githubusercontent.com/3743841/205707459-eee57b07-770e-4a51-bba1-cfa61239a37b.png">

This gives me this URL:
https://app.radicle.xyz/0xed99ca7e08f89ea2503ff3b59c5b35980fdde3e3

Let's call this an "Address Profile Page"

This "Address Profile Page" happens to be for a contract.
But it does not (a) have real contract info or (b) link to the vesting page.

## Proposal

It would be good if we had some sort of circular logic that more cohesively links this info to make the vesting information more discoverable.

For example, users' wallets do a pretty nice job of this:

If we go to this URL:
https://app.radicle.xyz/0x641edb587867a703e132a4210d63e93c52aa97ff

It offers the following info:
* Seed
* Address
* Profile

It would be nice if when entering a contract address to the URL, like https://app.radicle.xyz/0xed99ca7e08f89ea2503ff3b59c5b35980fdde3e3, it would provide:
* Contract (Radicle Vesting, Etherscan)

Whereby the pseudo code for displaying Contract (Radicle Vesting, Etherscan) might be like:
```
contract_address = '0xed99ca7e08f89ea2503ff3b59c5b35980fdde3e3'

radicle_vesting_url = f'https://app.radicle.xyz/vesting/{contract_address}'
etherscan_contract_url = f'https://etherscan.io/address/{contract_address}#code

markdown = f'Contract ([Radicle Vesting](radicle_vesting_url), [Etherscan](etherscan_contract_url)'
```

sebastinez:
I like this very much, we could even remove the whole `vesting/` route or redirect it to a "Address Profile Page" where we detect if it's a vesting contract and in that case display the whole vesting information.

sebastinez:
[...]

PR #552, opened 2022-12-21
Improve vesting discovery and error messages

sebastinez:
When loading a ETH address as vesting contract we show a cryptic error
message, which non tech-savy end users don't understand, 8c847b0
changes it to something more human readable.

We allow users to enter and load a contract that probably isn't a
vesting contract, we show this now by showing an error message.

The `getInfo` function in the `vesting` folder is very prone to possible
errors, and benefits from some parallelization between all the method
calls, and some error handling.

We allow hereby to enter a contract address in search and navigate to the
contract profile page where we provide a button to display the detailed vesting
information if available.

On the same profile page, the user has the possibility to withdraw their withdrawable balance
also the balance is being increased on every new block.


sebastinez:
Here a video of some of the improvements:

Show a error message when trying to load a contract that probably isn't a vesting contract
Bildschirm­foto 2022-12-21 um 18 25 23

Show a link to the vesting information on a contract profile page if available
We preload the vestingInfo so the loading time is instant.
Bildschirm­foto 2022-12-21 um 18 27 43

If we try to lookup a vesting contract that doesn't exist with `/vesting/<contract address>` we show a `NotFound` component with some text for the user
Bildschirm­foto 2022-12-21 um 18 29 59

public/typography.css
```diff
@@ -137,6 +137,10 @@ p {
   color: var(--color-foreground-5);
   font-style: italic;
 }
+.txt-error {
```
rudolfs, line 140: We're just using this in a single place, maybe we can just inline it there?

src/App.svelte
```diff
@@ -102,7 +102,7 @@
           session={$session}
           activeRoute={$activeRouteStore} />
       {:else if $activeRouteStore.resource === "vesting"}
-        <Vesting {wallet} session={$session} activeRoute={$activeRouteStore} />
+        <Vesting {wallet} activeRoute={$activeRouteStore} />
```
rudolfs, line 105: Do we even want to keep the special vesting route now that we can just use the search bar?

sebastinez: Yeah I wanted to remove it initially but some core contributors received mails with the /vesting route, to enter/load their contract.
So didn't want to break that.
Maybe at some point in the future?

rudolfs:
Why do we have the vesting toggle button? Could we just show in the badge that it's a vesting contract?

Screenshot 2022-12-22 at 18 02 39

rudolfs:
Gonna tag @brandonhaslegs here, maybe he has some ideas on the design. The interaction feels a bit weird to me, wonder if there's a better way? Also the link to Etherscan could be nicer.

https://user-images.githubusercontent.com/158411/209188684-db0bebf3-c91d-4249-8e0f-be34bbf1d52f.mov



Screenshot 2022-12-22 at 18 04 28

sebastinez:
> Why do we have the vesting toggle button? Could we just show in the badge that it's a vesting contract?

Yeah this was the initial way I thought of doing it, but I didn't liked the amount of lines, showing up, so the idea of the vesting toggle button is to hide all that information behind a toggle.

@brandonhaslegs do you have any idea here maybe on how to make this nicer, I don't think we should show all the vesting information from the get go, since it uses lot of space.

public/typography.css
sebastinez: Yeah I wanted to keep the line of doing it, and eventually with more error handling, we could start using something like this. But for now we can do it inline

src/App.svelte
rudolfs: Or we could just show instructions on that page that say: please paste your vesting contract into the search bar at the top?

src/base/vesting/vesting.ts
```diff
@@ -37,50 +46,88 @@ export async function withdrawVested(
...
+
+  try {
+    const tx: TransactionResponse = await contract
+      .connect(signer)
+      .withdrawVested();
 
-  const tx = await contract.connect(signer).withdrawVested();
+    state.set({ type: "withdrawing", tx });
+    await tx.wait();
+  } catch (e: any) {
```
rudolfs, line 58: We shouldn't cast errors to `any`, by default it's `unknown`. Further down we should check that the error actually has a `.reason` attribute.

src/base/vesting/vesting.ts
```diff
@@ -37,50 +46,88 @@ export async function withdrawVested(
...
+    console.log(e);
```
rudolfs, line 59: `console.warn`?

src/base/vesting/vesting.ts
```diff
@@ -37,50 +46,88 @@ export async function withdrawVested(
...
+): Promise<VestingInfo | undefined> {
+  const contract = getVestingContract(address, wallet);
+
+  let vestingInfo:
+    | [string, string, BigNumber, BigNumber, BigNumber, string, string, string]
+    | undefined = undefined;
+  let token: string | undefined = undefined;
+
+  try {
+    token = (await contract.token()) as string;
```
rudolfs, line 80: Shouldn't we handle the `undefined` case here instead of asserting that it's always a `string`?

Screenshot 2022-12-23 at 09 59 12

rudolfs, line 80: Could we move this out of the try block and just do:
```
token = await contract.token();
if(!token) {
return undefined;
}
```

Instead of the type assertion? I would find that easier to read.

src/base/vesting/vesting.ts
```diff
@@ -37,50 +46,88 @@ export async function withdrawVested(
...
     withdrawableBalance: utils.formatBalance(withdrawable),
     withdrawn: utils.formatBalance(withdrawn),
+    totalVesting: utils.formatBalance(total),
     vestingStartTime,
     vestingPeriod,
     cliffPeriod,
   };
 }
+
+export const parseVestingPeriods = (...timestamps: string[]): string => {
```
rudolfs, line 125:
```suggestion
export const parsedVestingPeriods = (...timestamps: string[]): string => {
```

src/App.svelte
sebastinez: Yeah that doesn't sound like good UX.
IMO either:
- we have a page like currently where we only show vesting information under a specific route.
- or we add it to a profile page like this PR wants to do it, and:
  -  show it in some nice condensed graphical easy to understand way with graphs maybe?
  -  or we hide it with a dropdown to avoid having lots of lines of information.

src/base/vesting/vesting.ts
sebastinez: Hmm the reason I left it inside the `try` block is to catch any errors that could happen when fetching it.
I think I could leave it inside the `try` block but do the `if(!token)` guard to avoid the casting to `string`

sebastinez: Something like
```js
  let token: string | undefined = undefined;

  try {
    token = await contract.token();
    if (!token) {
      return undefined;
    }
```

brandonhaslegs:
This whole page is about vesting right? I don’t think you need to hide anything behind a button. What else is there to do on this page?

sebastinez:
> This whole page is about vesting right? I don’t think you need to hide anything behind a button. What else is there to do on this page?

The page is the profile page of a ethereum address, so there will be a listing of projects corresponding to a user or eventual org, and personal information like seed maybe balances for multi sig and eventually other stuff.

public/typography.css
sebastinez: 634126e

src/base/vesting/vesting.ts
sebastinez: 28746cd should improve it

src/base/vesting/vesting.ts
sebastinez: 28746cd

brandonhaslegs:
Or instead of just showing a button, can you show some value there?

So instead of:

**Vesting** [Show details]

you could indead show a value with a [View more] button next to it

**Vesting Beneficiary** 👤rudolfs.eth [View more]

And the “view more” button would then expand all the vesting info.

brandonhaslegs:
Things I think should change:
- The [back] button is weird. It should just timeout.
- The loading indicator looked better as the three-dot that’s everywhere else in the app
- The eth address after “Waiting for confirmation” should be truncated just like the address is truncated with the “Tokens have been withdrawn” success state. I made a separate issue for truncating all eth addresses here #555

brandonhaslegs:
I am also wondering if this withdrawl flow should be more “generic” like just exist in a modal, just in case you have to withdrawl from a different UI at some point. Then we could repurpose the UI.

sebastinez:
Thanks @brandonhaslegs for the feedback! Really appreciated!

> Or instead of just showing a button, can you show some value there?

That's a great idea, I'll think about some values we can show together with a [View more] toggle

> The [back] button is weird. It should just timeout.

Ok I'll try some other timeout durations, maybe we could also do something like when moving your mouse out of the textarea we timeout after some seconds, kinda like a debounce after user interaction? 

> The loading indicator looked better as the three-dot that’s everywhere else in the app

Got it

> The eth address after “Waiting for confirmation” should be truncated just like the address is truncated with the “Tokens have been withdrawn” success state.

The address shown on "Waiting for confirmation" is the transaction hash of the withdrawal, I thought it would be nice to give users this information in case something goes bad.
The transaction hash is trimmed to it's first 20 characters of 66 in total, I took the inspiration from etherscan but we can do it differently or don't show the tx hash at all 🤷 
Bildschirm­foto 2022-12-23 um 22 40 46

> I am also wondering if this withdrawl flow should be more “generic” (for example just exist in a modal flow) just in case you have to withdrawl from a different UI at some point. Then we could repurpose the withdrawl UI. This would have to be designed of course, but we’ve designed many withdrawl flows at this point, I’m sure we could throw something together. Maybe this is something better to do in the future.

Really like the idea of a generic withdrawal component modal that can be shown at any point when there is some withdraw happening. Maybe that's something we can take into account in the modal refactor at #557

sebastinez:
Okay I addressed your feedback @brandonhaslegs and I'm open to review the etherscan icon https://github.com/radicle-dev/radicle-interface/pull/552#issuecomment-1363110256

I would like to get this PR closed before heartwood comes out.

rudolfs:
> I would like to get this PR closed before heartwood comes out.

You mean you want to merge it into master?

sebastinez:
> You mean you want to merge it into master?

Yeah exactly 👍

brandonhaslegs:
Here is a new etherscan icon

etherscan

brandonhaslegs:
Can you give me a URL where I can test this?

sebastinez:
> @sebastinez Can you give me a URL where I can test this?

Yeah sure, let me update the icon and I'll deploy it somewhere

brandonhaslegs:
Ah sorry I built it just fine from terminal, i just don’t know where to go to find this page

sebastinez:
> Ah sorry I built it just fine from terminal, i just don’t know where to go to find this page

Ok I just added a mocking function to test it more easily.
Pull the latest changes and try out this url http://localhost:3000/0x27BCA0692e13C122E6Fc105b3974B5df7246D464

sebastinez:
Or it can also be previewed here https://tubular-taiyaki-bf2c09.netlify.app/0x27BCA0692e13C122E6Fc105b3974B5df7246D464

rudolfs:
> I am also wondering if this withdrawl flow should be more “generic” (for example just exist in a modal flow) just in case you have to withdrawl from a different UI at some point. Then we could repurpose the withdrawl UI. This would have to be designed of course, but we’ve designed many withdrawl flows at this point, I’m sure we could throw something together. Maybe this is something better to do in the future.

@brandonhaslegs this is how it is implemented now actually and this PR changes it to not be in a modal.

See: https://app.radicle.xyz/vesting

Here are some sample Görli vesting contracts:
```
0x13b2Fc1f601Fb72b86BFAB59090f22bB6E73005A
0xAADcbc69f955523B0ff0A271229961E950538EbE
0x27BCA0692e13C122E6Fc105b3974B5df7246D464
0x9c882463B02221b0558112dec05F60D5B3D99b6a
```


Original:

Screenshot 2022-12-28 at 15 43 07

Screenshot 2022-12-28 at 15 43 43

brandonhaslegs:
Well why dont we keep this with a few changes?

sebastinez:
> Well why dont we keep this with a few changes?

The idea of implementing the whole vesting information into a Profile page came up during a discussion here: https://github.com/radicle-dev/radicle-interface/issues/526#issuecomment-1338144182

I think if we would like to keep everything as is, we could just redirect searches for vesting contracts to the modal view.
And improve the withdraw process, and error handling in the vesting modal. 🤔

sebastinez:
Ok I think this PR is still better than the current status quo.. but if anyone wants to take another stab at designing this or has already something, we can sync maybe over Discord and bring this PR over the finish line.

cloudhead:
Feedback here:
image

cloudhead:
Furthermore:

*withdraw* button should not show if there's nothing to withdraw or if you're not the beneficiary.
If you're not connected, it should be disabled.

sebastinez:
Ok addressed the feedback by @cloudhead.

Here
Bildschirm­foto 2023-01-09 um 15 37 33

And here some videos of the withdrawal flow.
One receives an error during the transaction wait, and one completes sucessfully.

https://user-images.githubusercontent.com/7912302/211340818-f4d60593-7c62-41ec-9829-52be996a869c.mov

https://user-images.githubusercontent.com/7912302/211340828-392d434c-88f8-4541-825f-268085471280.mov

rudolfs:
The etherscan icon doesn't match the rest of the social-media link icons:
Screenshot 2023-01-09 at 16 30 28

It also doesn't seem to be centered in the bounding box.
Screenshot 2023-01-09 at 16 30 55

Here are all our other icons:
Screenshot 2023-01-09 at 16 37 50

rudolfs:
It's possible to submit the input via the <kbd>enter</kbd> key even when the button is disabled:
Screenshot 2023-01-09 at 16 42 43

We should not allow submit on keypress when the form isn't valid.

When there is no vesting contract, there are no errors in the console. We should adjust the error message to just "No vesting account found under this address."
Screenshot 2023-01-09 at 16 45 08

rudolfs:
Is there a reason why this address is clickable? We are already on the profile page of it, no?
Screenshot 2023-01-09 at 16 49 37

cloudhead:
Looking better! A couple more things:

* usually we show errors using the error modal which is all red; we don't mix red error text in blue modals
* In the success case, we should show the withdrawal address inline, eg. "tokens withdrawn to [icon] xyz - abc"

rudolfs:
We should make the gap between the buttons a bit wider, like 1rem?
Screenshot 2023-01-09 at 16 54 10

rudolfs:
When funds have been withdrawn and I click on the address in the modal, I get an error:

Screenshot 2023-01-09 at 16 56 43

Screenshot 2023-01-09 at 16 56 46

cloudhead:
> Is there a reason why this address is clickable? We are already on the profile page of it, no?

I think that's a separate issue no? ie. it affects all profiles.

cloudhead:
The etherscan icon is just bigger than the other icons for some reason.

brandonhaslegs:
Yes! I recently redesigned our icon set (refined them, made them more streamlined). Here are the Github and Twitter ones. And the 🔗  one too!
github
link
twitter

sebastinez:
Ok did update the Icons.

Bildschirm­foto 2023-01-09 um 22 07 17

Now etherscan surely seems more like the others.
@brandonhaslegs if you have other icons, for more streamlining please lmk and I'll do that in another PR
image
I think there are some others that could be improved

sebastinez:
> It's possible to submit the input via the enter key even when the button is disabled: Screenshot 2023-01-09 at 16 42 43

Good catch! 💪

sebastinez:
Ok addressed with the latest commit:

- the `ErrorModal` usage
- the withdrawal not working 😓 thanks for the catch @rudolfs!
- the error when clicking on the address where to tokens where withdrawn to.
- the ability to click enter and to trigger the `on:submit` of a `TextInput` component even if the input isn't valid.
- modified the error message, removing the dev console part.

A video as reference
https://user-images.githubusercontent.com/7912302/211507254-af37d79c-5c55-49b9-8a8c-9a1a88051e3f.mov

cloudhead:
Thanks, looking good, just one thing left: spacing/alignment here is off:
image

I also wonder, what happens if you close the modal while it's confirming?

sebastinez:
> Thanks, looking good, just one thing left: spacing/alignment here is off: image
> 
> I also wonder, what happens if you close the modal while it's confirming?

If you close the modal while confirming, after having signed the tx, the tx goes out either way and you loose eventually feedback if something goes wrong.
We could eventually just show the `Close` button once the funds have been withdrawn 🤔 but I wanted to give users the ability to close the modal also during the tx.

sebastinez:
> Thanks, looking good, just one thing left: spacing/alignment here is off

@cloudhead Ok I fixed this by removing the `Address` component here, and do it as we do in the registration flow, where we just show the formatted address.

Bildschirm­foto 2023-01-10 um 10 34 25

rudolfs:
The `withdrawn` amount doesn't update automatically after the tx is confirmed. It only updates after a page refresh.

src/lib/utils.ts
```diff
@@ -290,10 +297,11 @@ export function isFulfilled<T>(
 
 // Get the explorer link of an address, eg. Etherscan.
 export function explorerLink(addr: string, wallet: Wallet): string {
```
rudolfs, line 299: We should rename the `addr` attribute to `addrOrTx`.

src/lib/utils.ts
```diff
@@ -310,6 +318,19 @@ export function parseRadicleId(id: string): string {
   }
 }
 
+// Checks that the input is an object with a `reason` property.
+// Is used to check if ether.js errors have a `reason` property.
+export function hasReasonProperty(
```
rudolfs, line 323: Were using this only in one place, could we inline it for now?

sebastinez: 1e4ebf0

src/lib/utils.ts
sebastinez: 1e4ebf0

sebastinez:
> The `withdrawn` amount doesn't update automatically after the tx is confirmed. It only updates after a page refresh.

1e4ebf0

src/lib/vesting.ts
```diff
@@ -37,50 +47,113 @@ export async function withdrawVested(
...
+        contract.beneficiary(),
+        contract.withdrawableBalance(),
+        contract.withdrawn(),
+        contract.totalVestingAmount(),
+        contract.vestingStartTime(),
+        contract.vestingPeriod(),
+        contract.cliffPeriod(),
+      ]);
+    } catch (e) {
+      return undefined;
```
rudolfs, line 108: Can we log this error with `console.warn`?

src/lib/vesting.ts
```diff
@@ -37,50 +47,113 @@ export async function withdrawVested(
...
+      vestingStartTime,
+      vestingPeriod,
+      cliffPeriod,
+    };
+  },
+  address => address,
+  { max: 1000 },
+);
+
+export const parseVestingPeriods = (...timestamps: string[]): string => {
```
rudolfs, line 138: Could we use a normal function definition instead of const where possible?

sebastinez: 16ead15

src/lib/vesting.ts
sebastinez: 16ead15

src/views/vesting/Withdraw.svelte
```diff
@@ -0,0 +1,72 @@
...
+    on:close={() => dispatch("close")} />
+{:else}
+  <Modal on:close floating>
+    <span slot="title">
+      {utils.formatAddress(contractAddress)}
+    </span>
+
+    <span slot="subtitle">
+      {#if $state.type === "withdrawingSign"}
+        <span class="txt-missing">Waiting for a signature...</span>
```
rudolfs, line 48: Could we use the ellipsis character … instead of `...` everywhere?

sebastinez: 5d2cbd3

rudolfs (approved):
👏 🚀

PR #554, opened 2022-12-22
Organise files more consistently

rudolfs:
We move all reusable components into `/src/components`, all typescript modules that are used globally into `/src/lib`. Rename `/src/base` -> `/src/views`. And clean up imports.

Here's what the proposed top-level structure now looks like:
```
.
├── App
├── App.svelte
├── components
├── config.json
├── global.d.ts
├── index.ts
├── lib
└── views
```

While shuffling things around I had the following thoughts:

- We should pay attention to the import graph, things that are used in multiple places should be in `lib`. Modules in `lib` should not pull in dependencies from outside (i.e. `components` or `views`), if that's the case, then that piece of code should be moved to `lib`.

- I left `App.svelte` at the top level, because it's the entry point (along with `index.ts`) and the place where top-level routing happens, all other app screens live in `views`.

- I keep components that are single-use close to where they are used, e.g. things that are only used by `App.svelte` go into the `App` subdirectory, and components that belong to `Header.svelte` go into the `Header` subdirectory:
   ```
  .
  ├── ColorPalette.svelte
  ├── Header
  │   ├── Logo.svelte
  │   ├── Search.svelte
  │   ├── SearchResults.svelte
  │   ├── SettingsDropdown.svelte
  │   └── ThemeToggle.svelte
  └── Header.svelte
   ```
  I applied the same rule in all other places in the file tree where components are broken down into smaller sub-parts.

- All reusable components go into `components`. Here we should strive to build components that are truly composable and reusable, that is they should only have logic that doesn't depend on specific data fetching or components outside of this directory. Components that are specific to only a single view should live within the directory of that view.

- We shouldn't cross import things within `views`, if there's logic that can be reused, it should be extracted either to `lib` or `components` and then imported from there.


src/App.svelte
```diff
@@ -1,22 +1,25 @@
...
-  import Vesting from "@app/base/vesting/Routes.svelte";
+  import { Connection, state, session } from "@app/lib/session";
+  import { getWallet } from "@app/lib/wallet";
+  import { initialize, activeRouteStore } from "@app/lib/router";
+  import { twemoji, unreachable } from "@app/lib/utils";
+
+  import ColorPalette from "./App/ColorPalette.svelte";
+  import Header from "./App/Header.svelte";
+
+  import Loading from "@app/components/Loading.svelte";
```
sebastinez, line 12: It seems to me that we maybe should add a new path alias? Something like `@components/Address.svelte` ?
Since it seems we'll be writing a lot `@app/components`

sebastinez: Maybe the same with `@app/views`?

src/App/Header/Search.svelte
```diff
@@ -1,13 +1,13 @@
...
 
   import { ethers } from "ethers";
 
-  import * as utils from "@app/utils";
-  import { Profile } from "@app/profile";
-  import { Project } from "@app/project";
-  import { config } from "@app/config";
+  import * as utils from "@app/lib/utils";
+  import { Profile } from "@app/lib/profile";
+  import { Project } from "@app/lib/project";
```
sebastinez, line 9: Like the `lib` folder 👍

sebastinez (commented):
Looks great to me, some minor nits

src/App.svelte
rudolfs: I actually did that initially (`@views`, `@lib` and `@components` ), but then reverted back to this. My reasoning was that maybe we'll still want to access files from the `@app` root and if we do that, then having aliases for the subdirectories is a bit over-kill. 🤷

sebastinez: ok agreed!

sebastinez (approved):
Nice! LGTM

cloudhead:
LGTM

PR #559, opened 2022-12-26
Adds syntax highlighting to file diff

sebastinez:
Adds syntax highlighting to file diffs.
Still a WIP, since the highlighting breaks down when context is missing, in svelte files e.g.


sebastinez:
Will close this PR, and reimplement it once we have a better file diff structure

PR #560, opened 2022-12-26
Weekly dependencies update

sebastinez:
by santa claus 😉 

Also updates the `package-lock.json` `lockVersion` to 3
I think we can safely assume that since we support node v18, we can drop npm v6 and lower

Also removes `@rsksmart/web3-mock-provider` since we don't use it and it still uses some deprecated packages.


src/views/projects/CloneButton.svelte
```diff
@@ -92,7 +92,7 @@
   <svelte:fragment slot="modal">
     <div class="dropdown">
       <div class="clone-url-wrapper">
-        <div class="clone-url" name="rad-clone-url">{radCloneUrl}</div>
+        <div class="clone-url" id="rad-clone-url">{radCloneUrl}</div>
```
rudolfs, line 95: Did svelte-check complain about this? Are we even using this for anything?

rudolfs (commented):
Could you also please update the transitive deps with:
```
npm cache clean --force
rm -rf node_modules
rm package-lock.json
npm install
```

src/views/projects/CloneButton.svelte
sebastinez: Yeah `svelte-check` started complaining
And we use it for the `label` tag a bit further down

PR #515, opened 2022-11-27
Migrate to the heartwood

xphoniex:

PR #564, opened 2022-12-28
Rename Seed class attributes

rudolfs:
```
seed.api -> seed.addr
seed.link -> seed.node
```

### DIFF
--- a/src/lib/seed.ts
+++ b/src/lib/seed.ts
@@ -23,16 +23,16 @@ export class InvalidSeed {
   }
 }
 
-export const defaultHttpApiPort = 8777;
-export const defaultLinkPort = 8776;
+export const defaultSeedPort = 8777;
+export const defaultNodePort = 8776;
 export const defaultGitPort = 443;
 
 export class Seed {
   valid = true as const;
 
-  api: { host: string; port: number | null };
+  addr: { host: string; port: number | null };
   git: { host: string; port: number | null };
-  link: { host: string; id: string; port: number };
+  node: { host: string; id: string; port: number };
 
   version?: string;
   emoji: string;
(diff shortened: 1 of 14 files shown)


src/lib/seed.ts
```diff
@@ -23,16 +23,16 @@ export class InvalidSeed {
   }
 }
 
-export const defaultHttpApiPort = 8777;
-export const defaultLinkPort = 8776;
+export const defaultSeedPort = 8080;
+export const defaultNodePort = 9050;
```
xphoniex, line 27: please revert `defaultNodePort` to 8776, this was mistakenly changed to 9050

src/lib/seed.ts
```diff
@@ -41,85 +41,85 @@ export class Seed {
...
     assert(/^[a-z0-9]+$/.test(seed.id), `invalid seed id ${seed.id}`);
 
-    let api = null;
-    let git = null;
-    let apiPort: number | null = defaultHttpApiPort;
-    let gitPort: number | null = defaultGitPort;
+    let _seed = null;
+    let _git = null;
+    let _seedPort: number | null = defaultSeedPort;
+    let _gitPort: number | null = defaultGitPort;
```
xphoniex, line 53: maybe leave a `NOTE` that `git` stuff should be merged into `seed`?

cloudhead:
So the default `seed` port, which is now both for Git and API should be `443` in production, which is the default HTTPS port. When developing locally, I guess we can add the port in the URL, eg. `/seeds/127.0.0.1:8080/`?

src/lib/seed.ts
rudolfs: I'll do that at the very end in #515.

rudolfs:
> So the default `seed` port, which is now both for Git and API should be `443` in production, which is the default HTTPS port. When developing locally, I guess we can add the port in the URL, eg. `/seeds/127.0.0.1:8080/`?

Actually this PR should be just a simple rename, nothing should change. I mixed up the port numbers, which I fixed now.

src/lib/seed.ts
```diff
@@ -131,10 +131,10 @@ export class Seed {
     projects: { count: number };
     users: { count: number };
   }> {
-    return new Request("/stats", this.api).get();
+    return new Request("/stats", this.addr).get();
   }
 
-  static async getPeer(host: Host): Promise<{ id: string }> {
+  static async getNode(host: Host): Promise<{ id: string }> {
```
xphoniex, line 137: why not create both `getNode` and `getPeer`, also with comments (that one is for heartwood, etc)?

If you want to keep the heartwood fn signature but call `/peer`, how are you planning to eventually resolve this?

rudolfs: That's for a follow up. I'd like to first just extract the renames out of #515.

xphoniex: hm... I still suggest a comment explaining the discrepancy, this might be helpful for your future self, but if you're certain then this lgtm as well.

xphoniex (approved):

PR #567, opened 2022-12-29
Rename issue/patch APIs

rudolfs:
To reduce the scope of #515.

The legacy seeds don't have any working issues, so I didn't bother to write any translation.
This fixes patch browsing, there was a regression.

Test it against: http://127.0.0.1:3000/seeds/clients.radicle.xyz/rad:git:hnrkq5sti58yw41szighu565oprgt5eduhmhy/patches/hnrkkktb1dyt9gkuufzuax1uzdoueiskxzrto

### DIFF
--- a/src/lib/issue.ts
+++ b/src/lib/issue.ts
@@ -14,7 +14,7 @@ export interface IIssue {
   state: State;
   comment: Comment;
   discussion: Thread[];
-  labels: Label[];
+  tags: Tag[];
   timestamp: number;
 }
 
@@ -37,7 +37,7 @@ export interface Comment<R = null> {
 
 export type Thread = Comment<Comment[]>;
 
-export type Label = string;
+export type Tag = string;
 
 export function groupIssues(issues: Issue[]): {
   open: Issue[];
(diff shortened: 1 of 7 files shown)


src/lib/issue.ts
```diff
@@ -69,14 +69,14 @@ export class Issue {
...
-    this.labels = issue.labels;
+    this.tags = issue.tags;
     this.timestamp = issue.timestamp;
   }
 
   // Counts the amount of comments and replies in a discussion
   countComments(): number {
     return this.discussion.reduce((acc, comment) => {
-      if (comment.replies) return acc + comment.replies.length + 1; // We add all replies and 1 forathe comment in this loop.
+      if (comment.replyTo) return acc + comment.replyTo.length + 1; // We add all replies and 1 forathe comment in this loop.
```
xphoniex, line 79: IIRC, `replyTo` is not an array anymore in the new model, but an index. Maybe keep it as `.replies` ?

rudolfs: 4067a2c03c830c7f7263c70bc2a434df083134ab

xphoniex (approved):

PR #571, opened 2022-12-31
Fix a couple of markdown styles

cloudhead:
It looks like there was maybe a regression at some point (I guess [here](https://github.com/radicle-dev/radicle-interface/commit/569b6601341af8cd9fb31232bf6bde4ecd1c8a25)) and we lost the colors on the code and blockquote blocks. This adds them back. We also give blockquotes some padding like regular paragraphs.

Edit: how can I fix the visual test failure?

Signed-off-by: Alexis Sellier <self@cloudhead.io>

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -253,6 +253,12 @@
     padding-left: 1.25rem;
     margin-bottom: 1rem;
   }
+  .markdown :global(li > ul) {
+    margin-bottom: 0rem;
+  }
+  .markdown :global(li > ol) {
+    margin-bottom: 0rem;
+  }
   .markdown :global(table) {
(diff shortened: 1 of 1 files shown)


rudolfs:
> Edit: how can I fix the visual test failure?

We diff the visuals against `master`. So if the changes in the visual diffs are expected (like they are here), we just ignore the failing test. Once the PR is merged into master the cache against which we diff will be updated and subsequent PRs will be compared against the new one.

rudolfs (approved):

rudolfs:
> Ah interesting. Is there no way to avoid the failing CI for that commit though? Quite unexpected.

Not at the moment. I looked into it but any other solution would require us to host the snapshots somewhere. Was too much effort to bother.

PR #574, opened 2023-01-04
Split up visual markdown specs into sections

rudolfs:
Before we had a single very long screenshot, that was hard to compare. Now we have a screenshot per section, which should make it easier to spot differences.

For example:

markdown-rendering-blockquotes-1-actual

markdown-rendering-emphasis-1-actual

markdown-rendering-footnotes-1-actual

### DIFF
--- a/.github/workflows/check-e2e.yml
+++ b/.github/workflows/check-e2e.yml
@@ -76,6 +76,14 @@ jobs:
             npm run test:e2e -- --project ${{ matrix.browser }};
           fi
 
+      # We don't care about the logs in visual snapshot tests, only the images.
+      - name: Cleanup artifacts
+        if: always()
+        run: |
+          if [ ${{ matrix.browser }} = "visual" ]; then
+              shopt -s globstar
+              rm -rf tests/visual/snapshots/**/*.log
+
       - name: Upload artifacts
         uses: actions/upload-artifact@v3
         if: always()
--- a/.github/workflows/check-e2e.yml
+++ b/.github/workflows/check-e2e.yml
@@ -76,13 +76,21 @@ jobs:
             npm run test:e2e -- --project ${{ matrix.browser }};
           fi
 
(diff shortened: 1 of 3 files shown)


sebastinez (commented):
I think all this makes sense and is much nicer to review, but I would still keep one full page visual diff, since I think there could be spacing issues between the end of one section and the start of the next one that maybe go unnoticed..
I also don't like the static viewport height numbers, but since we won't be editing this markdown cheatsheet I suppose it's safe.

rudolfs:
> I think all this makes sense and is much nicer to review, but I would still keep one full page visual diff, since I think there could be spacing issues between the end of one section and the start of the next one that maybe go unnoticed.. I also don't like the static viewport height numbers, but since we won't be editing this markdown cheatsheet I suppose it's safe.

Hm, fair points. The two issues I tried to solve was:
  1) when fullscreen screenshots are enabled, subtle spacing changes affect the total height of the screenshot and when the expected and actual screenshot sizes don't match, playwright doesn't generate a diff image
  2) for some reason the screenshots on CI don't fully render all content, some stuff at the bottom is always missing
  3) it's very hard to compare large snapshots without special tooling, with smaller screenshots it's easy to just flip back and forth, but with a huge one I can't see the details

I think 1 could be solved by just setting a static viewport size that captures all of the markdown file, then the height won't change and we'll always get a diff. 2 could be solved by doing an `expect` with some content from the bottom of the page and if that doesn't work with some `sleep`. And maybe 3 would be already solved by 1 if we get the diff always.

I'll try to implement this another way and we can compare.

rudolfs:
Hm, there's some bug that prevents making larger screenshots. Even when I add a lot of `sleep`s the screenshots have weird gaps in content:

Screenshot 2023-01-05 at 10 23 53

sebastinez (approved):
Lets see how this PR improves the visual diffing, if there are issues with spacing we can revisit it.

PR #576, opened 2023-01-05
Add footnotes support to markdown

### LINKED ISSUE #575: Markdown footnotes not rendered correctly
It looks like our markdown renderer is not correctly rendering footnotes. See here:
image

The markdown text for the above is `Ed25519[^1]`

**The correct rendering is:**

image

A footer is also supposed to be rendered:
image

You can see how github renders it here: https://github.com/radicle-dev/rips/blob/1c402116983be19e754fb14aa7ce38145f0a4b09/0002-identity.md
Versus radicle: https://app.radicle.xyz/seeds/clients.radicle.xyz/rad:git:hnrkx7ruakgb9ozzi7atwb1rh6moq8pjonp6o/tree/1c402116983be19e754fb14aa7ce38145f0a4b09/0002-identity.md

I'm guessing there must be some kind of extension we can enable..

sebastinez:
This PR adds support for footnotes in markdown.

It does so by adding two more markdown extensions:
- One for the inline footnote reference.
- And the other for the footnotes at the end of the document.

~~There is still some work left, I would like to wrap the footnotes at the end of the document inside some `<section>` to get a specific styling applied.
Also there are some edge cases I would like to fix.~~
This can be revisited in a different PR if needed, for now this solves the issue at hand.

This PR will eventually close #575 

### Pictures for reference:
Bildschirm­foto 2023-01-05 um 11 22 44
Bildschirm­foto 2023-01-13 um 14 09 06

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -8,15 +8,12 @@
   import { toDom } from "hast-util-to-dom";
 
   import { base } from "@app/lib/router";
+  import { getImageMime, isUrl, twemoji, scrollIntoView } from "@app/lib/utils";
   import { highlight } from "@app/lib/syntax";
   import {
     markdownExtensions as extensions,
     renderer,
-    getImageMime,
-    isUrl,
-    twemoji,
-    scrollIntoView,
-  } from "@app/lib/utils";
+  } from "@app/lib/markdown";
 
   export let content: string;
   export let doc = matter(content);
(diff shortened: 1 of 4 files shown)


sebastinez:
@cloudhead It seems that GH does use the classic blue font color for the footnote references and also the footnote links, AFAIC see we don't do that with the other links in our markdown documents, so we probably don't want to start do it for footnotes, wdyt?

cloudhead:
The blue arrow is a reverse link, but we should be using `↩` instead of the icon.

rudolfs (approved):

PR #580, opened 2023-01-05
Fix flaky visual landing page spec

rudolfs:
Current working theory: the dynamic import made the timer stubs load in the wrong order sometimes.
Not a 100% sure tho. 🤷 

This also removes the playwright deps install steps from the jobs. Apparently the tests work just fine without them. Saves us some build time.


sebastinez (approved):
Nice! Good work debbuging this!

PR #582, opened 2023-01-16
Add httpd session route

sebastinez:
This PR adds a session route with the following spec
```
/session/<sessionId>&sig=<Signature>&pk=<PublicKey>
```
This route gets shown in the cli when `rad web` is called to authenticate the web client with the httpd.

Once the web client lands on this route it sends the session id, signature and public key to the httpd to authenticate the session, and then stores the authenticated session id in the sessionStorage to keep persistance between page reloads.

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -10,6 +10,7 @@
 
   import NotFound from "@app/components/NotFound.svelte";
   import Home from "@app/views/home/Index.svelte";
+  import Session from "@app/views/session/Index.svelte";
   import Projects from "@app/views/projects/View.svelte";
   import Seeds from "@app/views/seeds/Routes.svelte";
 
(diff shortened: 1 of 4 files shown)


src/App.svelte
```diff
@@ -95,6 +96,8 @@
     <div class="wrapper">
       {#if $activeRouteStore.resource === "home"}
         <Home />
+      {:else if $activeRouteStore.resource === "login"}
```
cloudhead, line 99: "login" isn't a resource, but an action. What you want here is "session" (and in all other places)

src/lib/router.ts
```diff
@@ -241,6 +241,20 @@ function pathToRoute(path: string): Route | null {
...
     }
+    case "session": {
+      const id = segments.shift();
+      if (id) {
+        return {
+          resource: "session",
+          params: {
+            id,
+            sig: url.searchParams.get("sig") ?? "",
+            pk: url.searchParams.get("pk") ?? "",
```
rudolfs, line 252: Is `pk` a public key? If so we should probably use `publicKey` as the variable and query param.

sebastinez: Hey yeah I shortened it to avoid such a long querystring, but since the user won't type it by hand.
I have no issue in writing it out.

sebastinez: Ok this should be fixed with 3d9e644aaf039c6630e8b0cd54c052708746678b

sebastinez: I'm keeping the serialized querystring with pk and sig, since `radicle-httpd` is expecting that, we can eventually change that, but the PR on the heartwood repo just got merged.

rudolfs (approved):

Issue #584, opened 2023-01-18
`EnsProfile` is being created with the wrong params


sebastinez:
> but also, on updating profiles using app.radicle.xyz there is a visual glitch where after submitting updated profile, changes and transactions are ok, but visually, updates are not applied (but on other sites they are like drip) so this made me re-attempt publishing profile updates a few times losing gas each time

Source https://discord.com/channels/841318878125490186/841318878650302490/1065184988112363550

okanisis:
Here's a screenshot showing the site after making updates to profile. The only empty boxes should be the radicle related ones at the bottom for URN, Device ID. Haven't filled those out yet but did run `rad auth` to get them atm.

It's also odd that the Twitter username I have is now sitting in the Github username box.

![Profile Not Filled](https://user-images.githubusercontent.com/1724441/213270520-4b8bf4d6-51e7-4674-8489-2fc671ddd9cf.png)

### System Specs:
**OS:**
```sh
% uname -a
Linux cryptsus 6.1.6-arch1-3 #1 SMP PREEMPT_DYNAMIC Mon, 16 Jan 2023 12:51:23 +0000 x86_64 GNU/Linux
```

**Browser (all shields and script blocking disabled):**
```sh
% brave://version

Brave | 1.47.171 Chromium: 109.0.5414.87 (Official Build) (64-bit)
Revision | 2dc18eb511c56e012081b4abc9e38c81c885f7d4-refs/branch-heads/5414@{#1241}
OS | Linux
JavaScript | V8 10.9.194.9
User agent | Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/109.0.0.0 Safari/537.36
```

**Network/Wallet:**
```sh
Wallet: 0xC42eF624f54BA8A5191480530295308E76eB5C9f
Network: Mainnet and Goerli
Metamask Browser Extension: 10.23.3
```

okanisis:
On Goerli, I registered before trying mainnet but the name is unavailable to me looks like:
![image](https://user-images.githubusercontent.com/1724441/213272803-49614bee-a3d8-4789-86b6-7b15959abea1.png)

[Etherscan Tx - 0x6b4cc2e5198d2e0075db081581d5c1accc075c89517892be4388e7098fc83650](https://goerli.etherscan.io/tx/0x6b4cc2e5198d2e0075db081581d5c1accc075c89517892be4388e7098fc83650)

[...]

PR #587, opened 2023-01-18
Add `seedGit` to `getRegistration` destructuring

sebastinez:
We were destructuring an array of promises into individual values, but we forgot to add `seedGit` as a new value, this gave issues for the creation of a new `EnsProfile`.

This PR originated by #584 

## Before
Bildschirm­foto 2023-01-18 um 20 58 12

## After
Bildschirm­foto 2023-01-18 um 20 58 02

### DIFF
--- a/src/lib/registrar.ts
+++ b/src/lib/registrar.ts
@@ -90,8 +90,18 @@ export async function getRegistration(
     getText(resolver, "com.github"),
   ]);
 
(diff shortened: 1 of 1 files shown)


sebastinez:
This is probably a place where we would need a visual test, once we have proper ether provider mock up for testing

rudolfs (approved):
I think this regression came from the Heartwood PR.

PR #588, opened 2023-01-19
Conform to new JSON schema

xphoniex:


src/lib/commit.ts
```diff
@@ -81,24 +81,30 @@ export function formatGroupTime(timestamp: number): string {
 }
 
 export function groupCommits(
-  commits: { header: CommitHeader; context: CommitContext }[],
+  commits: { commit: CommitHeader; context: CommitContext }[],
 ): CommitGroup[] {
   const groupedCommits: CommitGroup[] = [];
   let groupDate: Date | undefined = undefined;
 
+  if (!window.HEARTWOOD) {
```
rudolfs, line 89: We should do all of these transforms closer to where the API request is made, like we did in the first PR.
So in `src/lib/project.ts` the `getCommits` and `getCommit` functions. So that they're all in one place.

src/lib/project.ts
```diff
@@ -282,7 +282,24 @@ export class Project implements ProjectInfo {
       page: opts?.page,
       verified: opts?.verified,
     };
-    return new Request(`projects/${id}/commits`, host).get(params);
+    const result = await new Request(`projects/${id}/commits`, host).get(
+      params,
+    );
+    if (!window.HEARTWOOD) {
+      for (const commit of result.headers) {
+        commit.commit = (commit as any).header;
```
rudolfs, line 290: We don't need the `as any` here.

```diff
diff --git a/src/lib/project.ts b/src/lib/project.ts
index 447a7f7d..306bec5c 100644
--- a/src/lib/project.ts
+++ b/src/lib/project.ts
@@ -287,10 +287,10 @@ export class Project implements ProjectInfo {
     );
     if (!window.HEARTWOOD) {
       for (const commit of result.headers) {
-        commit.commit = (commit as any).header;
+        commit.commit = commit.header;
         delete commit.header;

-        commit.commit.committer.time = (commit.commit as any).committerTime;
+        commit.commit.committer.time = commit.commit.committerTime;
         delete commit.commit.committerTime;

         commit.commit.id = commit.commit.sha1;
@@ -316,10 +316,10 @@ export class Project implements ProjectInfo {
     ).get();

     if (!window.HEARTWOOD) {
-      result.commit = (result as any).header;
+      result.commit = result.header;
       delete result.header;

-      result.commit.committer.time = (result.commit as any).committerTime;
+      result.commit.committer.time = result.commit.committerTime;
       delete result.commit.committerTime;

       result.commit.id = result.commit.sha1;
```

rudolfs (approved):
🚀

PR #589, opened 2023-01-24
Fix heartwood fixture

xphoniex:


xphoniex:
Pending https://github.com/radicle-dev/heartwood/pull/257 merge.

scripts/create-seed-fixture-heartwood
```diff
@@ -1,24 +1,24 @@
...
 set -euo pipefail
 
-PASSPHRASE=asdf
-SEED="0.0.0.0:8080"
+export RAD_PASSPHRASE=asdf
 
 REPO_ROOT=$(git rev-parse --show-toplevel)
-ID=$(echo $RANDOM | md5sum | head -c 8)
-BASE_PATH=$REPO_ROOT/tests/tmp/create-seed-fixture-$ID
+ID=$(echo $RANDOM | md5sum | head -c 4)
```
xphoniex, line 10: There's a limit to SUN_LEN (len of path to socket) so I made heartwood path shorter.

scripts/create-seed-fixture-heartwood
```diff
@@ -27,25 +27,68 @@ mkdir -p $TEST_REPO_PATH
...
+
+### BOB ###
 
-RAD_HOME=$ALICE_RAD_HOME RAD_PASSPHRASE=$PASSPHRASE rad init --name "source-browsing" --description "Git repository for source browsing tests" --default-branch "main" --no-confirm
-PROJECT_ID=$(RAD_HOME=$ALICE_RAD_HOME RAD_PASSPHRASE=$PASSPHRASE rad .)
+export GIT_AUTHOR_NAME="Bob Belcher"
+export GIT_AUTHOR_EMAIL="bob@radicle.xyz"
+export GIT_COMMITTER_NAME="Bob Belcher"
+export GIT_COMMITTER_EMAIL="bob@radicle.xyz"
+export GIT_COMMITTER_DATE="Mon Nov 21 14:00 2022 +0100"
```
xphoniex, line 68: Bob's commit doesn't match with non-heartwood version and the culprit I supposed is `GIT_COMMITTER_DATE`.

scripts/create-seed-fixture-heartwood
```diff
@@ -27,25 +27,68 @@ mkdir -p $TEST_REPO_PATH
...
+
+cd $BOB_CHECKOUT
+rad clone rad:zKtT7DmF9H34KkvcKj9PHW19WzjT
+
+killall radicle-node
+
+cd $TEST_REPO_NAME
+echo "Updated readme" > README.md
+git add README.md
+git commit --message "Update readme" --date "Mon Nov 21 14:00 2022 +0100"
```
xphoniex, line 90: Note that both `AuthorDate` and `CommitDate` need to be stable.

tests/e2e/project.spec.ts
```diff
@@ -322,31 +326,45 @@ test("peer and branch switching", async ({ page }) => {
...
-        "main 2b32f6f",
-      );
-      await expectCounts({ commits: 9, contributors: 2 }, page);
-      await expect(page.locator("text=2b32f6f Update readme")).toBeVisible();
+      if (process.env.HEARTWOOD) {
+        await expect(page.getByTitle("Current branch")).toContainText(
+          "main 2febee7",
+        );
+        await expectCounts({ commits: 9, contributors: 2 }, page);
+        await expect(page.locator("text=2febee7 Update readme")).toBeVisible();
```
xphoniex, line 347: ^

tests/e2e/project/commits.spec.ts
```diff
@@ -87,32 +103,30 @@ test("peer and branch switching", async ({ page }) => {
...
   if (process.env.HEARTWOOD) {
-    test.skip();
+    await page.locator(`text=${bobRemote}`).click();
+    await expect(page.getByTitle("Change peer")).toHaveText(
+      `${bobRemote.substring(0, 6)}…${bobRemote.slice(-6)}`,
+    );
+  } else {
+    await page.locator("text=bob hyyzz9").click();
+    await expect(page.getByTitle("Change peer")).toHaveText("bob");
+    // not applicable to heartwood?
```
xphoniex, line 118: We don't seem to have a `Verified` badge in heartwood. Thoughts on this @cloudhead ?

xphoniex:
also
bobbelcher

scripts/create-seed-fixture-heartwood
```diff
@@ -27,25 +27,68 @@ mkdir -p $TEST_REPO_PATH
...
+
+cd $TEST_REPO_NAME
+echo "Updated readme" > README.md
+git add README.md
+git commit --message "Update readme" --date "$GIT_COMMITTER_DATE"
+git push rad
 
 cd $BASE_PATH
-tar -cjf palm.tar.bz2 --exclude "post-receive" --exclude "pre-receive" -C $ALICE_RAD_HOME .
+tar -cjf palm.tar.bz2 --exclude "post-receive" --exclude "pre-receive" -C $BOB_RAD_HOME .
```
rudolfs, line 94: Why are we creating the fixture from Bob's workdir and not using a separate seed node?

xphoniex: Bob has everything needed. Why sync with a 3rd node?

rudolfs: To check that it works.

rudolfs: Seed nodes are a critical part of the architecture and previously we had a lot of problems that only showed up when the data went through one, so it is important to test this scenario.

scripts/create-seed-fixture-heartwood
rudolfs: Is this necessary when we already have an `if` in the specs?

scripts/create-seed-fixture-heartwood
xphoniex: sync and clone is already going through the seed nodes (called `radicle-node`), see [here](https://github.com/radicle-dev/radicle-interface/pull/589/files#diff-f06534a87c78edc1de590220fffc6cdd05edf7b5adef0d4b0f0e3c23ba7f4229R43) and [here](https://github.com/radicle-dev/radicle-interface/pull/589/files#diff-f06534a87c78edc1de590220fffc6cdd05edf7b5adef0d4b0f0e3c23ba7f4229R78). 

a more complex integration test doesn't really belong here in `radicle-interface` and lies on the `heartwood` itself.

rudolfs: I know that we're testing it elsewhere as well. But I would appreciate if we could do it here as well, because as I said we had problems in the past with data being mangled that only showed up in the interface. Would it be a lot of work to add it?

rudolfs:
> also bobbelcher

Did you figure out why the sorting is off?

xphoniex:
> > also bobbelcher
> 
> Did you figure out why the sorting is off?

I think it has got to do with the `GIT_COMMITTER_DATE`, can you give me the one you originally used?

rudolfs:
> > Did you figure out why the sorting is off?
> 
> I think it has got to do with the `GIT_COMMITTER_DATE`, can you give me the one you originally used?

The update commit should happen after "Adds a new markdown file with an image with a relative path", so Dec 21 should work.

scripts/create-seed-fixture-heartwood
rudolfs: Try Dec 21.

scripts/create-seed-fixture-heartwood
xphoniex: here's the [3rd node](https://github.com/radicle-dev/radicle-interface/compare/21b3cc087c228898a12c01b51bbb5b583e4b0e49..a7b1031e83d20c213314be9c243740663576b1a9)

xphoniex:
> 

`Nov 21` is used in the [original script](https://github.com/radicle-dev/radicle-interface/blob/master/scripts/create-seed-fixture#L88), want it changed?

here's the log:
```
commit 2febee75cd7731d62c5af47dc98b2ea60b782399 (HEAD -> main, rad/main)
Author:     Bob Belcher <bob@radicle.xyz>
AuthorDate: Mon Nov 21 14:00:00 2022 +0100
Commit:     Bob Belcher <bob@radicle.xyz>
CommitDate: Mon Nov 21 14:00:00 2022 +0100

    Update readme

commit fcc929424b82984b7cbff9c01d2e20d9b1249842 (z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi/main)
Author:     Alice Liddell <alice@radicle.xyz>
AuthorDate: Tue Dec 20 14:28:32 2022 +0100
Commit:     Alice Liddell <alice@radicle.xyz>
CommitDate: Tue Dec 20 14:28:32 2022 +0100

    Adds a new markdown file with an image with a relative path

    Signed-off-by: Alice Liddell <alice@radicle.xyz>
```

which makes sense why ordering is like that.

rudolfs:
> Nov 21 is used in the [original script](https://github.com/radicle-dev/radicle-interface/blob/master/scripts/create-seed-fixture#L88), want it changed?

Yeah, I think we updated the fixture and forgot to update the script. The update comment should always be the latest.

scripts/create-seed-fixture-heartwood
rudolfs: Thank you!

scripts/create-seed-fixture-heartwood
```diff
@@ -27,25 +31,90 @@ mkdir -p $TEST_REPO_PATH
...
+rad auth
+rad auth
+radicle-node --listen 0.0.0.0:3446 \
+   --connect z6MknSLrJoTcukLrE435hVNQT4JUhbvWLX4kUzqkEStBU8Vi@0.0.0.0:3444 \
+   --connect z6MksMTThc1aDU2Ztc43jJUivuyBLNWiLsDf4X65rABe7HbA@0.0.0.0:3445 &
+
+sleep 2
+
+cd $PALM_CHECKOUT
+rad clone rad:zKtT7DmF9H34KkvcKj9PHW19WzjT
```
rudolfs, line 115: We should change the setup slightly. It seems like now we connect the two peers first and then clone it onto the seed. The original setup was like this:

1. alice's peer creates a project
2. alice's peer publishes project onto seed
3. bob's peer fetches alice's project from the seed
4. bob's peer makes changes
5. bob's peer publishes changes to seed

Is it possible to do it this way with heartwood?

scripts/create-seed-fixture-heartwood
```diff
@@ -27,25 +31,90 @@ mkdir -p $TEST_REPO_PATH
...
+
+unset SSH_AUTH_SOCK
+unset SSH_AGENT_PID
+eval $(ssh-agent)
 
-RAD_HOME=$ALICE_RAD_HOME RAD_PASSPHRASE=$PASSPHRASE rad init --name "source-browsing" --description "Git repository for source browsing tests" --default-branch "main" --no-confirm
-PROJECT_ID=$(RAD_HOME=$ALICE_RAD_HOME RAD_PASSPHRASE=$PASSPHRASE rad .)
+rad auth
+rad auth
+radicle-node --listen 0.0.0.0:3445 \
```
rudolfs, line 83: Could we change the setup slightly? It seems like now we connect the two peers first and then clone it onto the seed. The original setup was like this:

alice's peer creates a project
alice's peer publishes project onto seed
bob's peer fetches alice's project from the seed (not directly from alice)
bob's peer makes changes
bob's peer publishes changes to seed

Is it possible to do it this way with heartwood?

xphoniex: [this](https://github.com/radicle-dev/radicle-interface/compare/c69755c740ef79f41bad616d8bd8413622703d0e..07db6d779f98a02970024d26f927517bf1342f82) is the best I can do atm

PR #590, opened 2023-01-25
Rework `Wallet` and `Session` objects, and migrate WalletConnect on the way

sebastinez:
Ok this is most definitely a draft PR.
Reviews are not welcome at the moment 😅 I need to bath a little bit more in this commit soup hehe.


sebastinez:
Okay according to what we talked over Discord (https://discord.com/channels/841318878125490186/856904179820920833/1069181291284660286) we'll remove all eth functionality from the web app, so this PR is not needed anymore

PR #577, opened 2023-01-05
Download radicle-httpd executable for e2e tests

adaszko:
~~Demonstrates how to obtain `radicle-httpd` binary for e2e tests.~~
Uses a statically linked `radicle-httpd` executable to run e2e tests in CI (i.e. no Docker).

PR #597, opened 2023-01-30
Fix clearing snapshot cache

rudolfs:
There was a problem with permissions introduced in https://github.com/radicle-dev/radicle-interface/pull/577.
```
Deleting Linux-snapshots cache
Error: Resource not accessible by integration
Done
```


sebastinez (approved):
🚀

PR #602, opened 2023-01-31
Use radicle-httpd from our public download location

rudolfs:
```
$ ./scripts/run-httpd-with-fixtures --download --non-interactive

Downloading radicle-httpd

Unpacking fixture /Users/rudolfs/work/radicle-interface/tests/fixtures/seeds/palm-heartwood.tar.bz2

Starting radicle-httpd-a5811e4
  RAD_HOME=/Users/rudolfs/work/radicle-interface/tests/tmp/palm RAD_PASSPHRASE=asdf radicle-httpd-a5811e4

2023-01-31T15:48:57.833117Z  INFO radicle_httpd: version 0.1.0-a5811e4
2023-01-31T15:48:57.836700Z  INFO radicle_httpd: git version 2.39.0
2023-01-31T15:48:57.837572Z  INFO radicle_httpd: using radicle home at /Users/rudolfs/work/radicle-interface/tests/tmp/palm
2023-01-31T15:48:57.838592Z  INFO radicle_httpd: listening on http://0.0.0.0:8080
```

If no option is specified, we use the `radicle-httpd` binary that is in `PATH`.
Downloaded binaries are cached so we don't download it every time the script is run.

### DIFF
--- a/.github/workflows/check-e2e.yml
+++ b/.github/workflows/check-e2e.yml
@@ -9,10 +9,6 @@ on:
 
 jobs:
   check-e2e:
-    permissions:
-      contents: "read"
-      id-token: "write"
-      actions: "write" # necessary for evicting snapshot cache
     continue-on-error: true
     strategy:
       matrix:
@@ -58,20 +54,11 @@ jobs:
         if: steps.playwright-dep-cache.outputs.cache-hit != 'true'
         run: npx playwright install chromium firefox
 
-      - id: auth
-        uses: google-github-actions/auth@v1
-        with:
-          workload_identity_provider: "projects/281042598092/locations/global/workloadIdentityPools/github-actions/providers/google-cloud"
-          service_account: "github-actions@radicle-services.iam.gserviceaccount.com"
-
-      - name: Set up Cloud SDK
-        uses: google-github-actions/setup-gcloud@v1
-
       - name: Start http-api test server
         run: |
           if [ ${{ matrix.browser }} = "heartwood" ]; then
             mkdir -p tests/artifacts
-            ./scripts/run-httpd-with-fixtures --non-interactive --binary 2>&1 | tee tests/artifacts/httpd-${{ matrix.browser }}.log &
+            ./scripts/run-httpd-with-fixtures --non-interactive --download 2>&1 | tee tests/artifacts/httpd-${{ matrix.browser }}.log &
           else
             ./scripts/run-http-api-with-fixtures --non-interactive --detach
           fi
--- a/scripts/run-httpd-with-fixtures
+++ b/scripts/run-httpd-with-fixtures
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -e
 
 REV=a5811e43a74043e3bf1e706720bac3ec01b079c4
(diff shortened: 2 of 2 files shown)


adaszko:
Looks good overall 👍 

I’d just a have one nit: With caching (it nice that thought of that), there’s a risk of testing the cached radicle-httpd version, instead of a desired one that’s indicated in  the`REV` env var.  This may happen e.g. when you’re switching git branches or simply forget to do `—download` after git fetch and merge.  In order to prevent that, maybe it’d be better to enforce that the cached version’s REV *always* matches the one indicated under `scripts/run-httpd-with-fixtures::REV`?  The script could error out when they don’t match — it’s better to error out than to test the wrong version.

Let me know if you want me to translate that into a PR.

rudolfs:
> Looks good overall 👍
> 
> I’d just a have one nit: With caching (it nice that thought of that), there’s a risk of testing the cached radicle-httpd version, instead of a desired one that’s indicated in the`REV` env var. This may happen e.g. when you’re switching git branches or simply forget to do `—download` after git fetch and merge. In order to prevent that, maybe it’d be better to enforce that the cached version’s REV _always_ matches the one indicated under `scripts/run-httpd-with-fixtures::REV`? The script could error out when they don’t match — it’s better to error out than to test the wrong version.
> 
> Let me know if you want me to translate that into a PR.

We already ensure that the REV should always match here:

```
  CACHED_BINARY_NAME="$BINARY_NAME-${REV:0:7}"
  if ! [ -x "$(command -v $BINARY_PATH/$CACHED_BINARY_NAME)" ]; then
```

We add a bit of the REV to the binary name and if the one in the filesystem doesn't match the one that's in the REV we download it. If you don't supply `--download` it will use the binary that is in `PATH` without any REV as we had it before. But the downloaded one is in a custom path in `test/tmp/radicle-httpd-xxxxx`, so that won't be run by accident. In other words, you have to always supply the `--download` option to use the one that was downloaded, or copy it into your system path without the REV suffix.

adaszko:
> > Looks good overall 👍
> > I’d just a have one nit: With caching (it nice that thought of that), there’s a risk of testing the cached radicle-httpd version, instead of a desired one that’s indicated in the`REV` env var. This may happen e.g. when you’re switching git branches or simply forget to do `—download` after git fetch and merge. In order to prevent that, maybe it’d be better to enforce that the cached version’s REV _always_ matches the one indicated under `scripts/run-httpd-with-fixtures::REV`? The script could error out when they don’t match — it’s better to error out than to test the wrong version.
> > Let me know if you want me to translate that into a PR.
> 
> We already ensure that the REV should always match here:
> 
> ```
>   CACHED_BINARY_NAME="$BINARY_NAME-${REV:0:7}"
>   if ! [ -x "$(command -v $BINARY_PATH/$CACHED_BINARY_NAME)" ]; then
> ...
>   export PATH="$BINARY_PATH:$PATH"
>   BINARY_NAME=$CACHED_BINARY_NAME
> ```
> 
> We add a bit of the REV to the binary name and if the one in the filesystem doesn't match the one that's in the REV we download it. If you don't supply `--download` it will use the binary that is in `PATH` without any REV as we had it before. But the downloaded one is in a custom path in `test/tmp/radicle-httpd-xxxxx`, so that won't be run by accident. In other words, you have to always supply the `--download` option to use the one that was downloaded, or copy it into your system path without the REV suffix.

Missed that.  Great 👍

adaszko (approved):

PR #603, opened 2023-01-31
New elevation variable

### LINKED ISSUE #507: Explore introducing higher contrast to the UI
I think this would improve the contrast of the UI a lot. Right now it feels a bit washed out, but I think this would do a lot. Need to do a fuller design exploration on it though. And not sure how it would translate to dark mode.

image

brandonhaslegs:
Adds new elevation variable which adds better contrast in light mode closes #507 

Signed-off-by: Brandon Oxendine <brandon.oxendine@gmail.com>

image
image
image
image
image

### DIFF
--- a/public/colors.css
+++ b/public/colors.css
@@ -56,6 +56,7 @@
   --color-foreground-6: #c2c3d4;
 
   --color-background: #0b131a;
+  --color-background-1: #121a21;
 }
 
 [data-theme="light"] {
@@ -116,4 +117,5 @@
   --color-foreground-6: #42425a;
 
   --color-background: #f3f6fd;
+  --color-background-1: #ffffff;
 }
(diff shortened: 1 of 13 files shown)


brandonhaslegs:
This problem didn't really exist in dark mode, but I still added a slightly lighter value for this new `elevation` variable in dark mode. Could just change it to match the `background` value though…

cloudhead:
Looks good to me!

Can you format your commit message according to https://github.com/radicle-dev/radicle-interface/blob/master/CONTRIBUTING.md#writing-git-commit-messages ? Mainly the summary shouldn't be longer than 50 chars otherwise it gets cropped.

rudolfs (commented):
- The `package-lock.json` file should not be committed in this case.
- Could we call it `--color-background-1` instead of `--color-elevation-1`? We use elevations for box-shadows, and they live in `elevations.css`.

Screenshot 2023-02-01 at 10 05 38
Screenshot 2023-02-01 at 10 05 44
Screenshot 2023-02-01 at 10 05 48

rudolfs (commented):
👍

PR #605, opened 2023-02-03
Add connect to node button in header

rudolfs:
We show a `Connect` button in the header when running the app locally in development mode. Clicking the button shows us instructions on how to authenticate with the locally running node. A session is shared and synchronised between tabs and windows.

📸 The visual spec failure is expected, there is a new button in the header.

https://user-images.githubusercontent.com/158411/217855126-3cec9e8a-8f32-4cea-9135-97c20d88b629.mov


cloudhead:
Looking good!

Design suggestions:
* Revert to `Inter` font for the connect button address
* Tweak the left/right padding in the connect button, so that the blockie has the same spacing all around it (remove left padding)
* It might look nicer if the copy button is like in the "Clone" drop down, where it appears on hover. That way the address can be centered, instead of being to the left.

rudolfs:
> Looking good!
> 
> Design suggestions:
> 
> * Revert to `Inter` (regular) font for the connect button address
> * Tweak the left/right padding in the connect button, so that the blockie has the same spacing all around it (remove left padding)
> * It might look nicer if the copy button is like in the "Clone" drop down, where it appears on hover. That way the address can be centered, instead of being to the left.

Is this what you had in mind?


https://user-images.githubusercontent.com/158411/216951139-a6406e09-8da4-4e86-be40-dec5d3b9658a.mov

rudolfs:
> Design suggestions

Is this what you had in mind?


https://user-images.githubusercontent.com/158411/216953207-b44d328c-af00-458f-9bc3-5f9d4428a26a.mov

src/lib/session.ts
```diff
@@ -0,0 +1,67 @@
...
+  );
+  if (request.ok) {
+    save(params.id, params.publicKey);
+  }
+
+  router.push({ resource: "home" });
+}
+
+export function disconnect() {
+  window.localStorage.removeItem("session");
```
sebastinez, line 34: Just FYI, this is not a review 😅
We have a session delete route now in `radicle-httpd`, which we should call when disconnecting the session.
https://github.com/radicle-dev/heartwood/blob/c138cd062c8251a67c0ca784251ae5d7503501b4/radicle-httpd/src/api/v1/sessions.rs#L29

rudolfs: This is implemented now.

src/App/Header/Connect.svelte
```diff
@@ -0,0 +1,236 @@
...
+    top: 5rem;
+    width: 16.5rem;
+  }
+  .info {
+    align-items: flex-start;
+    padding: 1rem;
+    width: 20.5rem;
+  }
+  .cmd {
+    background-color: var(--color-foreground-2);
```
sebastinez, line 54: Maybe we should increase the contrast here?
```suggestion
    background-color: var(--color-foreground-3);
```

In the light theme it's nice, but in the dark one, there is almost no difference between the tooltip and the background of the cmd

**current**
Bildschirm­foto 2023-02-09 um 14 13 53
**proposed**
Bildschirm­foto 2023-02-09 um 14 16 54

src/App/Header/Connect.svelte
```diff
@@ -0,0 +1,236 @@
...
+  import { closeFocused } from "@app/components/Floating.svelte";
+  import { sessionStore, disconnect } from "@app/lib/session";
+  import { toClipboard } from "@app/lib/utils";
+
+  import Avatar from "@app/components/Comment/Avatar.svelte";
+  import Button from "@app/components/Button.svelte";
+  import Floating from "@app/components/Floating.svelte";
+  import Icon from "@app/components/Icon.svelte";
+
+  function formatId(id: string) {
```
sebastinez, line 13: Could we maybe use the `formatRadicleId` from `utils.ts`? Or did you want to have a shorter id?
https://github.com/radicle-dev/radicle-interface/blob/2c5830a076fe08dc01c87476263f11ac9f396c9f/src/lib/utils.ts#L49

src/lib/session.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export async function authenticate(params: {
+  id: string;
+  signature: string;
+  publicKey: string;
+}): Promise<"success" | "failure"> {
+  disconnect();
+
+  const request = await fetch(
+    `http://0.0.0.0:8080/api/v1/sessions/${params.id}`,
```
sebastinez, line 18: I think for this MVP this is ok, but there is the possibility that users run their node with other ports.
Maybe we should allow at some point to add a input field to the Connect tooltip for custom ports?

sebastinez (commented):
Looking really nice 👏 and works fine, left some comments

src/App/Header/Connect.svelte
rudolfs: f6bc61813eb7e280edf5c4ae3fccfe610bffd975

src/App/Header/Connect.svelte
rudolfs: It won't work because we're doing `assert(isRadicleId(id));` for some reason in that function and this id doesn't include `rad:` so the assert would fail. But even if we remove that, I still don't think it looks good in the button:
Screenshot 2023-02-10 at 10 54 55

src/lib/session.ts
rudolfs: Fair. But I think we should wait for people to request this before we make it customisable.

rudolfs:
> Looking good!
> 
> Design suggestions:
> 
> * Revert to `Inter` (regular) font for the connect button address
> * Tweak the left/right padding in the connect button, so that the blockie has the same spacing all around it (remove left padding)
> * It might look nicer if the copy button is like in the "Clone" drop down, where it appears on hover. That way the address can be centered, instead of being to the left.

Made the changes, PTAL!

src/App/Header/Connect.svelte
sebastinez: okay yeah makes sense 👍

sebastinez (approved):
🚀

sebastinez:
PS: I would really like it if we could add some interval maybe 1 minute or something, to check if the session is still valid. Else we won't know if it expired, the API has been restarted or any other valid reason.

```diff
diff --git a/src/lib/session.ts b/src/lib/session.ts
index cf5c710..fad63f8 100644
--- a/src/lib/session.ts
+++ b/src/lib/session.ts
@@ -1,5 +1,13 @@
 import { derived, get, writable } from "svelte/store";
+import { unixTime } from "@app/lib/utils";
 
+interface SessionResponse {
+  sessionId: string;
+  status: string;
+  publicKey: string;
+  issuedAt: number;
+  expiresAt: number;
+}
 export interface Session {
   id: string;
   publicKey: string;
@@ -7,6 +15,30 @@ export interface Session {
 const store = writable<Session | undefined>(undefined);
 export const sessionStore = derived(store, s => s);
 
+setInterval(async () => {
+  const session = get(sessionStore);
+  if (session) {
+    await fetch(`http://0.0.0.0:8080/api/v1/sessions/${session.id}`, {
+      method: "GET",
+      headers: { "Content-Type": "application/json" },
+    })
+      .then(response => response.json())
+      .then((session: SessionResponse) => {
+        if (
+          session.status === "unauthorized" ||
+          session.expiresAt < unixTime()
+        ) {
+          window.localStorage.removeItem("session");
+          store.set(undefined);
+        }
+      })
+      .catch(() => {
+        window.localStorage.removeItem("session");
+        store.set(undefined);
+      });
+  }
+}, 10000);
+
 export async function authenticate(params: {
   id: string;
   signature: string;
```

cloudhead:
LGTM!

rudolfs:
> PS: I would really like it if we could add some interval maybe 1 minute or something, to check if the session is still valid. Else we won't know if it expired, the API has been restarted or any other valid reason.

I'll address this in a follow up. It's a bit more tricky than just doing a `setInterval` because we might have multiple tabs open and we don't want to fire too many requests. There's also a trickery with HMR in dev mode where with every reload the previous `setTimeout` isn't cleared, so we accumulate them over time.

PR #610, opened 2023-02-10
Improve `formatRadicleId` util fn

sebastinez:
- Improves detection of Heartwood Radicle ids.
- Instead of throwing an error if the input isn't a Radicle id we return the original input and display a dev console warning

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -1,7 +1,7 @@
+import bs58 from "bs58";
 import md5 from "md5";
 import twemojiModule from "twemoji";
 
-import { assert } from "@app/lib/error";
 import { base } from "@app/lib/router";
 import { config } from "@app/lib/config";
 
@@ -47,7 +47,10 @@ export function formatSeedId(id: string): string {
 }
 
 export function formatRadicleId(id: string): string {
-  assert(isRadicleId(id));
+  if (!isRadicleId(id)) {
+    console.warn("Received Radicle id isn't valid");
+    return id;
+  }
 
   if (window.HEARTWOOD) {
     return id.substring(0, 10) + "…" + id.substring(id.length - 6, id.length);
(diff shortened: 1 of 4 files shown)


tests/unit/utils.test.ts
```diff
@@ -20,8 +20,8 @@ describe("Format functions", () => {
 
   test("formatRadicleId", () => {
     if (process.env.HEARTWOOD) {
-      expect(utils.formatRadicleId("rad:zKtT7DmF9H34KkvcKj9PHW19WzjT")).toEqual(
-        "rad:zKtT7D…19WzjT",
+      expect(utils.formatRadicleId("zKtT7DmF9H34KkvcKj9PHW19WzjT")).toEqual(
+        "zKtT7D…19WzjT",
```
cloudhead, line 24: Hmm I'm not sure why we're changing this. The above is a valid radicle id in urn format. We also need to specify whether we're talking about repository ids or public keys..

src/lib/utils.ts
```diff
@@ -163,7 +166,13 @@ export const formatTimestamp = (
 // Check whether the input is a Radicle ID.
```
cloudhead, line 166: There's no such thing as a "radicle id" anymore in heartwood. There are node ids (or NIDs), which are public keys, and repository ids (or RIDs) which are Oids.

cloudhead: The former is 32 bytes long, and the latter is 20 bytes long. Both use a multi-base encoding using the bitcoin base58 encoding, though RIDs are often displayed as URNs, with the `rad:` prefix.

tests/unit/utils.test.ts
sebastinez: Interesting, yeah I think we need some design spec on all the identifiers.

Current state AFAIK
```
did:key:<public key>
Returned e.g.from delegate listings, seems more full format

rad:<public key>
Is this for projects only? Why do users don't have rad: prefixes?

<public key>
Defined as PublicKey or ActorId in httpd
```

I'm also ok with maybe a more general format function that can hande different identifier formats

src/lib/utils.ts
sebastinez: Ok this solves my other response wrt to a spec. So we need to two format functions, one for Oids with or without the prefix and one for public keys

sebastinez:
After some brainstorming, I'll close this PR for a more fleshed out one, that cleans up after all the different ids we have floating around

PR #612, opened 2023-02-13
Lower image diff threshold

rudolfs:
The threshold is expressed a percentage, so on full-page screenshots it depends on how big the entire screenshot is. We want to catch smaller changes. The threshold was too high to catch the rounded corners change in a full-page screenshot.


sebastinez (approved):
I'm okay with this change but I think we should watch if we don't get more false negatives out of it. Trying to fine tune the threshold

PR #613, opened 2023-02-13
Don't expand search bar on mobile

rudolfs:
Also don't show the connect button in the mobile layout. We don't support running nodes on mobile anyway.
On desktop still expand and contract the search bar depending on focus.

Mobile:
Screenshot 2023-02-13 at 17 53 38

Desktop:
Screenshot 2023-02-13 at 17 55 21

### DIFF
--- a/src/App/Header/Search.svelte
+++ b/src/App/Header/Search.svelte
@@ -16,14 +16,13 @@
   }>();
 
   export let input = "";
+
   let loading = false;
   let shaking = false;
+  let expanded = false;
 
   // Clears search input on user navigation.
   router.historyStore.subscribe(() => (input = ""));
-
-  const collapsedWidth = 13;
-  const expandedWidth = 20;
 
   function shake() {
     shaking = true;
@@ -69,7 +68,7 @@
         });
       }
       dispatch("finished");
-      searchBarWidth = collapsedWidth;
+      expanded = false;
     } else {
       unreachable(searchResult);
     }
@@ -77,7 +76,6 @@
   }
 
   $: valid = input !== "";
-  $: searchBarWidth = collapsedWidth;
 </script>
 
 <style>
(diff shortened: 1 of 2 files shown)


src/App/Header/Search.svelte
```diff
@@ -23,7 +23,12 @@
   router.historyStore.subscribe(() => (input = ""));
 
   const collapsedWidth = 13;
-  const expandedWidth = 20;
+  let searchBarWidth = collapsedWidth;
+
+  function expandSearchBar() {
+    const isMobile = window.matchMedia("(max-width: 720px)").matches;
```
sebastinez, line 29: Is there some reason we need this to be a js function doing css?
Couldn't we toggle between two CSS classes with different widths `on:focus` on the div wrapping the `TextInput`.
Something like 
```css
.expandedWidth {
 width: 20rem;
}
.collapsed {
  width: 13rem;
}
```
And the case when on mobile should be possible to be done with CSS media queries

sebastinez (commented):
Wasn't able to run the PR, due to current notebook constraints, but left a comment wrt to using css for styling instead of js

src/App/Header/Search.svelte
rudolfs: Theoretically yes, but I didn't want for the search bar to contract when it looses focus when a search is in progress. This would also happen when a search fails.

src/App/Header/Search.svelte
```diff
@@ -107,11 +111,6 @@
...
     transition: all 0.2s;
   }
-  @media (max-width: 720px) {
-    .search {
-      display: none;
-    }
-  }
 </style>
 
 <div class="search" style:width={`${searchBarWidth}rem`}>
```
sebastinez, line 116: wrt to my first comment here we could do something like this
```suggestion
<div class="search" class:collapsed={isCollapsed} class:expanded={!isCollapsed}>
```

src/App/Header/Search.svelte
rudolfs: Yeah, good idea. 90b04ebf2484fbcadcbd29c42da4651edd56b329

sebastinez (approved):
🚀

PR #614, opened 2023-02-14
Poll session every minute to make sure we are still connected

rudolfs:
This disconnects the frontend in case the backend is not online anymore or if the session has expired.


rudolfs:
I tried messing with exponential backoffs to reduce requests if the user has multiple tabs open, but it got complex very fast. Decided that the downsides weren't as critical to warrant the extra complexity.

sebastinez (approved):
👍

PR #621, opened 2023-02-16
Migrate to the heartwood protocol stack

rudolfs:
We remove all legacy radicle-link code. 🚮
From now on we can only interact with heartwood nodes.
The snapshot failures are expected.

Screenshot 2023-03-02 at 14 54 29


src/lib/utils.ts
```diff
@@ -77,9 +73,33 @@ export function parseNid(
...
+    }
+
+    return { prefix: match[1] || "rad:", pubkey: match[2] };
+  }
+
+  return undefined;
+}
+
+export function isNodeId(input: string): boolean {
+  return !!parseNodeId(input);
```
sebastinez, line 93: Instead of a double negation shouldn't we be able to do here
```suggestion
  return Boolean(parseNodeId(input));
```

src/lib/utils.ts
```diff
@@ -47,19 +49,13 @@ export function formatSeedId(id: string): string {
   return id.substring(0, 6) + "…" + id.substring(id.length - 6, id.length);
 }
 
-export function formatRadicleId(id: string): string {
-  assert(isRadicleId(id));
+export function formatRepositoryId(id: string): string {
```
sebastinez, line 52: I think this function could be a bit more similar to `formatNodeId`, it isn't very clear from the function why we do `substring(0, 10)` and if we pass a RID without a `rad:` prefix this functions doesn't behave as needed.
But this can be done in a separate PR

src/lib/issue.ts
```diff
@@ -70,29 +67,17 @@ export class Issue {
...
-        return acc + 1; // If there are no replies, we simply add 1 for the comment in this loop.
-      }, 0);
-    } else {
-      return this.discussion.reduce((acc, comment) => {
-        if (comment.replies) return acc + comment.replies.length + 1; // We add all replies and 1 forathe comment in this loop.
-        return acc + 1; // If there are no replies, we simply add 1 for the comment in this loop.
-      }, 0);
-    }
+    return this.discussion.reduce(acc => {
+      return acc + 1; // If there are no replies, we simply add 1 for the comment in this loop.
```
sebastinez, line 79: :thinking: At some point we added the `+ 1` to not have issues with 0 comments. But since we now always will have the initial description as part of the `discussion` thread, maybe we need to skip the first item or don't count the first comment..

src/lib/project.ts
```diff
@@ -2,28 +2,16 @@ import type { Commit, CommitHeader, CommitsHistory } from "@app/lib/commit";
...
 import type { ProjectResult } from "@app/lib/search";
 
+import * as utils from "@app/lib/utils";
 import { Request } from "@app/lib/api";
-import { Seed, defaultSeedPort } from "@app/lib/seed";
-import { isFulfilled, isOid, isRadicleId } from "@app/lib/utils";
+import { Seed } from "@app/lib/seed";
+import { isFulfilled, isOid, isRepositoryId } from "@app/lib/utils";
 
 export type Id = string;
```
sebastinez, line 10: Can we also remove this alias?
Since at one place it is used as a Repository Id and at the other as Node Id, it's confusing.

src/lib/project.ts
```diff
@@ -56,32 +43,22 @@ export interface ProjectInfo {
 
 export interface Tree {
```
sebastinez, line 44: AFAIK we don't return `kind` in heartwood, so we can remove it

See httpd json build fn https://github.com/radicle-dev/heartwood/blob/9d244b221a167c1c2bacbd00ecefc1a44ecab552/radicle-httpd/src/api/json.rs#L68

src/lib/project.ts
```diff
@@ -56,32 +43,22 @@ export interface ProjectInfo {
...
-  objectType: Kind;
-  lastCommit: CommitHeader;
-}
-
 export interface Entry {
   path: string;
-  info: EntryInfo; // TODO: Remove this after we have migrated to Heartwood.
   name: string;
   kind: Kind;
   lastCommit: CommitHeader;
```
sebastinez, line 64: We should be able to remove the `lastCommit` property in the `Entry` or in the `Blob` interface

See: https://github.com/radicle-dev/heartwood/blob/9d244b221a167c1c2bacbd00ecefc1a44ecab552/radicle-httpd/src/api/json.rs#L68

src/views/projects/CloneButton.svelte
```diff
@@ -1,14 +1,17 @@
...
   import { closeFocused } from "@app/components/Floating.svelte";
 
   export let seedHost: string;
   export let id: string;
+  export let name: string;
 
-  $: radCloneUrl = `rad clone rad://${seedHost}/${utils.parseRadicleId(id)}`;
-  $: gitCloneUrl = `https://${seedHost}/${utils.parseRadicleId(id)}.git`;
+  $: radCloneUrl = `rad clone ${id}`;
+  $: gitCloneUrl = `git clone https://${seedHost}/${id.replace(
```
sebastinez, line 11: If we would have a `parseRepositoryId` as mentioned in another comment, we could make use of it here instead of relying on `String.replace()`

sebastinez (commented):
Great work :partying_face: Really loving everything you were able to remove wrt to seed handling, hosts, etc.
Left some minor nits, and some thoughts for later PRs.

src/views/seeds/View.svelte
```diff
@@ -99,7 +101,9 @@
     <div class="fields">
       <!-- Seed Address -->
       <div class="txt-highlight">Address</div>
-      <SeedAddress {seed} port={seed.node.port} />
+      <SeedAddress
+        {seed}
+        port={seed.node.port ?? config.seeds.defaultHttpdPort} />
       <!-- Seed ID -->
       <div class="txt-highlight">Seed ID</div>
       <div class="seed-label">
```
sebastinez, line 109: We should change `formatSeedId` to `formatNodeId` here

sebastinez:
Since we said we wanted to show the `did:key:` prefix in all Node Ids maybe we also should show it in the connect button and the Connect Toggle?

Screenshot from 2023-03-02 22-49-34

sebastinez:
It's maybe not in my best interest, but I think we could also remove everything related to Patches with this PR, since it's based on legacy and has to be rewritten and rethought for heartwood. wdyt?

src/components/Comment/Avatar.svelte
```diff
@@ -23,10 +23,10 @@
     return avatar.toDataURL();
   }
 
-  if (isRadicleId(source) || isPeerId(source)) {
+  if (isNodeId(source)) {
     source = createContainer(source);
   }
-  grayscale = isPeerId(title) || isRadicleId(title);
+  grayscale = isNodeId(title);
```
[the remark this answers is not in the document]
rudolfs: 38050c7a1551b4a83d0a995d3efedd845c26741a

src/lib/utils.ts
rudolfs: 5696b913e4687afb0a54aa4eaa4e159769369204

src/lib/utils.ts
rudolfs: fb32bc381be89b503d67bfd6e7edada9fca6c02c

src/lib/issue.ts
rudolfs: Hm, I think it's kind of weird that we separate the issue title from the body. Now the body is a comment. If we remove the +1 then we lose the "1 comment" icon in the listing and the issue shows one comment which is the body.

Screenshot 2023-03-03 at 09 29 35
Screenshot 2023-03-03 at 09 29 41

If we keep the body separate as a comment, I think it's more consistent to leave the +1.

src/lib/project.ts
rudolfs: cb3451dce4a3a0f0f96c4e218b62ea82371ace39

src/lib/project.ts
rudolfs: eeacc38feec9059cfc5653449c3f7f39e2e312c9

src/lib/project.ts
rudolfs: This is why we need zod, we'd get an exception if there was a discrepancy like this. :P
`Blob` seems to have a `lastCommit` if I'm not misreading the code? Will remove from `Entry`.

f99248fc07a7ab02411565b63793ca66a7956c71

src/views/projects/CloneButton.svelte
rudolfs: 6e3b46c5bbe741367c826db1dc4cf31232598345

src/lib/project.ts
sebastinez: Fully agree on having zod :100:

src/views/seeds/View.svelte
rudolfs: 55b07766a9d5ba3a93f0f4a1e31db8355f0cd0ba

src/lib/issue.ts
sebastinez: Yeah I'm also not sure on having the description as a first comment tbh. It sounds better to me to have it together with the issue title, but for consistency we can leave it with the +1

src/lib/utils.ts
sebastinez: I think this looks very nice.
So a minor nit, and I know you don't like indirection.. but wdyt about having a `formatRadicleId` and a `formatNodeId` wrapper around `formatId`?

E.g.
```
export function formatRepositoryId(id: string): string {
  return formatId(id, "repository");
}
```

So we don't have to specify on every format what we want to format but it's already in the function name?

rudolfs:
Screenshot 2023-03-03 at 10 39 41
Screenshot 2023-03-03 at 10 39 45
Screenshot 2023-03-03 at 10 39 57
Screenshot 2023-03-03 at 10 39 34

rudolfs:
> It's maybe not in my best interest, but I think we should also remove the current patches implementation with this PR, since it's based on legacy and has to be rewritten and rethought for heartwood. wdyt?

67dc4cc457f4923fab2a24b6e5dbd64b07a523fc

src/lib/project.ts
```diff
@@ -29,11 +27,6 @@ export interface ProjectInfo {
   description: string;
   defaultBranch: string;
   delegates: string[];
-  patches: {
```
sebastinez, line 46: Ok I think I would still keep this part of patches, since the httpd already returns the count of patches.
Like seen here https://seed.radicle.xyz/api/v1/projects/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5.

And also keep it as part of the `Project` class.

The rest looks great. Black hole development :sweat_smile:

rudolfs: 5a299aab19e0a268c3c98e19b8746fd0a3e44739

src/lib/utils.ts
rudolfs: 126ecb52184b44df5ec45ff7ea7a5eb38f2d737a

tests/unit/utils.test.ts
```diff
@@ -9,37 +9,30 @@ describe("Format functions", () => {
...
+
   test.each([
     {
       id: "did:key:z6MkmzRwg47UWQxczLLLFfkEwpBGitjzJ1vKPE8U9ymd6fz6",
+      kind: "node" as const,
       expected: "did:key:z6Mkmz…md6fz6",
     },
     {
       id: "z6MkmzRwg47UWQxczLLLFfkEwpBGitjzJ1vKPE8U9ymd6fz6",
+      kind: "node" as const,
```
sebastinez, line 29: We should be able to remove this `kind` property from here right?

tests/unit/utils.test.ts
```diff
@@ -131,8 +145,8 @@ describe("Parse Functions", () => {
         pubkey: "z6MkmzRwg47UWQxczLLLFfkEwpBGitjzJ1vKPE8U9ymd6fz6",
       },
     },
-  ])("parseNid", ({ input, expected }) => {
-    expect(utils.parseNid(input)).toEqual(expected);
+  ])("parseNodeId", ({ input, expected }) => {
+    expect(utils.parseNodeId(input)).toEqual(expected);
```
sebastinez, line 149: Could we add a unit test for `parseRepositoryId`?

sebastinez (approved):
I have two minor changes for testing and the rest looks great.
Great work :clap:

tests/unit/utils.test.ts
rudolfs: e0cd45284c851e565ec0d9c81a638343c91c4251

tests/unit/utils.test.ts
rudolfs: 243e8e3b3d23221e06f96c3a0a8dad363bc90458

PR #622, opened 2023-02-22
Update dependencies

rudolfs:
To keep us up-to-date. This also updates transitive dependencies.


sebastinez:
It seems that the build smoke test is failing because of https://github.com/microsoft/playwright/pull/21106

rudolfs:
> It seems that the build smoke test is failing because of [microsoft/playwright#21106](https://github.com/microsoft/playwright/pull/21106)

Oh, good catch. I was trying to figure this out yesterday, but couldn't find what was wrong with it.

rudolfs:
There seems to be a regression in FF: https://github.com/microsoft/playwright/issues/21145.

sebastinez:
> There seems to be a regression in FF: [microsoft/playwright#21145](https://github.com/microsoft/playwright/issues/21145).

Maybe we should exclude playwright from this update? Since there are some other dependencies that are either major updates or at least minor ones.

rudolfs:
> Maybe we should exclude playwright from this update? Since there are some other dependencies that are either major updates or at least minor ones.

Yeah, can do.

sebastinez (approved):
LGTM, seems to work nicely

rudolfs:
Hm, on second thought, I think this won't work, because the problem seems to be with the newer firefox browser version. And we don't have a way to not update the browsers when bumping deps. So we either disable the firefox tests or wait for playwright to fix it on their end.

sebastinez:
> Hm, on second thought, I think this won't work, because the problem seems to be with the newer firefox browser version. And we don't have a way to not update the browsers when bumping deps.

So we are kinda vendor locked in by Microsoft :sweat_smile:

PR #627, opened 2023-03-07
Fix some router issues

sebastinez:
- Clicking on the project in various places goes now to `../<rid>` instead of `../<rid>/tree/<commit>`
- Project URL on seed allows trailing slash, eg. `../<rid>/`
- Navigating to a project URL with a `remotes/<nodeId>` suffix works now correctly w or w/o trailing slash


rudolfs (approved):
Nice work! 🚀

PR #629, opened 2023-03-07
Cleanup commit types

sebastinez:
Starting out for cleaning lines like `commit.commit...`

This commit removes the `CommitContext` type which isn't in Heartwood anymore.
Also the CommitVerifiedBadge component isn't of use, since we haven't this information currently, will bring back eventually


rudolfs (approved):
🧹 🔥

PR #630, opened 2023-03-07
Always pass diff step and show the status as a comment

rudolfs:
Splits out the visual diffs into a separate workflow and adds a Github bot that adds a comment about the visual diff status. It edits the existing comment if there is one.

Screenshot 2023-03-08 at 12 09 44

Screenshot 2023-03-08 at 12 21 46


cloudhead:
:rocket:

sebastinez (approved):
Nice solution :+1: and looks fine too!

PR #631, opened 2023-03-07
Issue editing

sebastinez:
This PR will add to issues:

- Add threads
- Add replies
- Edit assignees & tags
- Manage lifecycle

### DIFF
--- a/src/views/projects/Issue.svelte
+++ b/src/views/projects/Issue.svelte
@@ -1,169 +1,207 @@
-<script lang="ts">
+<script lang="ts" strictEvents>
   import type { Project } from "@app/lib/project";
-  import type { Issue } from "@app/lib/issue";
+  import type { State } from "@app/lib/issue";
 
-  import Authorship from "@app/components/Authorship.svelte";
-  import Avatar from "@app/components/Comment/Avatar.svelte";
-  import Chip from "@app/components/Chip.svelte";
-  import Comment from "@app/components/Comment.svelte";
-  import { formatNodeId, capitalize } from "@app/lib/utils";
-  import { formatObjectId } from "@app/lib/cobs";
+  import Button from "@app/components/Button.svelte";
+  import IssueHeader from "@app/views/projects/Issue/IssueHeader.svelte";
+  import IssueSidebar from "@app/views/projects/Issue/IssueSidebar.svelte";
+  import IssueStateButton from "@app/views/projects/Issue/IssueStateButton.svelte";
+  import Textarea from "@app/components/Textarea.svelte";
+  import Thread from "@app/components/Thread.svelte";
+  import { createAddRemoveArrays, Issue } from "@app/lib/issue";
+  import { createEventDispatcher } from "svelte";
+  import { isLocal } from "@app/lib/utils";
+  import { sessionStore } from "@app/lib/session";
 
   export let issue: Issue;
   export let project: Project;
+
+  const dispatch = createEventDispatcher<{ update: never }>();
+  const rawPath = project.getRawPath();
+
+  async function createReply({
+    detail: reply,
+  }: CustomEvent<{ id: string; body: string }>) {
+    if ($sessionStore && reply.body.trim().length > 0) {
+      await issue.replyComment(
+        project.id,
+        reply.body,
+        reply.id,
+        project.seed.addr,
+        $sessionStore.id,
+      );
+      issue = await Issue.getIssue(project.id, issue.id, project.seed.addr);
+    }
+  }
+
+  async function createComment(body: string) {
+    if ($sessionStore && body.trim().length > 0) {
+      await issue.createComment(
+        project.id,
+        body,
+        project.seed.addr,
+        $sessionStore.id,
+      );
+      issue = await Issue.getIssue(project.id, issue.id, project.seed.addr);
+    }
+  }
+
+  async function editTitle({ detail: title }: CustomEvent<string>) {
+    if ($sessionStore && title.trim().length > 0 && title !== issue.title) {
+      await issue.editTitle(
+        project.id,
+        title,
+        project.seed.addr,
+        $sessionStore.id,
+      );
+      issue = await Issue.getIssue(project.id, issue.id, project.seed.addr);
+    } else {
+      // Reassigning issue.title overwrites the invalid title in IssueHeader
+      issue.title = issue.title;
+    }
+  }
+
+  async function saveTags({ detail: tags }: CustomEvent<string[]>) {
+    if ($sessionStore) {
+      const { add, remove } = createAddRemoveArrays(issue.tags, tags);
+      if (add.length === 0 && remove.length === 0) {
+        return;
+      }
+      await issue.editTags(
+        project.id,
+        add,
+        remove,
+        project.seed.addr,
+        $sessionStore.id,
+      );
+      issue = await Issue.getIssue(project.id, issue.id, project.seed.addr);
+    }
+  }
+
+  async function saveAssignees({ detail: assignees }: CustomEvent<string[]>) {
+    if ($sessionStore) {
+      const { add, remove } = createAddRemoveArrays(issue.assignees, assignees);
+      if (add.length === 0 && remove.length === 0) {
+        return;
+      }
+      await issue.editAssignees(
+        project.id,
+        add,
+        remove,
+        project.seed.addr,
+        $sessionStore.id,
+      );
+      issue = await Issue.getIssue(project.id, issue.id, project.seed.addr);
+    }
+  }
+
+  async function saveStatus({ detail: state }: CustomEvent<State>) {
+    if ($sessionStore) {
+      await issue.changeState(
+        project.id,
+        state,
+        project.seed.addr,
+        $sessionStore.id,
+      );
+      dispatch("update");
+      issue = await Issue.getIssue(project.id, issue.id, project.seed.addr);
+    }
+  }
+
+  $: threads = issue.discussion
+    .filter(comment => !comment.replyTo)
+    .map(thread => {
+      return {
+        root: thread,
+        replies: issue.discussion
+          .filter(comment => comment.replyTo === thread.id)
+          .sort((a, b) => a.timestamp - b.timestamp),
+      };
+    }, []);
+
+  let commentBody: string = "";
 </script>
 
 <style>
-  header {
-    padding: 1rem;
-    background: var(--color-foreground-1);
-    border-radius: var(--border-radius);
-    margin-bottom: 2rem;
-  }
-  main {
-    display: flex;
-  }
   .issue {
+    display: grid;
+    grid-template-columns: minmax(0, 3fr) 1fr;
     padding: 0 2rem 0 8rem;
-  }
-  .comments {
-    flex: 1;
-  }
-  .metadata {
-    flex-basis: 18rem;
-    margin-left: 1rem;
-    border-radius: var(--border-radius);
-    font-size: var(--font-size-small);
-    padding-left: 1rem;
-  }
-  .metadata-section {
-    margin-bottom: 1rem;
-    border-bottom: 1px dashed var(--color-foreground-4);
-  }
-  .metadata-section-header {
-    font-size: var(--font-size-small);
-    margin-bottom: 0.75rem;
-    color: var(--color-foreground-5);
-  }
-  .metadata-section-body {
-    display: flex;
-    flex-wrap: wrap;
-    flex-direction: row;
-    gap: 0.5rem;
-    margin-bottom: 1.25rem;
-  }
-  .metadata-section-empty {
-    color: var(--color-foreground-6);
+    margin-bottom: 4.5rem;
   }
 
-  .summary {
+  .actions {
     display: flex;
     flex-direction: row;
-    align-items: center;
-    margin-bottom: 0.5rem;
+    justify-content: flex-end;
+    margin: 0 0 2.5rem 0;
+    gap: 1rem;
   }
-  .summary-title {
-    overflow: hidden;
-    white-space: nowrap;
-    text-overflow: ellipsis;
-  }
-  .id {
-    flex: 1 0 auto;
-    font-size: var(--font-size-tiny);
-    margin-left: 0.75rem;
-    color: var(--color-foreground-5);
-  }
-  .summary-state {
-    margin-left: 2rem;
-    padding: 0.5rem 1rem;
-    border-radius: var(--border-radius);
-  }
-  .open {
-    color: var(--color-positive);
-    background-color: var(--color-positive-2);
-  }
-  .closed {
-    color: var(--color-negative);
-    background-color: var(--color-negative-2);
-  }
-  .assignee {
-    display: flex;
-    flex-direction: row;
-    align-items: center;
-  }
-  .tag {
-    max-width: 15rem;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
+
+  @media (max-width: 720px) {
+    .issue {
+      display: grid;
+      grid-template-columns: minmax(0, 1fr);
+      padding-left: 2rem;
+    }
   }
 
   @media (max-width: 960px) {
     .issue {
       padding-left: 2rem;
     }
-    .summary-state {
-      margin-left: 0.5rem;
-    }
   }
 </style>
 
 <div class="issue">
-  <header>
-    <div class="summary">
-      <div class="summary-title txt-medium">
-        {issue.title}
-      </div>
-      <div class="txt-monospace id layout-desktop">{issue.id}</div>
-      <div class="txt-monospace id layout-mobile">
-        {formatObjectId(issue.id)}
-      </div>
-      <div
-        class="summary-state"
-        class:closed={issue.state.status === "closed"}
-        class:open={issue.state.status === "open"}>
-        {capitalize(issue.state.status)}
-      </div>
-    </div>
-    <Authorship
+  <div>
+    <IssueHeader
+      action="edit"
+      id={issue.id}
       author={issue.author}
       timestamp={issue.timestamp}
-      caption="opened" />
-  </header>
-  <main>
+      state={issue.state}
+      title={issue.title}
+      on:editTitle={editTitle} />
     <div class="comments">
-      {#each issue.discussion as comment}
-        <Comment {comment} rawPath={project.getRawPath()} />
+      {#each threads as thread, index (thread.root.id)}
+        <Thread
+          {thread}
+          {rawPath}
+          isDescription={index === 0}
+          on:reply={createReply} />
       {/each}
-    </div>
-    <div class="metadata layout-desktop">
-      <div class="metadata-section">
-        <div class="metadata-section-header">Assignees</div>
-        <div class="metadata-section-body">
-          {#if issue.assignees?.length}
-            {#each issue.assignees as assignee, key}
-              <Chip {key}>
-                <div slot="text" class="assignee">
-                  <Avatar inline source={assignee} title={assignee} />
-                  <span>{formatNodeId(assignee)}</span>
-                </div>
-              </Chip>
-            {/each}
-          {:else}
-            <div class="metadata-section-empty">No assignees</div>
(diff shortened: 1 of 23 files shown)


sebastinez:
I have some minor UI changes pending that I would like to add to this PR.
- moving the issue state label under the title and add the authorship component there.
- show an edit title button where the issue state label currently is

rudolfs:
#### UI/UX review

- what's this line in the issue preview?
  Screenshot 2023-03-10 at 11 17 49
  Screenshot 2023-03-10 at 11 17 55

sebastinez:
>     * what's this line in the issue preview?
This has to do with the reply icon, which is being separated by this border and I missed adding a `disableReply` property in the `New` component.

>     * the title `edit` button should change to `save` when editing is toggled, also the input box looks weird, it should probably expand maybe even to full width:
>       Screenshot 2023-03-10 at 11 20 59

Yeah I wanted to avoid that the content changes place when editing the title, but it seems also that GH changes the title for a text input so I probably will do that. 

>     * the header looks a bit messy, there should be more padding between the header and the body or potentially a dividing line? maybe the `closed` label should be smaller and it could be actually two labels, one for the status `closed`/`open` and another one for the reason `resolved`/`other`.
>       Screenshot 2023-03-10 at 11 22 56

Looking into this, wdyt @cloudhead? I'm all for improvements.

>     * for some reason hotkeys get triggered when typing in the title textarea, doesn't happen in the comment or reply textareas.

This could be some missing `stopPropagation`.. Will look into this.

>     * the dropdown button behaviour is a bit unexpected, i expected the action to be invoked when clicking in the dropdown instead of just changing the button title (similar to how it's on GH).
>       Screenshot 2023-03-10 at 11 32 04

The `IssueStateButton` behaviour is the same as on GH. On GH you either click the default selected action (Reopen / Close as solved) or you change the button action through the dropdown and then click on the button..
Maybe you are right, and clicking on the dropdown should already trigger the state change. wdyt? @cloudhead 

>     * the reply button is in a weird place, can we move it to the top right-hand corner? and maybe add a "reply" text. could be a button probably. and can we show the reply textbox at the bottom of the thread?
>       Screenshot 2023-03-10 at 11 42 30

Yeah I experimented with that, but I felt that the button was too far away from the rest of the comment most of the time..
My idea is to also add e.g. the reaction Icon to the right of the reply button, so we match the UX of GH. 
Screenshot from 2023-03-10 12-31-35
+1 for moving the textbox to the end of the thread

>     * i'm able to add the same assignee without any validation messages
>       Screenshot 2023-03-10 at 11 47 08

Ok this is a regression, thanks for the catch

>     * it is a bit odd that we can edit the issue title, but not the body shrug

Yeah that needs the implementation of thread editing on the httpd which isn't yet done.

>     * there shouldn't be any network requests if we toggle the edit/save buttons without any changes for tags and assignees

Yeah that is a good idea, I implemented it for the issue title but should also add it to assignees and tags.

rudolfs:
> The IssueStateButton behaviour is the same as on GH. On GH you either click the default selected action (Reopen / Close as solved) or you change the button action through the dropdown and then click on the button..
Maybe you are right, and clicking on the dropdown should already trigger the state change. wdyt? @cloudhead

Ah, you're right, nevermind.

src/views/projects/Issue.svelte
```diff
@@ -118,59 +210,114 @@
 <div class="issue">
   <header>
     <div class="summary">
-      <div class="summary-title txt-medium">
-        {issue.title}
-      </div>
+      {#if titleEditable}
+        <div
+          contenteditable="true"
```
rudolfs: We should probably use a TextInput or Textarea component here.

src/views/projects/Issue/IssueSidebar.svelte
```diff
@@ -0,0 +1,210 @@
...
+  function handleAddAssignee() {
+    const nodeId = parseNodeId(assignee);
+    if (nodeId) {
+      if (assignees.includes(nodeId.pubkey)) {
+        assigneeCaption = "This user is already assigned";
+        return;
+      }
+      assignees.push(`${nodeId.prefix}${nodeId.pubkey}`);
+      // Allow component to react to array change.
+      assignees = assignees;
```
rudolfs, line 35: We could just do:
```
assignees = [...assignees, `${nodeId.prefix}${nodeId.pubkey}`]
```

src/views/projects/Issue/IssueSidebar.svelte
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  function handleAddTag() {
+    if (tags.includes(tag)) {
+      tagCaption = "This tag already exists";
+      return;
+    }
+    tags.push(tag);
+    // Allow component to react to array change.
+    tags = tags;
```
rudolfs, line 52: Same as above.

src/views/projects/Issue/IssueSidebar.svelte
```diff
@@ -0,0 +1,210 @@
...
+          <Button
+            size="tiny"
+            variant="text"
+            on:click={() => {
+              dispatch("saveAssignees", assignees);
+              editAssignees = !editAssignees;
+            }}>
+            save
+          </Button>
+        {/if}
```
rudolfs, line 124: Can we clean up this `if`, it seems like `if editAssignees ... else ...` should do, no?

src/views/projects/Issue/IssueSidebar.svelte
```diff
@@ -0,0 +1,210 @@
...
+      <span>Tags</span>
+      {#if action === "edit"}
+        {#if !editTags}
+          <Button
+            size="tiny"
+            variant="text"
+            on:click={() => (editTags = !editTags)}>
+            edit
+          </Button>
+        {:else if editTags}
```
rudolfs, line 169: This should also be a plain `if ... else ...`.

rudolfs (commented):
Did a preliminary pass on the code. Will have a closer look when the UI/UX things have been resolved.

src/views/projects/Issue.svelte
sebastinez: done in 56f8377

src/views/projects/Issue.svelte
```diff
@@ -119,58 +204,102 @@
...
-  <main>
+  <main class="form">
     <div class="comments">
       {#each issue.discussion as comment}
-        <Comment {comment} {getImage} />
+        {#if comment.replyTo === null}
+          <Reply
+            bind:reply
+            allowReply={Boolean(session)}
+            on:reply={() => createReply(comment.id, reply)}
```
rudolfs, line 260: Do we need to both pass the reply via bind _and_ the event? If not I'd just use the event to pass it back.

sebastinez: Yeah good catch, I think we can just keep the event.

rudolfs:
I feel like the main actions should have buttons that are more prominent (i.e. with outline), similar to how we do it on issue creation. Also added a bit of margin to the bottom of the page so the dropdown fully opens and played a bit with the alignment.

Screenshot 2023-03-14 at 16 06 15

```diff
diff --git a/src/components/Button.svelte b/src/components/Button.svelte
index e46915a0..d6ae7f70 100644
--- a/src/components/Button.svelte
+++ b/src/components/Button.svelte
@@ -36,7 +36,6 @@
     color: var(--color-background);
   }
   .attachable {
-    border: 0;
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
     border-right: 0;
diff --git a/src/components/Comment/Reply.svelte b/src/components/Comment/Reply.svelte
index 4d76fc30..16844654 100644
--- a/src/components/Comment/Reply.svelte
+++ b/src/components/Comment/Reply.svelte
@@ -38,7 +38,7 @@
     <Textarea resizable bind:value={reply} placeholder="Leave your reply" />
     <div class="actions">
       <Button
-        variant="text"
+        variant="secondary"
         size="small"
         disabled={!reply}
         on:click={() => {
diff --git a/src/components/DropdownButton.svelte b/src/components/DropdownButton.svelte
index 84a9766e..f6af8696 100644
--- a/src/components/DropdownButton.svelte
+++ b/src/components/DropdownButton.svelte
@@ -28,7 +28,7 @@
   }
   .toggle {
     cursor: pointer;
-    border: 0;
+    border: 1px solid var(--color-foreground);
     border-radius: var(--border-radius-round);
     border-top-left-radius: 0;
     height: var(--button-small-height);
diff --git a/src/views/projects/Issue.svelte b/src/views/projects/Issue.svelte
index 7712fa14..056f8106 100644
--- a/src/views/projects/Issue.svelte
+++ b/src/views/projects/Issue.svelte
@@ -117,6 +117,7 @@
   }
   .issue {
     padding: 0 2rem 0 8rem;
+    margin-bottom: 4.5rem;
   }
   .comments {
     flex: 1;
@@ -276,7 +277,7 @@
           <IssueStateButton {issue} on:saveStatus={saveStatus} />
           <div class="spacer" />
           <Button
-            variant="text"
+            variant="secondary"
             size="small"
             disabled={!commentBody}
             on:click={() => {
diff --git a/src/views/projects/Issue/IssueStateButton.svelte b/src/views/projects/Issue/IssueStateButton.svelte
index 5a38815a..6b17c66d 100644
--- a/src/views/projects/Issue/IssueStateButton.svelte
+++ b/src/views/projects/Issue/IssueStateButton.svelte
@@ -53,13 +53,14 @@
   }
   .dropdown {
     position: absolute;
-    left: 0;
+    right: 11rem;
+    top: 2.5rem;
   }
 </style>
 
 <div class="main">
   <DropdownButton
-    variant="text"
+    variant="foreground"
     size="small"
     on:toggle={() => (showStateDropdown = !showStateDropdown)}
     on:click={() => dispatch("saveStatus", selectedItem.value)}>
```

rudolfs:
We could improve the sidebar by removing the border and adding more space between tags and asignees.

Before:

Screenshot 2023-03-14 at 16 10 48

After:
Screenshot 2023-03-14 at 16 09 42



```diff
diff --git a/src/views/projects/Issue/IssueSidebar.svelte b/src/views/projects/Issue/IssueSidebar.svelte
index 75ccfe28..39126d14 100644
--- a/src/views/projects/Issue/IssueSidebar.svelte
+++ b/src/views/projects/Issue/IssueSidebar.svelte
@@ -72,8 +72,7 @@
     padding-left: 1rem;
   }
   .metadata-section {
-    margin-bottom: 1rem;
-    border-bottom: 1px dashed var(--color-foreground-4);
+    margin-bottom: 4rem;
   }
   .metadata-section-header {
     display: flex;
```

rudolfs:
We should use the `Badge.svelte` component to keep the UI consistent, we already use it for commits.


Screenshot 2023-03-14 at 16 23 29
Screenshot 2023-03-14 at 16 23 34
Screenshot 2023-03-14 at 16 23 38


```diff
diff --git a/src/views/projects/Issue.svelte b/src/views/projects/Issue.svelte
index 056f8106..a8865ba5 100644
--- a/src/views/projects/Issue.svelte
+++ b/src/views/projects/Issue.svelte
@@ -4,6 +4,7 @@
   import type { State } from "@app/lib/issue";
 
   import Authorship from "@app/components/Authorship.svelte";
+  import Badge from "@app/components/Badge.svelte";
   import Button from "@app/components/Button.svelte";
   import Comment from "@app/components/Comment.svelte";
   import IssueSidebar from "@app/views/projects/Issue/IssueSidebar.svelte";
@@ -11,7 +12,7 @@
   import Reply from "@app/components/Comment/Reply.svelte";
   import TextInput from "@app/components/TextInput.svelte";
   import Textarea from "@app/components/Textarea.svelte";
-  import { capitalize, isLocal } from "@app/lib/utils";
+  import { isLocal } from "@app/lib/utils";
   import { createAddRemoveArrays, Issue } from "@app/lib/issue";
   import { formatObjectId } from "@app/lib/cobs";
 
@@ -156,18 +157,6 @@
     gap: 1rem;
     border-radius: var(--border-radius);
   }
-  .label {
-    padding: 0.3rem 0.5rem;
-    border-radius: var(--border-radius-round);
-  }
-  .open {
-    color: var(--color-positive);
-    background-color: var(--color-positive-2);
-  }
-  .closed {
-    color: var(--color-negative);
-    background-color: var(--color-negative-2);
-  }
   .reply {
     margin-left: 3rem;
   }
@@ -227,19 +216,16 @@
       </Button>
     </div>
     <div class="summary-state">
-      <div
-        class="label"
-        class:closed={issue.state.status === "closed"}
-        class:open={issue.state.status === "open"}>
-        {#if issue.state.status === "open"}
-          {capitalize(issue.state.status)}
-        {:else}
-          {capitalize(issue.state.status)}
-          <span class="txt-tiny">
-            {issue.state.reason}
-          </span>
-        {/if}
-      </div>
+      {#if issue.state.status === "open"}
+        <Badge variant="positive">
+          {issue.state.status}
+        </Badge>
+      {:else}
+        <Badge variant="negative">
+          {issue.state.status} as
+          {issue.state.reason}
+        </Badge>
+      {/if}
       <Authorship
         timestamp={issue.timestamp}
         author={issue.author}
```

rudolfs:
I find it a bit weird that the layout is different between an issue preview and the actual issue, could we align that? This would also solve the layout looking off in mobile.


Screenshot 2023-03-14 at 16 30 15
Screenshot 2023-03-14 at 16 30 19
Screenshot 2023-03-14 at 16 33 36

rudolfs:
The reply button still seems out of place for me. How about something like this?


https://user-images.githubusercontent.com/158411/225062815-4174693f-eff9-4bd9-aae3-5879a49e2c59.mov

```diff
diff --git a/src/components/Comment.svelte b/src/components/Comment.svelte
index 31edc1f7..1c486b4f 100644
--- a/src/components/Comment.svelte
+++ b/src/components/Comment.svelte
@@ -2,6 +2,7 @@
   import type { Comment } from "@app/lib/cobs";
 
   import Authorship from "@app/components/Authorship.svelte";
+  import Button from "@app/components/Button.svelte";
   import Icon from "@app/components/Icon.svelte";
   import Markdown from "@app/components/Markdown.svelte";
 
@@ -26,6 +27,7 @@
     align-items: center;
     justify-content: space-between;
     padding: 0.5rem 1rem;
+    height: 3rem;
   }
   .card-body {
     font-size: var(--font-size-small);
@@ -34,14 +36,11 @@
   }
   .actions {
     display: flex;
-    flex-direction: row;
-    align-items: center;
-    gap: 0.5rem;
-    color: var(--color-foreground-6);
+    justify-content: flex-end;
   }
   .action {
-    padding-top: 1rem;
-    cursor: pointer;
+    display: flex;
+    gap: 0.5rem;
   }
 </style>
 
@@ -52,6 +51,16 @@
         {caption}
         author={comment.author}
         timestamp={comment.timestamp} />
+      <div class="actions">
+        {#if allowReply}
+          <Button variant="text" size="tiny" on:click>
+            <div class="action">
+              <Icon name="chat" />
+              <span padding-left="1rem">reply</span>
+            </div>
+          </Button>
+        {/if}
+      </div>
     </div>
     <div class="card-body">
       {#if comment.body.trim() === ""}
@@ -59,13 +68,6 @@
       {:else}
         <Markdown {rawPath} content={comment.body} />
       {/if}
-      <div class="actions">
-        {#if allowReply}
-          <div class="action">
-            <Icon name="chat" on:click />
-          </div>
-        {/if}
-      </div>
     </div>
   </div>
 </div>
diff --git a/src/components/Comment/Reply.svelte b/src/components/Comment/Reply.svelte
index 16844654..ef314c65 100644
--- a/src/components/Comment/Reply.svelte
+++ b/src/components/Comment/Reply.svelte
@@ -21,7 +21,8 @@
     display: flex;
     flex-direction: row;
     justify-content: flex-end;
-    margin: 1rem 0;
+    margin: 1rem 0 2.5rem 0;
+    gap: 1rem;
   }
   .reply {
     margin-left: 3rem;
@@ -31,12 +32,20 @@
 <Comment
   {comment}
   {rawPath}
-  {allowReply}
+  allowReply={allowReply && !showReply}
   on:click={() => (showReply = !showReply)} />
 {#if showReply}
   <div class="reply">
     <Textarea resizable bind:value={reply} placeholder="Leave your reply" />
     <div class="actions">
+      <Button
+        variant="text"
+        size="small"
+        on:click={() => {
+          showReply = false;
+        }}>
+        Dismiss
+      </Button>
       <Button
         variant="secondary"
         size="small"
```

sebastinez:
Yeah I think we will be moving these around quite a lot in the future, but we can go with your solution.

src/views/projects/Issue.svelte
```diff
@@ -1,169 +1,241 @@
...
+  import IssueSidebar from "@app/views/projects/Issue/IssueSidebar.svelte";
+  import IssueStateButton from "@app/views/projects/Issue/IssueStateButton.svelte";
+  import Textarea from "@app/components/Textarea.svelte";
+  import { createAddRemoveArrays, Issue } from "@app/lib/issue";
+  import { isLocal } from "@app/lib/utils";
 
   export let issue: Issue;
   export let project: Project;
-</script>
+  export let session: Session | undefined;
```
rudolfs, line 17: I wonder if we should not just import session directly since it's a global thing in the app?

src/views/projects/Issue/New.svelte
```diff
@@ -4,69 +4,38 @@
...
-  import TextInput from "@app/components/TextInput.svelte";
-  import Textarea from "@app/components/Textarea.svelte";
+  import IssueSidebar from "@app/views/projects/Issue/IssueSidebar.svelte";
+  import IssueHeader from "@app/views/projects/Issue/IssueHeader.svelte";
   import { Issue } from "@app/lib/issue";
-  import { formatNodeId, parseNodeId } from "@app/lib/utils";
   import { createEventDispatcher } from "svelte";
+  import { stripDidPrefix } from "@app/lib/cobs";
 
   export let session: Session;
```
rudolfs, line 15: Could also just import the global session store.

src/views/projects/Issue.svelte
```diff
@@ -1,169 +1,241 @@
...
-      </div>
-      <div
-        class="summary-state"
-        class:closed={issue.state.status === "closed"}
-        class:open={issue.state.status === "open"}>
-        {capitalize(issue.state.status)}
-      </div>
-    </div>
-    <Authorship
+  <section>
```
rudolfs, line 149: What's the thinking behind using a custom tag vs just a div?

src/views/projects/Issue/New.svelte
```diff
@@ -164,32 +85,21 @@
...
-        </div>
-      {:else}
-        <Textarea bind:value={issueTitle} placeholder="Title" />
-        <Textarea
-          bind:value={issueText}
-          resizable
-          placeholder="Leave a comment" />
-      {/if}
+      <IssueHeader
+        id=""
```
rudolfs, line 89: Could we change the `id` type to `string | undefined` instead of passing and empty string here?

src/components/Button.svelte
```diff
@@ -34,6 +35,11 @@
   button:not([disabled]):hover {
     color: var(--color-background);
   }
+  .attachable {
```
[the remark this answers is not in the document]
sebastinez: I don't really like passing style as props around, but we can make an exception and if needed more in the future make it a class

src/views/projects/Issue.svelte
```diff
@@ -102,25 +100,21 @@
...
+    }, []);
 
-  // Scrolls into root level comment or scrolls to the reply textarea
-  afterUpdate(() => {
-    if (focusElement) {
-      scrollIntoView(focusElement, { behavior: "smooth", block: "center" });
-    }
-    focusElement = undefined;
-  });
+  $: console.log(threads);
```
rudolfs, line 115: 👀

tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "target": "es2020",
     "module": "es2020",
     "types": ["svelte", "vite/client"],
+    "lib": ["ES2021.String", "DOM"],
```
sebastinez, line 9: This is needed to let Typescript know of `String.prototype.replaceAll` which has been introduced in ES2021

PR #632, opened 2023-03-09
httpd-client api library

rudolfs:
We want some type safety when interacting with the API and we want to decouple the API code from the domain models that we have in the app. For now it'll live in this repository, but eventually we could move it out, it could live next to the httpd code in heartwood.


httpd-client/lib/projects.ts
```diff
@@ -0,0 +1,267 @@
...
+      },
+      projectsSchema,
+    );
+  }
+
+  public async getById(id: string, options?: RequestOptions): Promise<Project> {
+    return this.fetcher.fetchOk(
+      {
+        method: "GET",
+        path: `v1/projects/${id}`,
```
cloudhead, line 160: Wouldn't it make sense for the `v1/` part to be configured in the `fetcher`?

rudolfs: Yeah, probably. Still WIP.

rudolfs: Um, I think I tried. But the problem was that the root API call is without the v1. Maybe that has to change in the httpd backend?

rudolfs: If we stick this endpoint: https://seed.radicle.xyz/api under https://seed.radicle.xyz/api/v1, then it would all check out.

cloudhead: That's a bug yes :) cc @xphoniex

xphoniex: does this https://github.com/radicle-dev/heartwood/pull/409 fix it?

cloudhead: Merged.

rudolfs: db896aeecb8f63c165ffa2b1fa51935815f4983f

httpd-client/index.ts
```diff
@@ -0,0 +1,136 @@
...
+});
+
+export class HttpdClient {
+  private fetcher: Fetcher;
+  #baseUrl: string;
+
+  public project: project.Client;
+  public session: session.Client;
+
+  public constructor(baseUrl: string | BaseUrl, fetch?: BaseFetch) {
```
geigerzaehler, line 92: You can simplify the API if you require `baseUrl` to be of type `BaseUrl`. From what I can tell the string type is only really used as a shortcut and it’s easy to implement the shortcut at the call site. I think in general it’s desirable to keep a library like this as straightforward as possible.

src/lib/session.ts
```diff
@@ -1,40 +1,30 @@
...
-const endpoint = "http://localhost:8080/api/v1/sessions";
+const api = new HttpdClient("http://localhost:8080");
 
 export async function authenticate(params: {
   id: string;
   signature: string;
   publicKey: string;
-}): Promise<"success" | "failure"> {
+}): Promise<boolean> {
   disconnect();
```
geigerzaehler, line 20: There’s an `await` missing here which may lead to race conditions.

src/lib/session.ts
```diff
@@ -1,40 +1,30 @@
...
-  if (request.ok) {
+  try {
+    await api.session.create(params.id, params.signature, params.publicKey);
     save(params.id, params.publicKey);
-    return "success";
-  } else {
-    return "failure";
+    return true;
+  } catch {
+    return false;
```
geigerzaehler, line 27: This is probably an intentional shortcut but I want to highlight that you’re not disambiguating errors here. For example, if you try to authenticate but the server is not reachable you return `false`. I don’t yet know how this is used but this may lead to situations where you tell the user that their credentials are not valid when in fact the server has crashed.

src/lib/issue.ts
```diff
@@ -79,7 +79,7 @@ export class Issue {
     description: string,
     assignees: string[],
     tags: string[],
-    host: Host,
+    host: BaseUrl,
```
geigerzaehler, line 82: I would suggest aligning variable name and type name to have just one term for this thing.

src/views/projects/Browser.svelte
```diff
@@ -41,13 +46,17 @@
...
     }
 
     const promise =
-      path === "/" ? project.getReadme(commit) : project.getBlob(commit, path);
+      path === "/"
+        ? api.project.getReadme(project.id, commit)
+        : api.project.getBlob(project.id, commit, path);
 
     state = { status: Status.Loading, path };
     state = { status: Status.Loaded, path, blob: await promise };
```
geigerzaehler, line 62: Unrelated to this PR but these lines were really hard for me to understand. I think they obscure what is going on here. May I suggest the following:
```typescript
state = { status: Status.Loading, path };

let blob 
if  (path === "/") {
  blob = await api.project.getReadme(project.id, commit)
} else {
  blob = await api.project.getBlob(project.id, commit, path);
}

state = { status: Status.Loaded, path, blob };
return blob;
```

httpd-client/tests/client.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, expect, test } from "vitest";
+import { HttpdClient } from "../index";
+
+const client = new HttpdClient("http://0.0.0.0:8080");
```
geigerzaehler, line 4: I find it odd that this IP works. Shouldn’t it be `http://127.0.0.1:8080`?

httpd-client/tests/client.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe("client", () => {
+  test("#getRoot()", async () => {
+    await expect(client.getRoot()).resolves.not.toThrow();
```
geigerzaehler, line 8: You can also write this as
```suggestion
    await client.getRoot()
```
This tests the same thing though the test failure message will be different.

geigerzaehler (commented):
I think the direction of this is great and I have no suggestions for the design of this. (Also because I don’t know the architecture of the project very well.) The client library should be unopinionated and minimal only focusing on fetching and parsing and this is pretty much the case.

I just have a couple of smaller suggestions and observations.

httpd-client/tests/client.test.ts
rudolfs: Correct: https://github.com/radicle-dev/radicle-interface/pull/664.

httpd-client/index.ts
```diff
@@ -0,0 +1,136 @@
...
+export interface Node {
+  id: string;
+}
+
+const nodeSchema: ZodSchema<Node> = strictObject({
+  id: string(),
+});
+
+export class HttpdClient {
+  private fetcher: Fetcher;
```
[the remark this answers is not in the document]
rudolfs: e2287f5559f20c14ac9ad7be6c731a73bbca9d7e

httpd-client/tests/client.test.ts
rudolfs: Yeah, you're right. I think I had it that way when I started out but then decided to have the `expect` there just so it looks like a spec. Reverted back: 8305cd53fee4cc2d56dd1e3387f421de358205b3.

src/lib/session.ts
rudolfs: Good catch! abd623bdf15aab941d042c59e2c3331f9bb8a6a0

Was there a way to catch missing `await`s with eslint or something like that?

httpd-client/lib/project/commit.ts
```diff
@@ -0,0 +1,171 @@
...
+  line: string;
+  lineNo: number;
+  type: "deletion";
+}
+
+const deletionHunkLineSchema: ZodSchema<DeletionHunkLine> = strictObject({
+  line: string(),
+  lineNo: number(),
+  type: literal("deletion"),
+});
```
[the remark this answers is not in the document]
rudolfs: Nice, yeah, I think I also bumped into the union issue somewhere and just decided to omit the type annotation there. With `satisfies` I could apply this everywhere. 92b109a7687e54ffba97e8e37a302c08f4df554f

httpd-client/lib/fetcher.ts
```diff
@@ -0,0 +1,176 @@
...
+    query,
+    headers = {},
+  }: FetchParams): Promise<RawResponse> {
+    if (body !== undefined) {
+      headers["content-type"] = "application/json";
+    }
+
+    let url = `${this.baseUrl}/api/v1/${path}`;
+    if (query) {
+      url = `${url}?${qs.stringify(query)}`;
```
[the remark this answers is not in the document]
rudolfs: 609c404c8f30b24802fabdb0df9aefb82d90369c

But, ugh:
- https://github.com/microsoft/TypeScript/issues/32951
- https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1202

httpd-client/index.ts
rudolfs: bbcd2ab5a38caba1408d7f864b4fe853456b1199

src/lib/issue.ts
rudolfs: 216607835f223c450f8ebb6d736c9cbb377aea23
d7bcbed24e49afdc82c634f9bdf10adc99eb5545

src/views/projects/Browser.svelte
rudolfs: 471cf129735ba24a85f776c521cced9faf477686

src/lib/session.ts
geigerzaehler: There is a [no-floating-promises rule](https://typescript-eslint.io/rules/no-floating-promises/) that catches these issues. But you do need to enable type support for Eslint for this to work

httpd-client/lib/project/issue.ts
```diff
@@ -0,0 +1,60 @@
...
+import type { ZodSchema } from "zod";
+import { array, boolean, literal, strictObject, string, union } from "zod";
+
+import { commentSchema } from "./comment.js";
+
+export type IssueState =
+  | { status: "open" }
+  | { status: "closed"; reason: "other" | "solved" };
+
+const issueStateSchema = union([
```
geigerzaehler, line 11: This file defines schemas but does not define fetching functions. As an alternative you could stick with the hierarchical, resource-oriented approach that you followed for projects and sessions. That is, add a `Client` here that defines issue related method and assign an instance of `Client` to an `issues` property of the project client. Then you would get all issues with `client.project.issues.getAll()`.

I think this approach makes the code more modular because it keeps related things closer together and also more consistent. The downside is that it makes the code also more hierarchical and requires more indirection. (The same observation also applies to commits, comments and patches)

src/components/Dropdown.svelte
```diff
@@ -8,16 +8,14 @@
...
 
 <script lang="ts" strictEvents>
-  import type { State } from "@app/lib/cobs";
-
   import { createEventDispatcher } from "svelte";
   import { twemoji } from "@app/lib/utils";
   import Badge from "@app/components/Badge.svelte";
 
-  type T = $$Generic<State | string | number>;
+  type T = $$Generic;
```
geigerzaehler, line 15: :exploding_head:

rudolfs: lol 🚬 🤡

src/lib/session.ts
rudolfs: https://github.com/radicle-dev/radicle-interface/pull/674

src/views/projects/Header.svelte
```diff
@@ -128,15 +130,15 @@
     active={activeRoute.params.view.resource === "issues"}
     clickable
     on:click={() => toggleContent("issues", false)}>
-    <span class="txt-bold">{project.issues.open ?? 0}</span>
+    <span class="txt-bold">project.issues.open</span>
```
sebastinez, line 133:
```suggestion
    <span class="txt-bold">{project.issues.open}</span>
```

src/views/projects/Header.svelte
```diff
@@ -128,15 +130,15 @@
...
     {pluralize("issue", project.issues.open)}
   </HeaderToggleLabel>
   <HeaderToggleLabel
     ariaLabel="Patch count"
     active={activeRoute.params.view.resource === "patches"}
     clickable
     on:click={() => toggleContent("patches", false)}>
-    <span class="txt-bold">{project.patches.open ?? 0}</span>
+    <span class="txt-bold">project.patches.open</span>
```
sebastinez, line 141:
```suggestion
    <span class="txt-bold">{project.patches.open}</span>
```

sebastinez (commented):
Ok first of all again congrats on this beast of refactoring!
I left some minor nitpicks, also tested it locally and works great.
Will do a review over some of the endpoints.

src/views/projects/Header.svelte
rudolfs: Good catch!
ed68a64b12bad8eb1888b1734ae1f565e9348771

src/views/projects/Header.svelte
rudolfs: ed68a64b12bad8eb1888b1734ae1f565e9348771

httpd-client/lib/fetcher.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    if (!response.ok) {
+      throw new ResponseError(params.method, response, responseBody);
+    }
+
+    const result = schema.safeParse(responseBody);
+    if (result.success) {
+      return result.data;
+    } else {
+      console.log(JSON.stringify(result.error.errors, null, 2));
```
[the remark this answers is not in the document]
rudolfs: 31c82ec27ba3424dfa1c71ba1c2e963c28485274

httpd-client/index.ts
```diff
@@ -0,0 +1,138 @@
...
+  node: { id: string };
+  path: string;
+  links: { href: string; rel: string; type: Method }[];
+}
+
+const rootSchema = strictObject({
+  message: string(),
+  service: string(),
+  version: string(),
+  node: strictObject({ id: string() }),
```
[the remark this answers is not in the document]
rudolfs: Not sure I follow, the `/` endpoint returns
```
...
node: { id: string }
...
```
and the `/stats` endpoint returns
```
{
  projects: { count: number };
  users: { count: number };
}
```
There's no overlap in those types.

.github/workflows/check-httpd-api-unit-test.yml
```diff
@@ -0,0 +1,25 @@
...
+      - master
+
+jobs:
+  check-httpd-api-unit-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: "18.12.1"
```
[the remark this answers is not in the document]
rudolfs: Yep. 1458649879d511144cfb87df0fedd78df7300fbf

httpd-client/index.ts
```diff
@@ -0,0 +1,138 @@
+import type { BaseFetch, BaseUrl } from "./lib/fetcher";
+import type { RequestOptions, Method } from "./lib/fetcher";
+import type { ZodSchema } from "zod";
+
```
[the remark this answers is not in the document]
rudolfs: 7025bb297aaa249fd0a0c270fe66c47884c89916

src/views/projects/Header.svelte
```diff
@@ -1,23 +1,25 @@
 <script lang="ts">
-  import type { Project } from "@app/lib/project";
-  import type { Tree } from "@app/lib/project";
+  import type { BaseUrl, Project, Remote, Tree } from "@httpd-client";
   import type { ProjectRoute } from "@app/lib/router/definitions";
 
   import * as router from "@app/lib/router";
+
```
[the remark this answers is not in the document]
rudolfs: No other reason than personal preference, it helps me read the code faster. I chunk the imports by categories:
- libraries and helpers
- global reusable components
- single use components

If you want I can revert this back to how it was.

httpd-client/index.ts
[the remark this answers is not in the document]
sebastinez: Sorry I got them mixed up I meant the `nodeSchema`

src/views/projects/Header.svelte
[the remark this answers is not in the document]
sebastinez: No I think it's fine, we should then apply it gradually to all files

httpd-client/index.ts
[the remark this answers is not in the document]
rudolfs: Ah, yeah, totally. d8bd97e1b4468893a33f10b4ea4f4ec8f314ce77

httpd-client/lib/project/patch.ts
```diff
@@ -0,0 +1,150 @@
...
+  description: string(),
+  state: patchStateSchema,
+  target: string(),
+  tags: array(string()),
+  revisions: array(revisionSchema),
+}) satisfies ZodSchema<Patch>;
+
+export const patchesSchema = array(patchSchema) satisfies ZodSchema<Patch[]>;
+
+export type PatchUpdateAction =
```
sebastinez, line 136: It isn't yet implemented but it already works on `radicle-httpd` so I would implement the `Patch::Action::Redact`

It should look like the following
```
| { type: "redact"; revision: string }
```

src/components/Dropdown.svelte
sebastinez: Nothing better than to be explicit right? heeh

httpd-client/lib/fetcher.ts
```diff
@@ -0,0 +1,159 @@
...
+      try {
+        responseBody = JSON.parse(responseBody);
+      } catch (_e: unknown) {
+        // We keep the original text response body.
+      }
+      throw new ResponseError(params.method, response, responseBody);
+    }
+  }
+
+  private async fetch({
```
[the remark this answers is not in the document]
rudolfs: Hmm, if I understand the code correctly, it's a different problem. We're not really fetching the `rawPath` anywhere, we just supply it so that markdown can render some images. But I think it doesn't need to in all cases, like the `rawPath` prop should be optional and used depending on context. I don't think it makes sense to have it enabled for comments and issue/patch descriptions. It should only do image replacements for markdown files.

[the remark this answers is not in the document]
rudolfs: I think we should look at it, but in a different PR.

httpd-client/lib/project/patch.ts
```diff
@@ -0,0 +1,150 @@
...
+  | { type: "lifecycle"; state: PatchState }
+  | {
+      type: "review";
+      revision: string;
+      verdict: Verdict;
+      comment: string;
+      inline: CodeComment;
+    }
+  | { type: "merge"; revision: string; commit: string }
+  | { type: "thread"; action: ThreadUpdateAction };
```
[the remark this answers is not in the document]
rudolfs: 2b3f45fe07d08091799fea90622e2a2d6d949f07

httpd-client/lib/project/patch.ts
```diff
@@ -0,0 +1,150 @@
...
+  | { type: "edit"; title: string; description: string; target: string }
+  | { type: "editRevision"; revision: string; description: string }
+  | { type: "tag"; add: string[]; remove: string[] }
+  | { type: "revision"; description: string; base: string; oid: string }
+  | { type: "lifecycle"; state: PatchState }
+  | {
+      type: "review";
+      revision: string;
+      verdict: Verdict;
+      comment: string;
```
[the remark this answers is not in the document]
rudolfs: 5c85d0d47baf7c25a717e73f95bbdae57ec23a9a

httpd-client/lib/project/patch.ts
rudolfs: e1159da8c5378e0ed7ebe727d4631e0ef7e70c89

httpd-client/lib/project/patch.ts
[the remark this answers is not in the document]
rudolfs: I think patch commenting was actually broken before: 4db135abdfa3cb835efbb090f09bafc724acfb8e 😅

httpd-client/lib/project/issue.ts
rudolfs: I also thought about this. The only reason I split out the types was to keep the `project.ts` file smaller. If we would nest the clients further, where we would put the project ID? Like would we do `client.project.issues.getById(projectId, issueId)`?

geigerzaehler: I hadn’t thought about this. Yes, you could have the project ID as an argument to the issue client methods. But this doesn’t read to nicely. With this in mind, I think the way that it is currently is fine.

src/lib/session.ts
rudolfs: Yeah, this is a shortcut for now. We'll improve on this in a follow up.

PR #634, opened 2023-03-10
Update seed fixture to latest heartwood commit

sebastinez:
This PR is currently using a seed based on this PR
https://github.com/radicle-dev/heartwood/pull/421

~~Once it gets merged in the heartwood repo I'll update the seed fixture once more and it should be fine~~


scripts/run-httpd-with-fixtures
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-REV=f1de61ad8897a845f2be49e3e2b2951bc678b678
+REV=aacacf872fceb6b051a940341a4ce71de10be0e8
```
rudolfs, line 4: Can we bump this to `646376828d0520035246ccbbfcb82127160cd184`?

tests/support/utils.ts
```diff
@@ -0,0 +1,33 @@
+import bs58 from "bs58";
```
rudolfs, line 1: Why are we copying these functions here? Can't we just use the ones in `lib/utils`?

rudolfs (commented):
Nice work!

rudolfs:
There also seems to be some unintentional visual diff breakage.

scripts/run-httpd-with-fixtures
sebastinez: Yeah I'm currently on it

tests/support/utils.ts
sebastinez: I copied the functions, since I think we should make use of them instead of using `substring` everywhere.
But I wasn't able to use `lib/utils` directly since we import `config.json` in `config.ts` which requires type assertions from node
https://v8.dev/features/import-assertions
Which means we need to target a higher module in our typescript config like `esnext`

So either we go back to use `substring` or maybe update the typescript module?

rudolfs: I'm fine with using either `substring` or just hardcoding the expected results as strings in tests.

tests/visual/project.spec.ts
```diff
@@ -10,7 +10,7 @@ test("source tree page", async ({ page }) => {
   await expect(page).toHaveScreenshot();
 });
 
-test("commits page", async ({ page }) => {
+test.only("commits page", async ({ page }) => {
```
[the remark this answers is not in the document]
sebastinez: Have to write a linter rule for avoiding `only` leftovers hehe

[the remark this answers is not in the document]
rudolfs: I think playwright would have caught it anyway:
```
  forbidOnly: !!process.env.CI,
```

sebastinez:
> There also seems to be some unintentional visual diff breakage.

Ok this is weird, I tried to fix the two visual tests that still keep failing.

Black square image shows up instead of being removed
relative-image-not-able-to-being-loaded-1-diff

A different color for `did:key:` 
commit-page-1-diff

And both work flawlessly locally, but I'm not able to have them pass in the CI..
Could you maybe try them out locally?

sebastinez:
> Also just realised that we'll need to deploy the latest httpd on the seed node before we can merge this.

Yeah that makes sense :+1:

sebastinez:
Okay the one visual diff that is left is intentional, since the commit of the httpd has changed in the seed view.

PR #638, opened 2023-03-13
Join check and check-format scripts into one

sebastinez:
We don't need two separate scripts for checking the repo linting and style status.
This PR joins `scripts/check` and `scripts/check-format` and allows with `scripts/check` or `npm run check` to run tsc, eslint, prettier and svelte-check in one go.


scripts/check
```diff
@@ -3,3 +3,5 @@ set -e
 
 npx tsc --noEmit
 npx svelte-check --tsconfig tsconfig.json
+npx eslint .
+npx prettier "**/*.@(ts|js|svelte|json|css|html|yml)" --ignore-path .gitignore --check
```
rudolfs, line 7: Missing newline.

rudolfs (approved):

rudolfs:
We should remove this requirement from GH once this is merged.

Screenshot 2023-03-14 at 10 20 36

PR #642, opened 2023-03-14
Show local seed page upon successful authentication

rudolfs:
In most cases the user will want to interact with the local seed, so it makes sense to show that page. If the auth is unsuccessful, we go to the home screen.


sebastinez (approved):
:house:

PR #648, opened 2023-03-21
Update dependencies

sebastinez:
What's on the tin.

One of the big improvements is @wooorm/starry-night update to 1.7.0 includes grammar changes we worked on, so svelte files finally look nice, and also rust had some improvements IMO.

## left (current highlighting) / right (new highlighting with this PR)
Bildschirm­foto 2023-03-21 um 17 14 06
Bildschirm­foto 2023-03-21 um 17 12 59


tsconfig.json
```diff
@@ -8,13 +8,13 @@
     "types": ["svelte", "vite/client"],
     "sourceMap": true,
     "baseUrl": "./",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
rudolfs, line 11: Just for my own understanding, why do we need to change this?

rudolfs (approved):
Pushed another update and re-added `@tsconfig/svelte` because it was fixed in the meantime. LGTM from my end. @sebastinez maybe have another look at my commit.

rudolfs:
> Pushed another update and re-added @tsconfig/svelte because it was fixed in the meantime. 

Never mind. It wasn't updated yet.

tsconfig.json
sebastinez: Its not explicitly needed but I wanted to make the change since according to the typescript 5.0 article it's a good fit if we use Vite or a similar bundler

> If you are using a modern bundler like Vite, esbuild, swc, Webpack, Parcel, and others that implement a hybrid lookup strategy, the new bundler option should be a good fit for you.

Source: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler

PR #656, opened 2023-03-30
Fix hash router build and test configuration

sebastinez:
Some work done on seeing how we can handle env variables in Svelte/Vite/Playwright to be able to build the web app with hash router

- Moves `HASH_ROUTING` from being a vite constant which is statically replaced during build and else put on the `window` object, to the `import.meta.env` object, which also get replaced by Vite during build.
- Separates Playwright into history router config and hash router config to be able to handle them on a config level instead of inside the hashRouter test


rudolfs (approved):
Nice work! 👏

PR #659, opened 2023-03-31
Pluralize values instead of parenthesis

### LINKED ISSUE #650: Singularise/pluralise words properly
I think it would look much nicer if we didn't add pluralisation in brackets, like we do here for example:
Screenshot 2023-03-28 at 11 05 56

There are good libraries for that, for example [this one](https://github.com/plurals/pluralize).

sebastinez:
Ok I think this is a nice way of doing it without using a library, and with a nice function signature.

Closes #650

### DIFF
--- a/tests/e2e/project.spec.ts
+++ b/tests/e2e/project.spec.ts
@@ -15,11 +15,15 @@ async function expectCounts(
   page: Page,
 ) {
   await expect(page.locator('role=button[name="Commit count"]')).toContainText(
-    `${params.commits} commit(s)`,
+    `${params.commits} ${params.commits === 1 ? "commit" : "commits"}`,
   );
   await expect(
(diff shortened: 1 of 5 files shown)


rudolfs (commented):
Pretty cool that it works on the type level. 🌟 
The tests need some fixing.

tests/e2e/project.spec.ts
```diff
@@ -15,11 +16,13 @@ async function expectCounts(
   page: Page,
 ) {
   await expect(page.locator('role=button[name="Commit count"]')).toContainText(
-    `${params.commits} commit(s)`,
+    `${params.commits} ${pluralize("commit", params.commits)}`,
```
rudolfs, line 19: We should just hardcode the expectation. Otherwise our test result depends on the implementation of this function and that is now what is under test, so we're mixing concerns.

PR #664, opened 2023-04-05
Fix mixed content warning when connecting to local node

rudolfs:
This should work when the app is deployed somewhere other than localhost.


sebastinez:
Nice! This seems to do the trick! Did a deploy here
https://eloquent-gumdrop-e18f34.netlify.app/seeds/radicle.local/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5

sebastinez (approved):

sebastinez:
But it seems that some tests that look for `https://0.0.0.0` are failing now

PR #674, opened 2023-04-11
Enable the @typescript-eslint/no-floating-promises rule

rudolfs:
This will help us catch forgotten `await`s and let's us avoid nasty race conditions.

The only drawback is that it's slower:

Without the rule:
```
./scripts/check  15.52s user 1.06s system 188% cpu 8.817 total
```

With the rule:
```
./scripts/check  26.91s user 3.14s system 184% cpu 16.307 total
```


sebastinez (approved):
Nice, I think we can take that bump in speed for avoiding race conditions

PR #676, opened 2023-04-11
Add cli tooling

sebastinez:
WIP


tests/support/support.ts
```diff
@@ -0,0 +1,112 @@
...
+  const cwd = Path.join(peer.checkoutPath, name);
+  await Fs.mkdir(cwd, { recursive: true });
+  await peer.git(["init", "--initial-branch", "master"], {
+    cwd,
+    env: gitOptions["alice"],
+  });
+  await peer.git(["commit", "--allow-empty", "--message", "initial commit"], {
+    cwd,
+    env: gitOptions["alice"],
+  });
```
geigerzaehler, line 29: As you can see here you manually have to pass `gitOptions` every time you interact with git for the peer. Instead you could pass the git options when you create the peer and then add the environment variables in `RadiclePeer#spawn()`.

tests/support/support.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export async function modifySourceBrowsingFixture(
+  peer: RadiclePeer,
+  projectName = "source-browsing",
+) {
+  const cwd = Path.join(peer.checkoutPath, projectName);
+  // @ts-expect-error Some weirdness with the type of `spawn`.
+  await peer
+    .spawn("echo", ["Updated readme"])
+    .pipeStdout(Path.join(cwd, "README.md"));
```
geigerzaehler, line 96: I’m not sure this does what it is supposed to. Instead you can just change the file using [`fsPromises.writeFile()`](https://nodejs.org/api/fs.html#fspromiseswritefilefile-data-options).

tests/support/globalTeardown.ts
```diff
@@ -0,0 +1,19 @@
...
+import { fileURLToPath } from "url";
+
+const filename = fileURLToPath(import.meta.url);
+const dir = dirname(filename);
+
+export default async function globalTeardown(): Promise<void> {
+  const tmp = Path.resolve(dir, "..", "./tmp");
+  await Fs.rm(Path.join(tmp, "copy"), { recursive: true, force: true });
+  await Fs.rm(Path.join(tmp, "home"), { recursive: true, force: true });
+  await Fs.rm(Path.join(tmp, "source-browsing"), {
```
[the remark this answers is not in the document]
sebastinez: Moved the clean up to the beginning of `globalSetup`.

tests/support/peerManager.ts
```diff
@@ -0,0 +1,230 @@
...
+    this.#outputLog = props.logFile;
+  }
+
+  public static async create({
+    dataPath,
+    name,
+    seed,
+    outputLog: logFile,
+  }: PeerManagerParams): Promise<RadiclePeer> {
+    const checkoutPath = Path.join(dataPath, "copy", name);
```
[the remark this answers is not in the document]
sebastinez: Nice idea, will apply that 👍

geigerzaehler (approved):

scripts/install-binaries
```diff
@@ -0,0 +1,42 @@
...
+BINARIES=(rad radicle-node radicle-httpd git-remote-rad)
+REPO_ROOT=$(git rev-parse --show-toplevel)
+BINARY_PATH=$REPO_ROOT/tests/tmp/bin/${REV:0:7}
+OS=$(uname)
+
+mkdir -p "$BINARY_PATH"
+
+for BINARY_NAME in "${BINARIES[@]}"
+do
+  # Handle different naming of binaries than required locally
```
rudolfs, line 15: I think we can improve this comment by saying here why we need `rad-debug` for tests rather than the production `rad`. Nit: I try to end any comment with a full stop everywhere.

scripts/install-binaries
```diff
@@ -0,0 +1,42 @@
...
+    chmod a+x "$BINARY_PATH/$BINARY_NAME"
+  fi
+done
+
+# Add binaries to corresponding path, depending on if CI or local
+if [ -n "$GITHUB_PATH" ]; then
+  echo "$BINARY_PATH" >> "$GITHUB_PATH"
+else
+  export PATH="$BINARY_PATH:$PATH"
+fi
```
rudolfs, line 42: As discussed, we should just print the PATH here and offer the user instructions how to source it manually. This output could be used by the workflow to pass it to the next step.

tests/e2e/clipboard.spec.ts
```diff
@@ -74,14 +75,12 @@ test("copy to clipboard", async ({ page, browserName, context }) => {
     );
   }
 
+  // TODO: Navigating to radicle.local with a non default port doesn't work currently
```
rudolfs, line 78: Are you going to follow up on this in this PR? If not, I'd remove it, I think we're already aware of this shortcoming.

.github/workflows/check-e2e.yml
```diff
@@ -36,10 +36,10 @@ jobs:
         if: steps.playwright-dep-cache.outputs.cache-hit != 'true'
         run: npx playwright install chromium firefox
 
-      - name: Start http-api test server
+      - name: Install binaries
         run: |
           mkdir -p tests/artifacts;
-          ./scripts/run-httpd-with-fixtures --non-interactive --download 2>&1 | tee tests/artifacts/httpd-${{ matrix.browser }}.log &
+          source ./scripts/install-binaries;
```
rudolfs, line 42: I'm a bit curious as to why this works. It seems to me that there's some unexpected side-effect that makes it work. But I guess it doesn't matter, because we'll pass the path via GITHUB_OUTPUT to the next step as discussed.

tests/support/process.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+  if (childProcess.stderr) {
+    const stderrPrefix = new LinePrefix(pref);
+    childProcess.stderr.pipe(stderrPrefix).pipe(output, { end: false });
+  }
+
+  return childProcess;
+}
+
+function makePrefix(label: string): string {
```
rudolfs, line 69: I think we can reuse stuff from the `logLabel` module here.

tests/support/globalSetup.ts
```diff
@@ -1,31 +1,105 @@
...
-
-  try {
-    const response = await fetch(`http://0.0.0.0:${seedPort}/api/v1`);
-    const data = await response.json();
-    nodeId = data.node.id;
-  } catch (err) {
-    console.error(err);
-    throw new Error(notRunningMessage);
-  }
+  console.log("Setting up global test environment");
```
rudolfs, line 20: I would prefer if we extracted this into a `createSeedFixture` module and then call it here.

tests/support/globalSetup.ts
```diff
@@ -1,31 +1,105 @@
...
+  console.log("Resetting environment");
+  const tmp = Path.resolve(dir, "..", "./tmp");
+  const projectName = "source-browsing";
+  const sourceBrowsingDir = Path.join(tmp, "repos", projectName);
+  await Fs.rm(Path.join(tmp, "peers"), { recursive: true, force: true });
```
rudolfs, line 25: I'm a bit weary about deleting stuff from a user's computer without any prompts. Especially because this is run as part of the test suite, it might be unexpected. Also if we have a bug in the code it could go horribly wrong. That is why in the original script we had a `--non-interactive` flag, which we used on CI and by default the script showed the user what would be deleted and prompted for a confirmation.

I don't have a clear idea how to do it here now that we moved it out of a shell script. But might be worth to think about it a bit more.

tests/support/globalSetup.ts
```diff
@@ -1,31 +1,105 @@
...
+  await bob.git(["push", "rad"]);
+}
 
-  if (nodeId !== seedRemote) {
-    const wrongSeedMessage = `The server on port ${seedPort} doesn't have the right fixtures.\n`;
-    throw new Error(wrongSeedMessage + notRunningMessage);
+// Assert that the `rad` CLI is installed and has the correct version.
+async function assertRadInstalled(): Promise<void> {
+  const result = await execa("rad", ["--version"]);
+  const versionConstraint = "rad 0.8.0 (a6a3290)";
```
rudolfs, line 98: As discussed, I think it would be good to have the heartwood version in a single place. Like for example at the top-lovel of the repo in `heartwood-version` and then have the SHA in there and source it in all places where we need the version.

rudolfs (commented):
Had a preliminary look. Will do a more in-depth review once you addressed the things we talked about.

scripts/install-binaries
sebastinez: Ok changed it to
```
# To provide deterministic Node and Repo IDs, we need a rad CLI compiled with the --debug flag.
```

.github/workflows/check-e2e.yml
sebastinez: In the end I wrote a small script to compute the current binaries folder based on the `heartwood-version` file.
So we can just run the `install-binaries` script and then pass the output of `get-binaries-folder` to `$GITHUB_OUTPUT` and export it to PATH in the required step

tests/support/globalSetup.ts
sebastinez: Yeah I concurr with you, will think on how we can get a prompt user before deletion of files

sebastinez: Ok I think I found a nice way, using the node `readline` interface.

sebastinez: Ok with 892fa0d we can check if we are running CI and if not, we show the user a prompt to confirm the deletion of the `repos` and `peers` folder.

scripts/install-binaries
rudolfs: Nice! I would suggest the following edits to the output:
```
$ ./scripts/install-binaries

Using revision a6a3290833f6da4a7785a426bed342df069ce47f

Downloading rad from https://files.radicle.xyz/a6a3290833f6da4a7785a426bed342df069ce47f/aarch64-apple-darwin/rad-debug
Downloading radicle-node from https://files.radicle.xyz/a6a3290833f6da4a7785a426bed342df069ce47f/aarch64-apple-darwin/radicle-node
Downloading radicle-httpd from https://files.radicle.xyz/a6a3290833f6da4a7785a426bed342df069ce47f/aarch64-apple-darwin/radicle-httpd
Downloading git-remote-rad from https://files.radicle.xyz/a6a3290833f6da4a7785a426bed342df069ce47f/aarch64-apple-darwin/git-remote-rad

⚠️  You already have a radicle bin folder in your PATH variable.

To ensure you are using the correct binaries when running tests, run this command:
 👉 export PATH=/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290:$PATH
```

- add a newline at the top and separate the warning message from the action with newlines
- print the full command that a user can just copy-paste

rudolfs:
This fails when I run it locally. I assume some of the GIT global variables aren't fully stubbed, because it seems like it's still trying to use my own key signing material when creating a commit in the test fixtures:

```
radicle-interface » ./scripts/install-binaries                                                                                                                                                                             sebastinez/add-cli-tooling
Using revision a6a3290833f6da4a7785a426bed342df069ce47f

✅ rad
✅ radicle-node
✅ radicle-httpd
✅ git-remote-rad

⚠️  You already have a radicle bin folder in your PATH variable.
To ensure you are using the correct binaries make sure you add the following path to your PATH variable:

 👉 /Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290
radicle-interface » which rad                                                                                                                                                                                              sebastinez/add-cli-tooling
/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290/rad
radicle-interface » which radicle-httpd                                                                                                                                                                                    sebastinez/add-cli-tooling
/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290/radicle-httpd
radicle-interface » npm run test:e2e -- tests/e2e/historyRouter.spec.ts --project chromium                                                                                                                                 sebastinez/add-cli-tooling

> test:e2e
> TZ='UTC' playwright test tests/e2e/historyRouter.spec.ts --project chromium

Setting up global test environment

This will irrevocably destroy the following directories:

/Users/rudolfs/work/radicle-interface/tests/tmp/peers
/Users/rudolfs/work/radicle-interface/tests/tmp/repos

Are you sure you want to continue? [yes/no]: yes
Done
Error: Command failed with exit code 128: git commit --message Update readme --date Mon Dec 21 14:00 2022 +0100
error: Load key "/var/folders/yk/1zh3lm1n03v18kmk4gx4wsf40000gn/T//.git_signing_key_tmp4jv4MG": invalid format?

fatal: failed to write commit object

   at ../support/fixtures.ts:236

  234 |   );
  235 |   await bob.git(["add", "README.md"]);
> 236 |   await bob.git([
      |   ^
  237 |     "commit",
  238 |     "--message",
  239 |     "Update readme",

    at makeError (/Users/rudolfs/work/radicle-interface/node_modules/execa/lib/error.js:59:11)
    at handlePromise (/Users/rudolfs/work/radicle-interface/node_modules/execa/index.js:124:26)
    at createSeedFixture (/Users/rudolfs/work/radicle-interface/tests/support/fixtures.ts:236:3)
    at globalSetup (/Users/rudolfs/work/radicle-interface/tests/support/globalSetup.ts:22:3)
    ```

File: scripts/install-binaries (hunk 6 above)
Comment (rudolfs) [reply scripts/install-binaries]: I also think we should not print the warning if the correct binaries are in the path. Right now I get:

```
radicle-interface » which rad
/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290/rad

radicle-interface » which radicle-node
/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290/radicle-node

radicle-interface » which radicle-httpd
/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290/radicle-httpd

radicle-interface » which git-remote-rad                                                                                                                                                                                   sebastinez/add-cli-tooling

/Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290/git-remote-rad

radicle-interface » ./scripts/install-binaries
Using revision a6a3290833f6da4a7785a426bed342df069ce47f

✅ rad
✅ radicle-node
✅ radicle-httpd
✅ git-remote-rad

⚠️  You already have a radicle bin folder in your PATH variable.
To ensure you are using the correct binaries make sure you add the following path to your PATH variable:

 👉 /Users/rudolfs/work/radicle-interface/tests/tmp/bin/a6a3290
```

File: .github/workflows/check-e2e.yml [hunk 13]
```diff
@@ -36,10 +36,11 @@ jobs:
         if: steps.playwright-dep-cache.outputs.cache-hit != 'true'
         run: npx playwright install chromium firefox
 
-      - name: Start http-api test server
+      - name: Install rad binaries
         run: |
           mkdir -p tests/artifacts;
-          ./scripts/run-httpd-with-fixtures --non-interactive --download 2>&1 | tee tests/artifacts/httpd-${{ matrix.browser }}.log &
+          ./scripts/install-binaries;
+          ./scripts/get-binaries-folder >> $GITHUB_PATH;
```
rudolfs, line 43: Could we maybe move this to the other script and hide the functionality behind a flag? I feel like both scripts have an implicit contract about how to construct the path, so it would be nice to have it all in one place.

Something like: `install-binaries --show-path`.

.github/workflows/check-e2e.yml
```diff
@@ -36,10 +36,11 @@ jobs:
         if: steps.playwright-dep-cache.outputs.cache-hit != 'true'
         run: npx playwright install chromium firefox
 
-      - name: Start http-api test server
+      - name: Install rad binaries
```
rudolfs, line 39:
```suggestion
      - name: Install Radicle binaries
```

tests/support/globalSetup.ts
```diff
@@ -1,31 +1,74 @@
...
-async function assertHttpApiRunning(): Promise<void> {
-  const notRunningMessage =
-    "The http-api server with test fixtures needs to be running.\n" +
-    "👉 You can start it with `./scripts/run-httpd-with-fixtures`\n";
+async function promptWorkspaceRemoval(): Promise<void> {
+  console.log("");
+  console.log("This will irrevocably destroy the following directories:");
+  console.log("");
+  console.log(Path.join(tmp, "peers"));
+  console.log(Path.join(tmp, "repos"));
```
rudolfs, line 43: We should print the actual paths that are to be deleted here, so we should either store them in a variable and pass that to `removeWorkspace()` or combine the two functions. Otherwise if a dev changes the `removeWorkspace()` function and makes a mistake, the user will get the wrong idea of what is getting deleted.

sebastinez:
> This fails when I run it locally. I assume some of the GIT global variables aren't fully stubbed, because it seems like it's still trying to use my own key signing material when creating a commit in the test fixtures:

Could you try again, it's possible that the `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_NOSYSTEM` were missing

.github/workflows/check-e2e.yml
sebastinez: One other reason I added this is so we get all the download logs, etc from the `install-binaries` script and then just need the `get-binaries-folder` one for the CI output..

So if we add a `--show-path` flag, this would need to be something like `--show-only-path` so we can pipe that path into `$GITHUB_PATH`

If we want to have the path construction in one place, maybe it would be better to make `install-binaries` query `get-binaries-folder` for the correct path instead of computing it by itself?

rudolfs:
> > This fails when I run it locally. I assume some of the GIT global variables aren't fully stubbed, because it seems like it's still trying to use my own key signing material when creating a commit in the test fixtures:
> 
> Could you try again with the latest commit, it's possible that the `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_NOSYSTEM` were missing

Ok, that worked.

rudolfs:
Now I wonder if we should somehow allow running the tests with a previously created seed? Cause the seed creation takes around 13s on my laptop. It's quite a lot to wait if I locally want to run a single test.

sebastinez:
> Now I wonder if we should somehow allow running the tests with a previously created seed? Cause the seed creation takes around 13s on my laptop. It's quite a lot to wait if I locally want to run a single test.

Yeah that's true, but maybe we can keep that for a follow up?
It's getting kinda big.

I'm thinking we could allow people to run the `globalSetup` function locally, since the tests by itself don't run the globalSetup

.github/workflows/check-e2e.yml
rudolfs: Can't we run it twice? Like:

```
./scripts/install-binaries
./scripts/install-binaries --show-path
```

Where the second invocation only prints the path, but doesn't download anything.

tests/support/globalSetup.ts
sebastinez: Done in 4c9466d

.github/workflows/check-e2e.yml
sebastinez: ah yeah that would work 👍

sebastinez: fixed in 3c9fbd7
Also added a small help

rudolfs:
> > Now I wonder if we should somehow allow running the tests with a previously created seed? Cause the seed creation takes around 13s on my laptop. It's quite a lot to wait if I locally want to run a single test.
> 
> Created an issue to track this #757

I was thinking of something very simple, like this:
```diff
diff --git a/tests/support/fixtures.ts b/tests/support/fixtures.ts
index 7e4714bf..e53a4bf5 100644
--- a/tests/support/fixtures.ts
+++ b/tests/support/fixtures.ts
@@ -179,6 +179,15 @@ export function appConfigWithFixture() {
   };
 }
 
+export async function startPalmHttpd() {
+  const peerManager = await createPeerManager({
+    dataDir: Path.resolve(dir, "..", "./tmp/peers"),
+    outputLog: FsSync.createWriteStream(Path.resolve(dir, "..", "./tmp/log")),
+  });
+  const palm = await peerManager.startPeer({ name: "palm" });
+  await palm.startHttpd();
+}
+
 export async function createSeedFixture() {
   const projectName = "source-browsing";
   const sourceBrowsingDir = Path.join(tmp, "repos", projectName);
diff --git a/tests/support/globalSetup.ts b/tests/support/globalSetup.ts
index f25f7a1c..56b84f0d 100644
--- a/tests/support/globalSetup.ts
+++ b/tests/support/globalSetup.ts
@@ -3,7 +3,7 @@ import type { FullConfig } from "@playwright/test";
 import * as Fs from "node:fs/promises";
 import * as Path from "node:path";
 import * as readline from "node:readline/promises";
-import { createSeedFixture, tmp } from "./fixtures.js";
+import { createSeedFixture, startPalmHttpd, tmp } from "./fixtures.js";
 import { execa } from "execa";
 
 const workspacePaths = [Path.join(tmp, "peers"), Path.join(tmp, "repos")];
@@ -16,12 +16,19 @@ export default async function globalSetup(_config: FullConfig): Promise<void> {
     process.exit(1);
   }
 
-  console.log("Setting up global test environment");
-  if (!process.env.CI) {
-    await promptWorkspaceRemoval();
+  if (!process.env.SKIP_FIXTURE_CREATION) {
+    console.log("Setting up global test environment");
+    if (!process.env.CI) {
+      await promptWorkspaceRemoval();
+    }
+    await removeWorkspace();
+
+    console.log("Creating seed fixture");
+    await createSeedFixture();
+    console.log("Running tests");
+  } else {
+    await startPalmHttpd();
   }
-  await removeWorkspace();
-  await createSeedFixture();
 }
 
 // Assert that the `rad` CLI is installed and has the correct version.

```

sebastinez:
> I was thinking of something very simple, like this:

Ok added it with 08bf121

heartwood-version
```diff
@@ -0,0 +1 @@
+a6a3290833f6da4a7785a426bed342df069ce47f
```
rudolfs, line 1: Could we move this to `tests/support/heartwood-version`? Feel like it belongs there instead of the top-level, since it's only used by tests.

scripts/install-binaries
sebastinez: Ok think I found a good way with 9fc3807

sebastinez: Which returns if `$BINARY_PATH` is found in PATH
```

Using revision a6a3290833f6da4a7785a426bed342df069ce47f

✅ rad
✅ radicle-node
✅ radicle-httpd
✅ git-remote-rad

✅ /Users/sebastinez/Projects/radicle-interface/tests/tmp/bin/a6a3290 is already in your PATH variable.
```

scripts/install-binaries
```diff
@@ -0,0 +1,90 @@
...
+RADICLE_HOME=${RAD_HOME:-"${HOME}/.radicle/bin"}
+
+# Add a separator between binaries download and PATH instructions.
+echo
+
+if [[ ":$PATH:" != *":$BINARY_PATH:"* ]]; then
+  if [[ ":$PATH:" == *":$RADICLE_HOME:"* ]]; then
+    echo "⚠️  You already have a radicle bin folder in your PATH variable."
+    echo
+    echo "To ensure you are using the correct binaries make sure you add the following path to your PATH variable:"
```
rudolfs, line 79:
```suggestion
    echo "To ensure you are using the correct test binaries, run the following command:"
```

scripts/install-binaries
```diff
@@ -0,0 +1,90 @@
...
+  else
+    echo "Before running the tests, make sure you add the following path to your PATH variable:"
```
rudolfs, line 81:
```suggestion
    echo "Before running the tests, make sure your PATH variable is set up correctly by running:"
```

heartwood-version
sebastinez: Yeah was my first move too, but the top-level was also nice, since we have all the nvmrc and config files there

scripts/install-binaries
sebastinez: d4348e393266d900af52a8cea624bb32c13f0356

scripts/install-binaries
sebastinez: d4348e393266d900af52a8cea624bb32c13f0356

heartwood-version
sebastinez: c33c61048723439f0db577a8f05ced0c9d0474c2

PR #680, opened 2023-04-12
Refactor issue/patch assignee and tag inputs

rudolfs:
We remove the `CobInput` abstraction and just have two components. It simplifies the interfaces and allows us to have more control over the validation of the element array. Before it was possible to add assignees with the same `pubkey` with a `did` prefix and without and it was hard to work around it with the abstraction in place.

This also fixes some regressions introduced in https://github.com/radicle-dev/radicle-interface/pull/632.


sebastinez (approved):
Nice LGTM!

Issue #645, opened 2023-03-17
Relative markdown links don't seem to work


cloudhead:
Eg. try clicking on `CONTRIBUTING.md` in the README here:

https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5

PR #686, opened 2023-04-14
Fix relative markdown file links

### LINKED ISSUE #682: Relative markdown links don't work
Eg. a link like [README](README.md) links to the wrong place.

See here links in readme here: https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5

sebastinez:
This PR should close #682, #645

It adds a `.file-link` class to eventual relative file links inside markdowns that also aren't hash anchor links.
Then we are able in the `Markdown` component to query those elements and call `updateProjectRoute` with the new path

This is something I would like to write a e2e test about.. But I'll wait for CLI tooling, since it needs changes to the seed fixture and I think that can wait.

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -5,7 +5,8 @@
   import { onMount } from "svelte";
   import { toDom } from "hast-util-to-dom";
 
-  import { base } from "@app/lib/router";
+  import * as utils from "@app/lib/utils";
+  import { base, updateProjectRoute, activeRouteStore } from "@app/lib/router";
   import { isUrl, twemoji, scrollIntoView, canonicalize } from "@app/lib/utils";
   import { highlight } from "@app/lib/syntax";
   import {
@@ -29,6 +30,20 @@
   const render = (content: string): string =>
     // eslint-disable-next-line @typescript-eslint/naming-convention
     dompurify.sanitize(marked.parse(content), { SANITIZE_DOM: false });
+
+  function navigateToMarkdownLink(event: any) {
+    if (event.target.matches(".file-link")) {
+      event.preventDefault();
+      if ($activeRouteStore.resource === "projects") {
+        updateProjectRoute({
+          path: utils.canonicalize(
+            event.target.getAttribute("href"),
+            $activeRouteStore.params.path ?? "",
+          ),
+        });
+      }
+    }
+  }
 
   onMount(async () => {
     // Don't underline <a> tags that contain images.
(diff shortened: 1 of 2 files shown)


src/components/Markdown.svelte
```diff
@@ -30,6 +32,21 @@
     // eslint-disable-next-line @typescript-eslint/naming-convention
     dompurify.sanitize(marked.parse(content), { SANITIZE_DOM: false });
 
+  function handleMarkdownFileLinkNavigation(event: any) {
+    if (event.target.matches(".file-link")) {
+      event.preventDefault();
+      const route = get(activeRouteStore);
```
rudolfs, line 38: We can just use Svelte syntax here:
```diff
diff --git a/src/components/Markdown.svelte b/src/components/Markdown.svelte
index a8ba36c5..81c54171 100644
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -3,7 +3,6 @@
   import matter from "@radicle/gray-matter";
   import { marked } from "marked";
   import { onMount } from "svelte";
-  import { get } from "svelte/store";
   import { toDom } from "hast-util-to-dom";

   import * as utils from "@app/lib/utils";
@@ -35,7 +34,7 @@
   function handleMarkdownFileLinkNavigation(event: any) {
     if (event.target.matches(".file-link")) {
       event.preventDefault();
-      const route = get(activeRouteStore);
+      const route = $activeRouteStore;
       if (route.resource === "projects") {
         updateProjectRoute({
           path: utils.canonicalize(
```

sebastinez: Ah you're right, good catch 👍

src/components/Markdown.svelte
```diff
@@ -79,6 +101,15 @@
     }
 
     await Promise.allSettled(treeChanges);
+
+    return () => {
```
rudolfs, line 105: I think we don't need to clean up anything here, because it wouldn't work on an `async` function anyway:

> This behaviour will only work when the function passed to onMount synchronously returns a value. async functions always return a Promise, and as such cannot synchronously return a function.

and there wouldn't be anything to click on when the component is destroyed, so I think any click handlers would be cleaned up automatically.

src/components/Markdown.svelte
```diff
@@ -30,6 +32,21 @@
     // eslint-disable-next-line @typescript-eslint/naming-convention
     dompurify.sanitize(marked.parse(content), { SANITIZE_DOM: false });
 
+  function handleMarkdownFileLinkNavigation(event: any) {
```
rudolfs, line 35: I prefer it when we callback name is describing the action, so in this case it would be `navigateToMarkdownLink`.

src/components/Markdown.svelte
sebastinez: Yeah you're right!

rudolfs (approved):

PR #695, opened 2023-04-18
Fetch cobs by selected or default state

sebastinez:
Currently we don't query by state, which means we get patches of different states and by paginating them we eventually loose patches that should be shown by selecting their state.

Created a patch on heartwood for the `radicle-httpd` part with id `2aaaf04d2c1335b0db41f7d401d18de6d29292a6` but isn't being replicated yet on seed.radicle.xyz.

### DIFF
--- a/src/views/projects/View.svelte
+++ b/src/views/projects/View.svelte
@@ -235,19 +235,11 @@
           </div>
         {/if}
       {:else if activeRoute.params.view.resource === "issues"}
-        {#await api.project.getAllIssues(project.id)}
-          <Loading center />
-        {:then issues}
-          <Issues
-            {baseUrl}
-            issueCounters={project.issues}
-            status={issueFilter}
-            {issues} />
-        {:catch e}
-          <div class="message">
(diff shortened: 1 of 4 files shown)


src/views/projects/View.svelte
```diff
@@ -35,6 +36,7 @@
   $: issueFilter = (searchParams.get("state") as IssueStatus) || "open";
   $: patchTabFilter =
     (searchParams.get("tab") as "activity" | "commits") || "activity";
+  $: console.log(patchFilter);
```
rudolfs, line 39: 👀

rudolfs:
LGTM. Let's merge this once the backend is deployed.

sebastinez:
So I finally created a PR since my radicle patch didn't get to seed.radicle.xyz 
https://github.com/radicle-dev/heartwood/pull/534 let's wait for it be merged and deployed

src/views/projects/View.svelte
sebastinez: a393d0b

rudolfs (approved):

PR #696, opened 2023-04-18
Refactor project list fetching

rudolfs:
We lift the loading of data to the top, to prepare for moving it to the router altogether at some point. Simplify the `Widget` -> `ProjectCard` component by moving any loading out of it and passing down only the necessary props. Also refactor the error message component and add a copy error stack trace to clipboard button.


src/components/ProjectCard.svelte
```diff
@@ -0,0 +1,139 @@
...
+    <div class="right">
+      <div class="id">
+        <span class="rid layout-desktop">{id}</span>
+      </div>
+      <div class="layout-desktop activity">
+        <Diagram
+          points={activity}
+          strokeWidth={3}
+          viewBoxHeight={100}
+          viewBoxWidth={600} />
```
sebastinez, line 135: I feel like these params were nice when testing this feature, but since we use this Diagram only in one place maybe we can hardcode them? rename the component to `ActivityDiagram` or something and rename the `points` prop to `activity`?

src/views/seeds/View.svelte
```diff
@@ -18,121 +22,177 @@
...
+        <div class="more">
+          <Loading small />
+        </div>
+      {/if}
+      {#if showMoreButton}
+        <div class="more">
+          <Button variant="foreground" on:click={loadProjects}>More</Button>
+        </div>
+      {/if}
+    {/if}
```
sebastinez, line 191: Why don't we use the `List` component here, so we don't have to rewrite the display elements for fetching more, loading, etc. for every list?

sebastinez (commented):
Nice work, just left two minor comments

src/views/seeds/View.svelte
rudolfs: I think the `List` component abstraction is making things harder for us and we should not use it. For one it adds more indirection, so it is harder to follow the code. Then it also encapsulates data loading into the component tree, making it harder to control when and how the loading happens.

I think the architecture we should aim for is to move most data loading to the top-level "view" component or possibly later even to the router. Then pass down only the data that the components actually need. This gives us certain benefits:
  - it's easier to reason about more complex component structures, because we know that the components don't have any side effects, they only receive data via props and show it
   - data fetching happens in one place, so we don't have to look around for it, but we also have a more fine-grained control over how the loading happens, this we can leverage to chunk requests into parallel groups where it makes sense

geigerzaehler: I agree with the approach to data loading that @rudolfs outlined. In my experience it makes the data side of things less complex and more robust. With that in mind it makes sense to not use `List` here and deprecate its use in general.

I’d also like to expand a bit on this because I feel the `List` component is a good example for the hidden complexity if you spread out data loading. If you look at `List` you can see that the value of `items` can be changed by the component itself through user input, which is the main purpose of the component. But `items` can also be changed by the parent component because it is a property. Now its easy to imagine a case where the `items` is constructed dynamically in an ancestor component (e.g. through `.map()` or `.filter()`) and so `items` may be recreated as a different object depending on some other state. In that case, the `List` component would re-render but loose all the additional items that were fetched.

This behavior is acknowledged in `projects/History.svelte` where `List` is used with a two-way binding `bind:items={history}`. But `history` itself is only bound one way where `History` is used and actually the result of a `.map()`. Now we’re already two components up in the tree and what we write here may (or may not) create loss of state in `List`. These are the surprising and hard to debug issues that are best avoided.

One other thing I’d like to point out is that `List` is only used once right now (in `History`). It can be a common pitfall to abstract things prematurely only to find out later that the abstraction creates more problems than it solves. I think this is probably the case here for a three reasons:
1. If we remove the abstraction and inline `List` into `History`, `History` is not harder to understand (and arguably easier to understand)
2. `List` does not have a simple, easy to understand contract with the parent component. (I mentioned all the subtle behavior above.)
3. `List` is only used in one place.

As a rule of thumb it would delay abstracting data management components until a clear pattern for its use emerges.

sebastinez: Okay message received. I'll then avoid using `List` in the patch and issue listing 😁

sebastinez (approved):
🚢 it

PR #697, opened 2023-04-18
Improve click behaviour code highlighting

### LINKED ISSUE #692: Highlight code line is a bit buggy
If you try to highlight a line of code in a file, it seems like it only works if you click right in the middle. Even though the line number is highlighted, the click doesn't necessarily go through..

sebastinez:
Removes the padding from the `td` parent, and gives it to the clickable anchor tag.

Closes #692

### DIFF
--- a/src/views/projects/Blob.svelte
+++ b/src/views/projects/Blob.svelte
@@ -125,9 +125,12 @@
   .code :global(.line-number) {
     color: var(--color-foreground-4);
     text-align: right;
-    padding-right: 1rem;
-    padding-left: 1rem;
+    padding: 0;
     user-select: none;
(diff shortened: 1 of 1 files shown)


rudolfs:
There still seems to be an area that is not covered, i.e. the cursor changes to a hand indicating that the underlying element is clickable, but it's not.

https://user-images.githubusercontent.com/158411/232835892-d9cc0b77-0cda-4d2e-8f4e-ba07c85048f3.mov

sebastinez:
Ok the last commit, did achieve to cover the entire area without having to rewrite the line gutter

rudolfs (approved):

PR #698, opened 2023-04-19
Simplify Authorship component

sebastinez:
The `Authorship` component has taken too many responsibilities, so I renamed it to `Author` and just let it show an optional `Avatar` and its `NodeId`.
In the view and components that used `Authorship` I replaced the caption and timestamp with `span` elements.

Does not make the codebase much smaller, but makes it easier to understand and handle Authorship information


src/components/Author.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+  import Avatar from "@app/components/Avatar.svelte";
+  import { formatNodeId } from "@app/lib/utils";
+
+  export let nodeId: string;
+  export let noAvatar: boolean = false;
+</script>
+
+<style>
+  .authorship {
```
rudolfs, line 10: I'd rename all the `.authorship` classes to `.author`.

src/components/Author.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+  import Avatar from "@app/components/Avatar.svelte";
+  import { formatNodeId } from "@app/lib/utils";
+
+  export let nodeId: string;
+  export let noAvatar: boolean = false;
```
rudolfs, line 6: Doesn't look like we're using this prop anywhere in the callsites. Can we drop it?

src/components/Author.svelte
```diff
@@ -0,0 +1,30 @@
...
+    white-space: nowrap;
+  }
+</style>
+
+<span class="author txt-tiny">
+  <Avatar inline {nodeId} />
+  <span class="id layout-desktop">
+    {formatNodeId(nodeId)}
+  </span>
+  <span class="id layout-mobile">
```
rudolfs, line 27: One thing I've noticed that we're doing is mixing the layout concern into components. Ideally components should be "dumb" and only concerned with their own state and visuals. Any kind of component positioning should happen from outside components, i.e. with a wrapper at the call site. And any kind of modification to how a component looks should come through its props. This should theoretically make it easier to reason about visual changes when refactoring code and making changes in general, because stuff isn't coupled.

I don't have a clear idea how on how we should tackle layouts and responsivity yet so we don't need to change anything here now, just some food for thought.

rudolfs:
Seems like there are some unintended visual changes (btw. we really need those tests):
- Before:
  Screenshot 2023-04-19 at 11 52 57
- After:
  Screenshot 2023-04-19 at 11 53 00

- Before:
  Screenshot 2023-04-19 at 11 53 21

- After:
  Screenshot 2023-04-19 at 11 53 18

src/components/Author.svelte
sebastinez: I agree with your reasoning, I think one way would be to pass some of the responsivness to js, so we enable or disable some properties on a component according to the viewport width.
In the beginning we solely wanted to limit the styling to css, but maybe revisit this decision, or add additional responsivness through js, and if no js is available or not working fallback to a basic css layout.

sebastinez:
> btw. we really need those tests

Yeah I agree, I'm still trying to befriend my radicle-node but he doesn't really want to talk with me hehe

src/components/Author.svelte
rudolfs: Hm, not sure JS is the answer either. It's would be just a different way to do the same thing. There's something that bothers me about using the global layout classes, but I can't put my finger on exactly what it is. I think there's something getting covered up that shouldn't, I also see a lot of `!important` statements in the whole global layout system which makes me a bit uneasy.

I think I have to do a bit more research on how other frameworks handle this to have a better opinion.

geigerzaehler: > Any kind of component positioning should happen from outside components, i.e. with a wrapper at the call site.

I agree with this. However, I think the story here is more ambiguous: `layout-mobile` and `layout-desktop` don’t control positioning—they control what is shown based on viewport size. (The class names reinforce that confusion and should probably be changed to `display-mobile-only` and `display-desktop-only`.) It’s reasonable to assume that the component knows best what to show based on the globally available space. 

On the other hand it may turn out that the parent component should control what is being shown because it knows best how much space is actually available. 

> And any kind of modification to how a component looks should come through its props.

> In the beginning we solely wanted to limit the styling to css, but maybe revisit this decision, or add additional responsivness through js, and if no js is available or not working fallback to a basic css layout.

It may be tricky to get this to function properly without a lot of work and complexity. As an alternative to Svelte properties to customize styling you could consider adding a “CSS selector API” to your component. This is something that is pretty [common for Web Components](https://meowni.ca/posts/part-theme-explainer/) and provided by the [`::part()`](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) selector. Basically the idea is to give some elements (parts) in your component a name so that parent components can select these elements and modify their styles through CSS.

For example in this case you could have
```svelte
<span class="id">
  <span part="prefix">did:key:</span>{formatNodeId(nodeId).replace("did:key:", "")}
</span>
```

and in the parent write
```svelte
<style>
  @media (max-width: 720px) {
   .author [part="prefix"] { display: none   }
  }
</style>

<Author class="author" ... />
```

(Note that you cant use the `::part()` selector because it only works with the shadow DOM. Also I haven’t tested this so you may need to play around with it and figure out how to use it in Svelte.)

The important thing is that you treat the styles of any element with a `part` attribute as part of the public API of the component.

> There's something that bothers me about using the global layout classes, but I can't put my finger on exactly what it is.

I think in this case a global CSS class is a fine choice. It’s a meaningful abstraction for well-defined, reusable behavior that is used by multiple independent components. Often something else (like a component) is a better choice than CSS to create abstraction and isolation. But in this case it is fine.

> I also see a lot of `!important` statements in the whole global layout system which makes me a bit uneasy.

Yes, this is general a smell. But this can be avoided by structuring the classes a bit differently. For example:
```css
@media (max-width: 720px) {
  .hide-on-desktop {
    display: none;
  }
}

@media (min-width: 721px) {
  .hide-on-mobile {
    display: none;
  }
}
```
With these classes you can let the components that use the class define the default display value when the element is shown but still have it hidden depending on the viewport.

rudolfs: > As an alternative to Svelte properties to customize styling you could consider adding a “CSS selector API” to your component.

In svelte we can pass down CSS variables like this:

```html
<!-- Component -->
<style>
.box {
  width: var(--size);
  height: var(--size);
  }
</style>
<div class="box" />


<!-- Callsite -->
<Component --size="10rem" />
```

sebastinez:
Thanks for the feedback I think I'll do another try at this but from a fresh HEAD

Issue #699, opened 2023-04-19
Support basic markdown in issue/patch titles


cloudhead:
It would be nice if we rendered markdown inside titles like github does. Eg. this patch: https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/issues/1cb7cafa3fcf81536afe903a94e8f87b96e2a983

... The inline-code could be rendered as markdown.

PR #702, opened 2023-04-20
Allow markdown in issue/patch titles

rudolfs:
Closes https://github.com/radicle-dev/radicle-interface/issues/699.

### DIFF
--- a/src/views/projects/Cob/CobHeader.svelte
+++ b/src/views/projects/Cob/CobHeader.svelte
@@ -5,6 +5,7 @@
 
   import Button from "@app/components/Button.svelte";
   import Clipboard from "@app/components/Clipboard.svelte";
+  import InlineMarkdown from "@app/components/InlineMarkdown.svelte";
   import TextInput from "@app/components/TextInput.svelte";
 
   export let action: "create" | "edit" | "view" = "view";
@@ -67,7 +68,9 @@
         <TextInput transparent variant="form" bind:value={title} />
       {:else}
         {#if title}
-          <span class="txt-medium title">{title}</span>
+          <div class="title">
+            <InlineMarkdown fontSize="medium" content={title} />
+          </div>
         {:else}
           <span class="txt-missing">No title</span>
         {/if}
--- a/src/components/InlineMarkdown.svelte
+++ b/src/components/InlineMarkdown.svelte
@@ -1,0 +1,34 @@
+<script lang="ts">
+  import dompurify from "dompurify";
+  import { marked } from "marked";
+
+  import { renderer } from "@app/lib/markdown";
+  import { twemoji } from "@app/lib/utils";
+
+  export let content: string;
+  export let fontSize: "small" | "medium" = "small";
+
+  marked.use({ renderer });
+
+  const render = (content: string): string =>
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    dompurify.sanitize(marked.parseInline(content), { SANITIZE_DOM: false });
+</script>
+
+<style>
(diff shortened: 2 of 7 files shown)


sebastinez (commented):
Seems to check out, the only thing that bothers me a bit is that the monospace, depending on the font one uses is so much smaller than the regular font.
Not sure if this is how it should be or if we could/should do something about it.

Bildschirmfoto 2023-04-20 um 08 58 44

GitHub seems to increase the font-size or make it more equal
Bildschirmfoto 2023-04-20 um 08 59 59

rudolfs:
@sebastinez good points. I was also wondering whether the title markdown shouldn't be restricted to only `backticks`. In that case maybe we can roll some special handling just for that.

cloudhead:
Yeah we should definitely make sure the font sizes match up everywhere. If we want to support a subset of markdown, I'd say we support the inline styles: code, em, strong, strikethrough.

rudolfs:
> Yeah we should definitely make sure the font sizes match up everywhere.

You mean the markdown in the title or adjust the code block font size for all markdown. Just asking because I remember making it a notch smaller at some point after you said it looked too big :)

cloudhead:
> You mean the markdown in the issue/patch titles or adjust the code block font size for all markdown everywhere. Just asking because I remember making it a notch smaller at some point after you said it looked too big :)

I think it looks good in code blocks that are in the body now, it's just the titles where it looks small.

src/views/projects/Cob/CobHeader.svelte
```diff
@@ -67,7 +68,11 @@
         <TextInput transparent variant="form" bind:value={title} />
       {:else}
         {#if title}
-          <span class="txt-medium title">{title}</span>
+          <div class="title">
+            <InlineMarkdown
+              --font-size="var(--font-size-medium)"
```
sebastinez, line 73: Isn't there any way to type this prop?
I tried some things but wasn't successful..

rudolfs: Not really. We could encapsulate this in a regular prop, like we do elsewhere.

E.g.

`export let size: "small" | "regular" = "regular"`

And then just add the usual CSS in the component.

rudolfs: Will do that, so I don't introduce another way of doing things that's not aligned with the rest.

rudolfs: cc2aab77926cb423f3b70df39cda9a3e90a44196

sebastinez: Nice, I like the idea of passing the variables directly, but without type safety I feel like we're introducing new potential regressions.

sebastinez (approved):

src/views/projects/Cob/CobHeader.svelte
rudolfs: Hmmmmmm, just realised that it's not the same thing actually.
When we pass down the CSS variable, we use it in the `<script>` section with the `:global(code)` modifier. So basically it's not possible to achieve the same effect. Because applying it to the whole element changes the size of all the text.

rudolfs: Screenshot 2023-04-27 at 16 55 21
Screenshot 2023-04-27 at 16 56 10

sebastinez: Ok gotcha, I'm okay with 175b513 will be on the lookout if something changes with typed props to support CSS variables maybe we can circle back here

Issue #378, opened 2022-09-09
Allow using middle mouse button to open a link in a new tab


rudolfs:
A common web usage pattern is to open links for later reading in separate tabs by clicking the middle mouse button or holding command/ctrl and left clicking. This currently doesn't work, because we're only handling the `on:click` event. We could potentially use also the `on:auxclick` event, but unfortunately that [doesn't work on Safari](https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event#browser_compatibility).

If we're considering switching out the router, we should also think about how we could enable this functionality.

One potential way to pull this off it is to only use anchor tags (`<a href="#xxx>"`) and hash routing for navigation elements that are clickable. Then we wouldn't need to bother manually handling click events.

Here are a couple of areas in the app that I've tried middle-clicking out of habit and wondered why it didn't work:

<img width="1840" alt="Screenshot 2022-09-09 at 16 17 14" src="https://user-images.githubusercontent.com/158411/189371843-3f194e85-b7d5-497e-838a-27a802c5b2cc.png">
<img width="1840" alt="Screenshot 2022-09-09 at 16 17 30" src="https://user-images.githubusercontent.com/158411/189371866-28bfdd00-6bb5-41e0-9e39-560f72d8a04d.png">
<img width="1840" alt="Screenshot 2022-09-09 at 16 17 35" src="https://user-images.githubusercontent.com/158411/189371883-745444d4-4bf7-45db-a499-da2ef39f0409.png">
<img width="1840" alt="Screenshot 2022-09-09 at 16 19 36" src="https://user-images.githubusercontent.com/158411/189371890-9d9fd547-bbc5-4aa4-b4e5-9fb33667707a.png">

sebastinez:
Interesting, I think that the hash routing would surely be a good way to go forward, since this would also allow us to host the web interface on IPFS.

Was initially thinking about providing hash routing only when needed e.g. in IPFS deployments, but if we could provide middle mouse button clicking by going for hash routing it would make sense.

Not entirely sure if I'm missing something that the current way of routing provides (besides the url schema) that we would miss.

sebastinez:
Just for reference, hash routing is being worked on at #478
We should remember that this could be solved by using hash routing

PR #704, opened 2023-04-21
Allow opening links in a new tab

rudolfs:
Closes: https://github.com/radicle-dev/radicle-interface/issues/378


sebastinez (commented):
Looks really fine! Like how the `ProjectLink` component reduces to `on:click => updateProject` cases.
The only thing we should be careful is to not add other links inside elements that are wrapped by `ProjectLink` due to

> Content model:
Transparent, but there must be no interactive content descend, a element descendant, or descendant with the tabindex attribute specified.

Source: https://html.spec.whatwg.org/#the-a-element

rudolfs:
Here are some selected screenshots of the visual tweaks:

Screenshot 2023-04-27 at 13 04 19
Screenshot 2023-04-27 at 13 04 37
Screenshot 2023-04-27 at 13 04 44
Screenshot 2023-04-27 at 13 05 03
Screenshot 2023-04-27 at 13 05 19
Screenshot 2023-04-27 at 13 06 00
Screenshot 2023-04-27 at 13 06 14
Screenshot 2023-04-27 at 13 06 19

src/views/projects/Header.svelte
```diff
@@ -82,67 +37,99 @@
...
+  <Link
+    route={baseUrl.port !== config.seeds.defaultHttpdPort
+      ? {
+          resource: "seeds",
+          params: { hostnamePort: `${baseUrl.hostname}:${baseUrl.port}` },
+        }
+      : {
+          resource: "seeds",
+          params: { hostnamePort: baseUrl.hostname },
+        }}>
```
geigerzaehler, line 75:
```suggestion
    route={{
          resource: "seeds",
          params: { hostnamePort: baseUrl.port === config.seeds.defaultHttpdPort ? baseUrl.hostname : `${baseUrl.hostname}:${baseUrl.port}` },
        }}>
```

src/views/projects/Patches.svelte
```diff
@@ -118,47 +107,62 @@
 
 <div class="patches">
   <div style="margin-bottom: 1rem;">
-    <TabBar {options} on:select={switchState} active={state} />
+    <div style="display: flex; gap: 0.5rem;">
+      {#each options as option}
+        {#if !option.disabled}
```
geigerzaehler, line 112: You can switch the body of the if statement around and make the check `#if option.disabled`. This is easier to read.

geigerzaehler (commented):
This is a niece change overall. I haven’t looked at changes to state management in the components to deeply. If you think that’s important I can take a closer look.

I think this PR is too big which made it hard to review. I would suggest to split PRs like this up into meaningful chunks. For example, you could create a PR that introduces `Link` and `ProjectLink` and uses it in a couple of places, then a PR that removes `TabBar`, then PRs that change component responsibilities like PeerSelector or state management. Finally a PR for all the style changes.

For me, 200 lines is already an upper bound for change size. If the changes are very repetitive a higher number is ok, though.

rudolfs:
> For me, 200 lines is already an upper bound for change size. If the changes are very repetitive a higher number is ok, though.

Yeah, you're right, I just kept cramming all the stuff in here. Will make smaller PRs from now on.

src/views/projects/Patches.svelte
rudolfs: f09d68094e03b82194194eb87f8c9d67509c0431

src/views/projects/Header.svelte
rudolfs: 3bb3ca2ce8bca051301ee8052eab55ddeb5bb763

sebastinez (approved):
Nice cleaning action 💪 
Great work

EDIT: adding to what @geigerzaehler said, I have to agree it was kinda hard to review all the different changes.

PR #724, opened 2023-04-28
Update deps

rudolfs:
🐿️ 🎡


sebastinez (approved):
Looks good to me, if you want to you can update it to the latest and greatest or merge it like this

rudolfs:
> Looks good to me, if you want to you can update it to the latest and greatest or merge it like this

I updated the deps again. PTAL.

I didn't update the node types to v20, because that seems wrong and doesn't make sense. We're using node v18. Also didn't update the marked v4 -> v5 as it requires a bit more work to get rid of the deprecation warnings.

sebastinez (approved):
Ok nice! LGTM
I opened a issue for updating marked #731

PR #735, opened 2023-05-07
Add `noUnusedLocals` and `noUnusedParameters` to `tsconfig`

### LINKED ISSUE #732: `scripts/check` doesn't detect unused variables
I've noticed that my editor still detects these via the language server, but svelte check stopped detecting them. We should fix this so that we catch unused stuff via CI.

Screenshot 2023-05-05 at 15 05 49

Screenshot 2023-05-05 at 15 05 56

sebastinez:
This PR fixes #732, with adding the `noUnusedLocals` and `noUnusedParameters` option to `tsconfig`

It seems that once we added the `--tsconfig` param to `svelte-check` in f215f9f4fd54587fa0e5ba21f1289314e6e2a954 we overrode some defaults which included the checking for unused local variables.

So I added some checks to tsconfig

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,6 +11,8 @@
     "moduleResolution": "node",
     "strict": true,
     "resolveJsonModule": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
     "useDefineForClassFields": true,
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,


rudolfs:
> It seems that once we added the --tsconfig param to svelte-check in https://github.com/radicle-dev/radicle-interface/commit/f215f9f4fd54587fa0e5ba21f1289314e6e2a954 we overrode some defaults which included the checking for unused local variables.

Gotcha. Thanks for the digging! 🙇

rudolfs (approved):

PR #737, opened 2023-05-08
Add hr before first footnote in footer

### LINKED ISSUE #734: Markdown footer needs margin
See here, when there are footnotes:
image

We should add a `<hr>` when there's footnotes, and give it perhaps 1rem of margin on each side.

sebastinez:
This PR closes #734 

This PR adds a `<hr>` element before the first footnote in the footer if there is one.
And the styling of a `<hr>` element has currently a margin of 1rem on each side.

Bildschirmfoto 2023-05-08 um 10 12 05

### DIFF
--- a/src/lib/markdown.ts
+++ b/src/lib/markdown.ts
@@ -85,7 +85,9 @@ const footnoteMarkedExtension = {
     }
   },
   renderer: (token: marked.Tokens.Generic) => {
-    return `<p class="txt-small" id="${footnotePrefix}:${token.reference}">${
+    return `${
+      token.reference === "0" ? "<hr />" : ""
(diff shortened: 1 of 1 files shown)


rudolfs (approved):
Hm, would be cool if we could add a markdown doc with references to the seed fixture, so we have visual specs for this. If it's a lot of work we can also create an issue and do it as a follow up. Will leave it up to you.

sebastinez:
> Hm, would be cool if we could add a markdown doc with references to the seed fixture, so we have visual specs for this. If it's a lot of work we can also create an issue and do it as a follow up. Will leave it up to you.

Yeah I would like to keep that for a follow-up since adam is about to have a rad debug binary ready, and then we can generate our fixtures with the CLI tooling and do the visual checks for this

PR #741, opened 2023-05-08
Fix wrong icon used for iOS icon

sebastinez:
Closes #738
 
Fixes usage of wrong asset, but I feel like the "alien" is now very small.. wdyt @rudolfs ?


rudolfs:
> Fixes usage of wrong asset, but I feel like the "alien" is now very small.. wdyt @rudolfs ?

Try the one that Brandon provided in https://github.com/radicle-dev/radicle-interface/issues/738.

rudolfs (approved):

Issue #643, opened 2023-03-15
Small markdown spacing issue


cloudhead:
I noticed a small spacing issue when lists have paragraphs/code under them. See here:
![image](https://user-images.githubusercontent.com/40774/225269259-065c6385-c3cf-4825-b5f7-bc9636b87d51.png)

We should probably have `1rem` of spacing there, and same at the top, even if a code block isn't used.

URL: https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z2Jk1mNqyX7AjT4K83jJW9vQoHn4f/tree/master/README.md

Issue #733, opened 2023-05-06
Markdown issue with list line-breaks


cloudhead:
Notice the hard line-breaks in the list items here, that are unusual:

<img width="628" alt="image" src="https://user-images.githubusercontent.com/40774/236631938-0c11e140-615e-4248-8a78-d6c43d7d99d6.png">

See here: https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3trNYnLWS11cJWC6BbxDs5niGo82/tree/master/0003-storage-layout.md#overview

The markdown looks like this:

```
1. The storage layer is capable of maintaining a local copy of the working
   dataset.
2. The storage layer can store any number of repositories.
3. For each repository, it can represent multiple views, or *forks*, of
   the repository.
4. The storage layer can natively interoperate with Git.
```

PR #742, opened 2023-05-09
Fix padding and linebreaks in markdown

sebastinez:
Ok it seems I introduced a regression with #643 that now impacts in list items with soft line breaks.
So I changed the implementation removing that renderer and keep the padding as it was and just removing margin when there is only a paragraph without other elements.

image
image

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -187,7 +187,7 @@
     margin-bottom: 0.625rem;
   }
 
-  .markdown :global(p:last-child) {
+  .markdown :global(p:only-child) {
     margin-bottom: 0;
   }
 
@@ -282,7 +282,7 @@
     margin: 1rem 0;
   }
   /* Allows the parent to specify its own bottom margin */
-  .markdown :global(:last-child) {
+  .markdown :global(> :last-child) {
     margin-bottom: 0;
   }
   .markdown :global(li > ul) {
--- a/src/lib/markdown.ts
+++ b/src/lib/markdown.ts
@@ -132,17 +132,6 @@ export const renderer = {
 
     return `<h${level} id="${escapedText}">${text}</h${level}>`;
   },
-  listitem(text: string) {
-    const hasLineBreaks = text.trim().indexOf("\n");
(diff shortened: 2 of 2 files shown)


rudolfs (approved):
I don't think I understood what the actual problem was in https://github.com/radicle-dev/radicle-interface/issues/643, to me the spacing looked okay.

Now when I look at the same readme with [this change applied locally](http://localhost:3000/seeds/seed.radicle.xyz/rad:z2Jk1mNqyX7AjT4K83jJW9vQoHn4f/tree/master/README.md ) it seems like it still doesn't have the padding that #643 asked for, but [on production it looks the same way](https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z2Jk1mNqyX7AjT4K83jJW9vQoHn4f/tree/master/README.md), so I don't know whether something needs to be done further here. 🤷 

However, I do see that this fixes the weird line breaks, so it's a ✅ from my side.

sebastinez:
Hmm ok it seems that this commit added a `margin-bottom: 0;` to all `last-child`s that overwrites the behaviour I was looking for with this PR.

I feel the `Markdown` component needs a style refactoring at some point 😅

sebastinez:
This improves the padding.
AFAI understood #643 the raw text has a separator line between the end of the bullet points and the next paragraph, but in Markdown we don't show a separation.
image

rudolfs:
Are the visual spec changes as you intended? There now seems to be always a padding at the bottom of all markdown files.

cloudhead:
Can you show a screenshot of the fix for #773?

sebastinez:
> Can you show a screenshot of the fix for #733?

Yeah sure

image

sebastinez:
> Are the visual spec changes as you intended? There now seems to be always a padding at the bottom of all markdown files.

The changes have been removed with the second commit.
But I think I'll add a visual test at some point for these line breaks.
Since the `Markdown` component is ending up handling a lot of different components markdown

cloudhead:
Brilliant

rudolfs (approved):

PR #751, opened 2023-05-11
Fix httpd client root query

rudolfs:
This fixes the request to not add the additional `/` which broke the seeds view https://app.radicle.xyz/seeds/seed.radicle.xyz.

We were querying `https://seed.radicle.xyz/api/v1/` instead of `https://seed.radicle.xyz/api/v1`. The previous backend version was happily accepting this, but that broke recently. I don't know why it broke now.

Screenshot 2023-05-11 at 18 01 27


sebastinez (approved):
Nice work! Solves it for me 👍

Issue #752, opened 2023-05-12
Replace "opened X minutes ago" with "updated X minutes ago" in patches


xphoniex:
So this patch was "opened 1 week ago" but "updated 3 minutes ago".

![Screenshot from 2023-05-11 17-04-47](https://github.com/radicle-dev/radicle-interface/assets/14107758/0d5c189e-37ec-4dc0-a8b2-95f759c88e14)

Either we show both, "updated 3 minutes ago, opened 1 week ago"
or just the latest action which would be "**updated** 3 minutes ago".

PR #754, opened 2023-05-12
Adjust action wording to show only latest action in patch list

rudolfs:
It will show either "Opened x days ago" or "Updated x days ago" depending whether there is one or multiple revisions.

Closes https://github.com/radicle-dev/radicle-interface/issues/752.

Screenshot 2023-05-12 at 06 52 55

### DIFF
--- a/src/views/projects/Patch/PatchTeaser.svelte
+++ b/src/views/projects/Patch/PatchTeaser.svelte
@@ -147,9 +147,9 @@
     </div>
     <div class="summary">
       <span class="meta id">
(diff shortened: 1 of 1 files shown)


xphoniex (approved):
lgtm, do we need to do the same for issues as well?

rudolfs:
> do we need to do the same for issues as well?

hmm, potentially. although I don't know how issue editing is supposed to work, so I'd tackle that when we have editing.

sebastinez (approved):

PR #756, opened 2023-05-12
Add `alias` to PatchTeaser

xphoniex:
Screenshot from 2023-05-12 09-29-24

### DIFF
--- a/src/components/Authorship.svelte
+++ b/src/components/Authorship.svelte
@@ -3,6 +3,7 @@
   import { formatNodeId, formatTimestamp } from "@app/lib/utils";
 
   export let authorId: string;
+  export let authorAlias: string | undefined = undefined;
   export let caption: string | undefined = undefined;
   export let noAvatar: boolean = false;
   export let timestamp: number | undefined = undefined;
@@ -27,6 +28,9 @@
   .body {
     white-space: nowrap;
   }
+  .alias {
+    color: var(--color-foreground-5);
+  }
 </style>
 
 <span class="authorship txt-tiny">
@@ -35,9 +39,15 @@
   {/if}
   <span class="id layout-desktop">
     {formatNodeId(authorId)}
+    {#if authorAlias}
+      <span class="alias">({authorAlias})</span>
+    {/if}
   </span>
   <span class="id layout-mobile">
     {formatNodeId(authorId).replace("did:key:", "")}
+    {#if authorAlias}
+      <span class="alias">({authorAlias})</span>
+    {/if}
   </span>
   {#if !caption}
     <slot />
--- a/httpd-client/lib/project/patch.ts
+++ b/httpd-client/lib/project/patch.ts
@@ -111,7 +111,7 @@ const revisionSchema = strictObject({
 
 export interface Patch {
   id: string;
-  author: { id: string };
+  author: { id: string; alias?: string };
   title: string;
   description: string;
   state: PatchState;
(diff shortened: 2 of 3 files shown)


src/components/Authorship.svelte
```diff
@@ -3,6 +3,7 @@
   import { formatNodeId, formatTimestamp } from "@app/lib/utils";
 
   export let authorId: string;
+  export let authorAlias: string;
```
rudolfs, line 6:
```suggestion
  export let authorAlias: string | undefined = undefined;
```

src/components/Authorship.svelte
```diff
@@ -35,9 +36,15 @@
...
   <span class="id layout-desktop">
     {formatNodeId(authorId)}
+    {@html authorAlias
+      ? '<span class="alias">(' + authorAlias + ")</span>"
+      : ""}
   </span>
   <span class="id layout-mobile">
     {formatNodeId(authorId).replace("did:key:", "")}
+    {@html authorAlias
+      ? '<span class="alias">(' + authorAlias + ")</span>"
```
xphoniex, line 46: when I add a styling for `"alias"` class it doesn't render, e.g.:
```
<style>
...
.alias {
  color: red !important;
}
</style>
```

any idea why?

src/components/Authorship.svelte
```diff
@@ -35,9 +36,15 @@
...
+      : ""}
```
rudolfs, line 47:
```suggestion
    {#if authorAlias}
    <span class="alias">{authorAlias}</span>
    {/if}
```

src/components/Authorship.svelte
rudolfs: Not sure, but could be because you're rendering with `@html`.

xphoniex: yes, works now without `@html`.

src/components/Authorship.svelte
```diff
@@ -27,6 +28,9 @@
   .body {
     white-space: nowrap;
   }
+  .alias {
+	  color: gray;
```
rudolfs, line 32:
```suggestion
  var(--color-foreground-6)
```

xphoniex: that's not the same color

xphoniex: `5` works

xphoniex: https://github.com/radicle-dev/radicle-interface/compare/647bf9c9d5eee91f89bd059d70aed86eb042f196..78a0d79c49f9110fb40997f0fbbca7040ddb5780

rudolfs: Then pick one that matches from our color system: https://github.com/radicle-dev/radicle-interface/blob/master/public/colors.css

rudolfs: You can see all available colors in dev mode by pressing <kbd>d</kbd> in the app.

Screenshot 2023-05-12 at 08 17 40

rudolfs:
How do you add an alias via the CLI?

xphoniex:
> How do you add an alias via the CLI?

```
$ rad track z6MksFqXN3Yhqk8pTJdUGLwATkRfQvwZXPqR2qMEhbS9wzpT --alias cloudhead
```

rudolfs:
I'm running `radicle-httpd` version 0.1.0-7037b52 locally. Cloned heartwood with `rad clone` and added a couple of aliases, but they don't show up in the patch list. What am I doing wrong?

xphoniex:
> I'm running `radicle-httpd` version 0.1.0-7037b52 locally. Cloned heartwood with `rad clone` and added a couple of aliases, but they don't show up in the patch list. What am I doing wrong?

you need to compile off this patch
```
7e9cb11  httpd: Add `alias` field to `/patches` and `/issues` routes
```

PR #760, opened 2023-05-15
Add issue and patch tests

sebastinez:


tests/support/fixtures.ts
```diff
@@ -229,15 +230,35 @@ export async function createSeedFixture() {
     "Git repository for source browsing tests",
     "--announce",
   ]);
+  await createIssue(
```
rudolfs, line 233: I know this is still a draft, but I was wondering whether we shouldn't already start a new fixture for the cob tests so it doesn't end up growing into one huge fixture?

sebastinez:
Closing this PR and will re open it based on #789

Issue #747, opened 2023-05-10
Navigation quirks


cloudhead:
I noticed two quirks in project navigation:

1. When navigating away from a patch page, the `?tab=files` query sticks around
2. It's possible to view patches/issues etc. with a remote selected. This doesn't make sense, as patches are loaded with all remotes always. So selecting a remote should go back to source browsing always. Having a remote selected should only work in the source code or commits tab, basically anything code related.

cloudhead:
Yes, I think that would make sense. I'm just not sure how it's going to look without some good design work :sweat_smile:

rudolfs:
I think it could work 🤷
You can play around with it here if you like: https://github.com/radicle-dev/radicle-interface/tree/rudolfs/header-hierarchy-mockup

<img width="1840" alt="Screenshot 2023-05-12 at 10 00 34" src="https://github.com/radicle-dev/radicle-interface/assets/158411/14deee0c-70d1-4478-8394-cda69132e70e">
<img width="1840" alt="Screenshot 2023-05-12 at 10 00 39" src="https://github.com/radicle-dev/radicle-interface/assets/158411/a4328f26-a2a9-4cfe-8364-3f112cefcb03">
<img width="1840" alt="Screenshot 2023-05-12 at 10 01 06" src="https://github.com/radicle-dev/radicle-interface/assets/158411/efedb5ab-c96a-4167-a9d6-247dc02b0a12">

cloudhead:
I would start like this, but still needs work eventually:
<img width="1480" alt="image" src="https://github.com/radicle-dev/radicle-interface/assets/40774/b3737b26-277a-4a18-935c-676e3ff8235d">

* Removed content bg
* Removed content padding
* Restored button size
* Added 2rem bottom margin to header

PR #761, opened 2023-05-15
Improve project header information hierarchy

rudolfs:
We split out tabs that are global to a project to the top level.

Refs https://github.com/radicle-dev/radicle-interface/issues/747.


rudolfs:
@cloudhead do we want the issue/patch filter buttons to also be the same size as the top-level menu items?

Screenshot 2023-05-15 at 10 46 28
Screenshot 2023-05-15 at 10 46 31

sebastinez:
Will be looking into the code, but I think due to setting to `undefined` the `peer`, `route` and `revision` properties, we get a hard refresh of the entire project view, at points where we shouldn't get it.
https://github.com/radicle-dev/radicle-interface/assets/7912302/de4a174b-6ea9-4447-811d-f3b8bdae6361

sebastinez:
Hmm also the issues and patches togge we had earlier is now a bit out of place.
I think I would prefer to disable the click once patches or issues listing is active
https://github.com/radicle-dev/radicle-interface/assets/7912302/615a2d0a-137c-4604-b6dc-075a4e0e27df

rudolfs:
> Hmm also the issues and patches toggle we had earlier is now a bit out of place. I think I would prefer to disable the click once patches or issues listing is active https://github.com/radicle-dev/radicle-interface/assets/7912302/615a2d0a-137c-4604-b6dc-075a4e0e27df

Fixed here: a2352d4b4c11ed217f110b10095c2c236aa6b25e.

rudolfs:
> Also it seems that we don't reset the peer view, when navigating to issues or patches.
> 
> > It's possible to view patches/issues etc. with a remote selected. This doesn't make sense, as patches are loaded with all remotes always. So selecting a remote should go back to source browsing always. Having a remote selected should only work in the source code or commits tab, basically anything code related.
> > [#747 (comment)](https://github.com/radicle-dev/radicle-interface/issues/747#issue-1704030387)
> 
>  Bildschirmaufnahme.2023-05-15.um.16.22.51.mov

Good catch. Fixed in b3b36cf887e5d1fb7360a9390c413df89ec1bdf3.

sebastinez (approved):
Nice, works great

Issue #665, opened 2023-04-05
Patch TabBar states shorten too much


sebastinez:
![image](https://user-images.githubusercontent.com/7912302/230114301-254bc4ba-0354-486c-a693-53e83577a6d0.jpeg)

sebastinez:
Ok this seems to be some responsivness issue that only shows up on my iPhone 14, not able to reproduce it on any browser, or responsivness simulation

Issue #720, opened 2023-04-28
Fix issue/patch mobile views


rudolfs:
<img width="389" alt="Screenshot 2023-04-28 at 12 07 12" src="https://user-images.githubusercontent.com/158411/235119731-1d702f55-4a77-4f35-94f3-5825fa1895ac.png">
<img width="391" alt="Screenshot 2023-04-28 at 12 07 23" src="https://user-images.githubusercontent.com/158411/235119734-31108650-b453-4291-a381-de3a6133771f.png">

PR #762, opened 2023-05-15
Make sure filter tabs are displayed properly on mobile

rudolfs:
We make sure that the space between the counter and wording doesn't wrap and then wrap the tabs themselves if they take more space than the viewport.

Closes: https://github.com/radicle-dev/radicle-interface/issues/665, https://github.com/radicle-dev/radicle-interface/issues/720

Screenshot 2023-05-15 at 11 28 51


sebastinez (approved):
📐

PR #763, opened 2023-05-15
Add more sleep to branch creations

sebastinez:
There seems to be some issue with the creation of the `orphaned-branch` in the seed fixture..
I added some sleeping, but the issue should be solved differently..


rudolfs (approved):
LGTM. We should follow this up with checking whether the data is available and removing any `sleep`s entirely.

PR #770, opened 2023-05-17
Update dependencies

sebastinez:
Also update node to v18.16


sebastinez:
Nice, yeah forgot about the transitive ones

PR #776, opened 2023-05-18
Reduce e2e tests flakiness

sebastinez:
This PR removes all of the sleep instructions from the seed fixture creation, and is able to rely on checking node events and node routing tables.

This allows the tests to await the necessary events and routing table entries instead of sleeping and hoping for them.


tests/support/peerManager.ts
```diff
@@ -89,6 +92,14 @@ export class RadiclePeer {
     this.#outputLog = props.logFile;
   }
 
+  public waitForEvent(searchEvent: NodeEvent, timeoutInMs: number) {
+    const start = new Date().getTime();
+
+    while (new Date().getTime() - start > timeoutInMs) {
+      this.#eventRecords.find(event => lodash.isEqual(searchEvent, event));
```
rudolfs, line 99: I'd add a 100ms sleep here, otherwise this will load the CPU 100%.

tests/support/peerManager.ts
```diff
@@ -164,9 +168,69 @@ export class RadiclePeer {
...
+              } catch (e) {
+                console.log("Error parsing event", data);
+              }
+            }
+          });
+      },
+    );
+  }
+
+  public async routesTo(rid: string, ...nodes: string[]) {
```
rudolfs, line 194: Could we call this function `addRoute`?

rudolfs (commented):
Nice work!

Noticed that if we don't have the binaries downloaded and run the `e2e` tests, we don't print instructions to the developer on how to download the binaries. Would be nice if we could add something like (this could also be a follow-up):

```
npm run test:e2e -- --project chromium

> test:e2e
> TZ='UTC' playwright test --project chromium

Error: rad version rad 0.8.0 (fc951b8) does not satisfy 29507b5
    at assertRadInstalled (file:///Users/rudolfs/work/radicle-interface/tests/support/globalSetup.ts:47:11)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
...

To download the required test binaries, run:
  👉 ./scripts/install-binaries

```

tests/support/peerManager.ts
```diff
@@ -1,10 +1,12 @@
 /* eslint-disable @typescript-eslint/naming-convention */
 import type { ExecaChildProcess, Options } from "execa";
+import type { NodeEvent, RoutingEntry } from "./cliTypes.js";
```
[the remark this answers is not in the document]
sebastinez: Yeah sure, didn't want to overload the `peerManager` file.

tests/support/peerManager.ts
sebastinez: Sure

tests/support/peerManager.ts
sebastinez: I have no issue with renaming the function, but the goal is to wait for the routing table to have a specific entry.
So maybe we could rename it to `waitForRoutes` ? So it's similar to `waitForEvent`wdyt?

rudolfs: Yeah, `waitForRoutes` makes more sense.

tests/support/peerManager.ts
sebastinez: e4c2861

tests/support/peerManager.ts
sebastinez: e4c2861

sebastinez:
> Nice work!
> 
> Noticed that if we don't have the binaries downloaded and run the `e2e` tests, we don't print instructions to the developer on how to download the binaries. Would be nice if we could add something like (this could also be a follow-up):
> 
> ```
> npm run test:e2e -- --project chromium
> 
> > test:e2e
> > TZ='UTC' playwright test --project chromium
> 
> Error: rad version rad 0.8.0 (fc951b8) does not satisfy 29507b5
>     at assertRadInstalled (file:///Users/rudolfs/work/radicle-interface/tests/support/globalSetup.ts:47:11)
>     at processTicksAndRejections (node:internal/process/task_queues:95:5)
> ...
> 
> To download the required test binaries, run:
>   👉 ./scripts/install-binaries
> ```

Good idea, addressed it with e4c2861

rudolfs (approved):

PR #789, opened 2023-05-24
Reset revision param when clicking on commit count button

sebastinez:
Allows to see the latest state when refreshing the page.

Closes #778


tests/support/fixtures.ts
```diff
@@ -129,18 +131,38 @@ export const test = base.extend<{
...
+      outputLog,
+    });
+    await use(peerManager);
+  },
+
   // eslint-disable-next-line no-empty-pattern
   stateDir: async ({}, use, testInfo) => {
-    const stateDir = testInfo.outputDir;
+    // Ok, we're moving the state dir to /tmp to avoid hitting the SUN_LEN limit, due to socket files.
+    // We still create a symlink in the testInfo.outputDir, so that we can easily find the state dir.
```
rudolfs, line 145: Ideally we would make the `control.sock` path configurable on `radicle-node`. And then only create a temporary OS tmp dir for that, instead all the spec state. This way once the test has run, the OS tmp dir with the socket would be cleaned up automatically. Right now all the state still remains in the system `/tmp` dir until the machine gets rebooted.

Here's some inspiration on how we did it in Upstream: https://github.com/radicle-dev/radicle-upstream/blob/main/test/support/peerManager.ts#L305.

tests/support/fixtures.ts
```diff
@@ -235,6 +257,7 @@ export async function createSeedFixture() {
   await bob.connect(palm);
 
   await alice.git(["clone", sourceBrowsingDir], { cwd: alice.checkoutPath });
+  alice.setCwd(Path.join(alice.checkoutPath, "source-browsing"));
```
rudolfs, line 260: I would rather prefer if we pass in the path to each command that we run explicitly, so that there is only one way do do it and it is explicit. I feel like having this extra state in the peer could make it easier to make mistakes.

tests/support/fixtures.ts
sebastinez: hmm I think one thing that we can propose is to have a `RAD_SOCKET_DIR` env variable or something similar that would only work with debug binaries. But will have to look into that first in heartwood

src/views/projects/Tree.svelte
```diff
@@ -9,7 +9,9 @@
 
   export let fetchTree: (path: string) => Promise<Tree | undefined>;
   export let path: string;
+  export let projectDefaultBranch: string;
```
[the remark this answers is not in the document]
sebastinez: yeah we can do that 👍

tests/support/fixtures.ts
sebastinez: ok, I found it verbose to add `{ cwd: Path.join(alice.checkoutPath, "source-browsing" }` to a lot of commands.
But maybe I can save it to a `projectCheckoutPath` variable and reference it like this

src/lib/utils.ts
```diff
@@ -24,17 +24,29 @@ export function getRawBasePath(
...
   if (branches) {
     const oid = branches[revision];
+    if (oid) return oid;
+  }
+  return undefined;
+}
 
-    if (oid) {
-      return oid;
+export function parseRevisionToOid(
```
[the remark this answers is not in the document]
sebastinez: ok sure 👍

tests/e2e/project/commits.spec.ts
```diff
@@ -115,3 +116,42 @@ test("relative timestamps", async ({ page }) => {
...
+  const alice = await peerManager.startPeer({
+    name: "alice",
+    gitOptions: gitOptions["alice"],
+  });
+  await alice.startHttpd(8090);
+  await alice.startNode();
+  const { rid, projectFolder } = await alice.createProject("alice-project");
+  await page.goto(`/seeds/127.0.0.1:8090/${rid}`);
+  await page.locator('role=link[name="1 commit"]').click();
+  await page.goto(`/seeds/127.0.0.1:8090/${rid}/history`);
```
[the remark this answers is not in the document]
sebastinez: I think this has been a leftover, but I think it makes sense to assert the url here. So will add the expect.

src/lib/utils.ts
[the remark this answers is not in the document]
sebastinez: 8abe9a6

tests/support/fixtures.ts
sebastinez: 8abe9a6

tests/e2e/project/commits.spec.ts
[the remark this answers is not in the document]
sebastinez: 8abe9a6

src/views/projects/Tree.svelte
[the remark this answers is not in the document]
sebastinez: 8abe9a6

tests/support/fixtures.ts
sebastinez: bbcf9ac

src/lib/router.ts
```diff
@@ -4,7 +4,7 @@ import { get, writable } from "svelte/store";
 
 import * as mutexExecutor from "@app/lib/mutexExecutor";
 import { loadRoute } from "@app/lib/router/definitions";
-import { unreachable } from "@app/lib/utils";
+import { isOid, unreachable } from "@app/lib/utils";
```
rudolfs, line 7: Could we also move `isOid` in here? 😬

src/views/projects/View.svelte
```diff
@@ -48,23 +49,22 @@
...
 
     if (branch) {
       const [rev, path] = [
         input.slice(0, branch[0].length),
         input.slice(branch[0].length + 1),
       ];
-
       parsed.revision = rev;
-      parsed.path = path ? path : "/";
+      parsed.path = path || "/";
```
rudolfs, line 64:
```diff
diff --git a/src/views/projects/View.svelte b/src/views/projects/View.svelte
index 59231789..c41456d3 100644
--- a/src/views/projects/View.svelte
+++ b/src/views/projects/View.svelte
@@ -61,10 +61,10 @@
         input.slice(branch[0].length + 1),
       ];
       parsed.revision = rev;
-      parsed.path = path || "/";
+      parsed.path = path ?? "/";
     } else if (utils.isOid(commitPath[0])) {
       parsed.revision = commitPath[0];
-      parsed.path = commitPath[1] || "/";
+      parsed.path = commitPath[1] ?? "/";
     } else {
       parsed.path = input;
     }

```

tests/support/peerManager.ts
```diff
@@ -108,7 +110,9 @@ export class RadiclePeer {
   public nodeId: string;
 
   #seed: string;
+  #socket: string;
   #radHome: string;
+  #cwd?: string;
```
rudolfs, line 115:
```diff
diff --git a/tests/support/peerManager.ts b/tests/support/peerManager.ts
index 5330909f..885e66c2 100644
--- a/tests/support/peerManager.ts
+++ b/tests/support/peerManager.ts
@@ -112,7 +112,6 @@ export class RadiclePeer {
   #seed: string;
   #socket: string;
   #radHome: string;
-  #cwd?: string;
   #eventRecords: NodeEvent[] = [];
   #outputLog: Stream.Writable;
   #gitOptions?: Record<string, string>;
@@ -350,7 +349,6 @@ export class RadiclePeer {
     opts?: Options,
   ): ExecaChildProcess {
     opts = {
-      cwd: this.#cwd,
       ...opts,
       env: {
         ...opts?.env,

```

src/lib/router.ts
sebastinez: yeah sure 👍

src/views/projects/View.svelte
sebastinez: Yeah this was also something I did first, but ended up using the `||` operator due to

`parsed.path` only ends up being `"/"` if `path` is either `null` or `undefined`.
If `path` is e.g. an empty string, (which can happen) we don't fallback to `"/"`.

The same happens with `commitPath[1]`.

We could be more explicit and set `path` to `undefined` or `null` explicitly if we found some edge cases like an empty string.

rudolfs: Let's leave it as is then.

tests/support/peerManager.ts
sebastinez: ok done

Issue #353, opened 2022-08-31
Create a packaged version for local deployment


sebastinez:
To allow end users mount a local instance of the web client on their machines, we need to package the web client so it can be served with the execution of a single binary.

**Process**
1. End user downloads binary
2. Executes binary
3. A static file server is mounted
4. The default browser is opened at `http://radicle.local`

**Pending points to be defined**
- [ ] The binary would need a compiled version of the web client
- [ ] How would we allow users to navigate to `http://radicle.home` and show the locally served site on different OS 
- [ ] When should we compile the application?

**Versions**
I wrote a small config file for [binserve](https://github.com/mufeedvh/binserve) which can be our static file server, it is 🦀 and quick
https://github.com/radicle-dev/radicle-interface/tree/sebastinez/packaging

rudolfs:
Would be interesting to investigate whether it's possible to bundle the server and assets into the same binary, à la https://github.com/peku33/web-static-pack.

rudolfs:
Regarding using `.local`, from what I've read it shouldn't be used for pointing to services running on your localhost.
It's original purpose is [multicast-dns](https://www.rfc-editor.org/rfc/rfc6762):

>  To remedy this problem, this document allows any computer user to
   elect to give their computers link-local Multicast DNS host names of
   the form: "single-dns-label.local.".  For example, a laptop computer
   may answer to the name "MyComputer.local.".  Any computer user is
   granted the authority to name their computer this way, provided that
   the chosen host name is not already in use on that link.  Having
   named their computer this way, the user has the authority to continue
   utilizing that name until such time as a name conflict occurs on the
   link that is not resolved in the user's favor.  If this happens, the
   computer (or its human user) MUST cease using the name, and SHOULD
   attempt to allocate a new unique name for use on that link.  These
   conflicts are expected to be relatively rare for people who choose
   reasonably imaginative names, but it is still important to have a
   mechanism in place to handle them when they happen.

So if someone with a mac on your local network names their coputer "radicle", you're in trouble.

<img width="780" alt="Screenshot 2022-08-31 at 13 45 16" src="https://user-images.githubusercontent.com/158411/187671239-f96bf4b4-ef48-49d7-9e26-100ef6124922.png">

rudolfs:
Another potential avenue to explore would be PWAs or "add-to-homescreen". Unfortunately [only chrome supports it](https://caniuse.com/?search=PWA) at the moment.

cloudhead:
[...]

PR #806, opened 2023-06-02
Display a seed URL that matches `isLocal` as `radicle.local`

sebastinez:
Bildschirmfoto 2023-06-02 um 10 19 23


rudolfs (approved):
Nice. This makes sense from a consistency standpoint.

To me the bigger question is whether we should call the local node `radicle.local`, because it implies that it is a domain name that can be resolved. And our initial explorations revealed that [using the `.local` TLD is not a good idea](https://github.com/radicle-dev/radicle-interface/issues/353#issuecomment-1232829806). Same with any other DNS name, the issue being that we would have to explicitly set it up somehow before it would work and it's hard to set it up correctly across different platforms.

PR #807, opened 2023-06-02
Add font-family monospace to commit description

### LINKED ISSUE #784: Inconsistent monospace fonts
I noticed that the font used in diffs is not the same as the one in commit messages. See:

image

One is `monospace`, and the other is `JetBrains Mono`. What's strange on my system is that those are supposed to be the same font, so I'm not sure why this is happening. But regardless, they should have the same CSS font family.

sebastinez:
Closes #784 

Bildschirmfoto 2023-06-02 um 10 56 03

### DIFF
--- a/src/views/projects/Commit.svelte
+++ b/src/views/projects/Commit.svelte
@@ -30,6 +30,7 @@
     align-items: center;
   }
   .description {
+    font-family: var(--font-family-monospace);
     margin: 1rem 0;
     white-space: pre-wrap;
   }


rudolfs:
Looks ok to me. But the bigger question is whether it shouldn't be [markdown altogether](https://github.com/radicle-dev/radicle-interface/issues/784#issuecomment-1573533077)?

rudolfs (approved):

PR #809, opened 2023-06-02
Relax radicle-httpd API schema checking

rudolfs:
This way we can deploy back-end API changes that are backwards compatible without breaking the UI.

The downside is that we might miss new fields that are added to `radicle-httpd`.

### DIFF
--- a/httpd-client/index.ts
+++ b/httpd-client/index.ts
@@ -25,7 +25,7 @@ import type {
 import type { RequestOptions, Method } from "./lib/fetcher.js";
 import type { ZodSchema } from "zod";
 
-import { array, literal, number, strictObject, string, union } from "zod";
+import { array, literal, number, object, string, union } from "zod";
 
 import * as project from "./lib/project.js";
 import * as session from "./lib/session.js";
@@ -57,7 +57,7 @@ export interface Node {
   id: string;
 }
 
-const nodeSchema = strictObject({
+const nodeSchema = object({
   id: string(),
 }) satisfies ZodSchema<Node>;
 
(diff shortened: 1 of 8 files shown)


sebastinez:
I think changing from `strictObject` to `object` doesn't do what you think it does.
`zod` strips by default unrecognized keys from objects.

For it to be backwards compatible we would need to use `.passthrough()` before calling `.parse()`
https://github.com/colinhacks/zod#passthrough

Apart from this, is there maybe some way to test the httpd-api test against strictObjects?

sebastinez:
Ok just an idea, maybe we can export the schemas inside a `testExports` module and call the `strict()` method on them, like this we can have a test that lets us know if we are missing some unknown new field

rudolfs:
> But.. maybe we can export the schemas inside a testExports module and call the strict() method on them, like this we can have a test that lets us know if we are missing some unknown new field?

Hm, good call. Will check if we can get that to work.

rudolfs:
> > But.. maybe we can export the schemas inside a testExports module and call the strict() method on them, like this we can have a test that lets us know if we are missing some unknown new field?
> 
> Hm, good call. Will check if we can get that to work.

On a second look, there doesn't seem to be a way to easily enable/disable object strictness recursively during runtime without declaring all the schemas twice, once with `object` and another time with `strictObject`. More info [here](https://github.com/colinhacks/zod/issues/1370). I'll punt on this for now.

sebastinez (approved):
Test pass, and I don't see any `strictObject` anymore. LGTM 👍

PR #811, opened 2023-06-04
Add required changes for svelte@4

sebastinez:
At some point soon Svelte@4 is going to be released, here are some changes that will be required or beneficial, for our codebase.

- `|local` is now the default behaviour for transitions so I removed it from the codebase.
- `SvelteComponentTyped` has been deprecated in favor of `SvelteComponent`
- New a11y rule for noninteractive elements with interactions, so we need to ignore a new a11y rule, (we should tackle those at some point...)
- Made `CobStateButton` more explicit to only take `IssueState` currently, since Svelte@4 is stricter on `createEventDispatcher`, we should be able to add more types when needed.

Pending:
- [x] Still an issue with `ComponentType` not exported by Svelte https://github.com/sveltejs/svelte/issues/8692


rudolfs:
Did you also bump transitive deps? Seems like a lot of additions in the lockfile.

sebastinez:
Yeah did also a dependencies update and bump transitive deps.

rudolfs (approved):
🚀

Issue #745, opened 2023-05-10
Some improvements to new patch page


cloudhead:
It's great to finally have this land! I have some suggestions on how we can improve it further:

1. The alignment on these looks funny; I think maybe if we use the same font size for both it might look better, I don't know. Currently it looks like "5 days ago" is higher than the +-
2. I would swap the order of the two, so that it's `5 days ago +23 -43`
<img width="352" alt="image" src="https://github.com/radicle-dev/radicle-interface/assets/40774/da557b74-846e-437c-a41c-79f1a1141d89">

3. We should probably dim the string "Revision", to let the revision id stand out more. Alternatively, we could experiment with an icon instead of "Revision", or even in combination.
4. The "opened" and "updated" messages should be part of the same event as the commits, otherwise the commits look like they're their own event. So they should be grouped with one background around them.
5. The commits are not clearly commits, so I'd suggest adding an icon like on github, to the left of each of them. I'd also say we should use monospace font on them, as they are their own type of content, and have the hover effect be `underline`.

<img width="1122" alt="image" src="https://github.com/radicle-dev/radicle-interface/assets/40774/4e0cfe05-17d1-4a2e-bffa-2dbfb71b1d87">

Nice work :+1:

rudolfs:
> 2. I would swap the order of the two, so that it's 5 days ago +23 -43

We deliberately put the date last, because it's always available, contrary to the diff badge, which can also not be present. This way the diff badge/date columns always align neatly.

For example, here the diff badge is missing:
  <img width="1840" alt="Screenshot 2023-05-17 at 09 43 06" src="https://github.com/radicle-dev/radicle-interface/assets/158411/3526c73b-4a2a-46e6-a9ea-04ae8a662f53">

If it's the other way around it looks odd:
  <img width="1840" alt="Screenshot 2023-05-17 at 09 45 07" src="https://github.com/radicle-dev/radicle-interface/assets/158411/cefd57c3-1fa1-454f-b13a-373d8c6f881d">

cloudhead:
Oh ok, I mean that should not happen anymore, it was fixed. If we really care though we can add an `(error)` badge, as a bunch of other things won't work correctly if the diff stats aren't available.

cloudhead:
Alternatively we move the badge to in the expanded area:
<img width="583" alt="image" src="https://github.com/radicle-dev/radicle-interface/assets/40774/a57c75ef-d86f-480c-bba0-e06ea56ef58c">

[...]

PR #812, opened 2023-06-05
Tweak patch view visuals

rudolfs:
Closes: https://github.com/radicle-dev/radicle-interface/issues/745.

- commit hover state is underlined
- indent revision commits
- dim "Revision" text in header
- fix vertical alignment of header items in right-hand-side
- don't show first revision description, it's already shown as part of the patch description

Screenshot 2023-06-05 at 12 09 27

Screenshot 2023-06-05 at 12 07 40


cloudhead:
Very cool! The only comment is I wouldn't dim "Revision" that much, maybe just one notch down from the fg color.

sebastinez:
Nice @rudolfs!
I'm seeing that comments don't have a indentation and the `.revision-body` has a `margin-top: 1rem` and a `gap: 1rem` and this ends up being a lot of space.

Bildschirmfoto 2023-06-05 um 14 06 19

sebastinez:
Merges seem to render nicely but also without the `margin-left: 1rem`
is that something we want on all revision elements or only on the review ones?

Bildschirmfoto 2023-06-05 um 14 10 15

rudolfs:
> minor suggestion: Can we also brighten the box-drawing character on hover; might look cool.

It's not part of the actual link, so it wouldn't make sense.

sebastinez (approved):
LGTM, issues fixed

PR #814, opened 2023-06-05
Add a cobs fixture and issue and patch tests

sebastinez:
- Get visual tests running for issues and patches
- Navigates through patches and issues checks that buttons and links do what promised.
- Adds some utility functions to create issues and patches easily. 
- Adds a cobsFixture to the globalSetup, created by palm

Pending:
- [x] We need [heartwood@bd37b94](https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/bd37b94d06e7aaf2c3ef2e8303b8e415066d12e9) to be merged and deployed on https://files.radicle.xyz before merge


rudolfs (commented):
Nice work! This will help us a lot with regressions.

One thought: I think we would benefit from splitting up the create issue/patch function actions into separate functions. For one the function wouldn't grow more in complexity when we add more actions and would be easier to reason about. Another benefit is that we could run the actions even after an issue is created if we need to later on. And we would still retain the abstraction over the bare rad commands, so if the underlying thing changes we just have to change it in one place instead of all the tests.

I'm thinking something like this:

```
const issue = createIssue({title: "issue title", description: "issue description, ...});
tagIssue("tag one", issue);
commentOnIssue("this is a comment", issue);
...
```

rudolfs:
This could also be a follow up, since it's stuff that the visual tests uncovered, rather than the tests themselves:

- should we also indent these messages? and should the indentation width match issue reply comments (or vice versa)?
  patch-page-4-actual
issue-page-1-actual

- what's the triple backtick here?
issue-page-3-actual

tests/support/cobs/shared.ts
```diff
@@ -0,0 +1,48 @@
...
+  await peer.rad(["tag", objectId, tag], {
+    cwd: options.projectFolder,
+    env: { RAD_COMMIT_TIME: options.timestamp },
+  });
+}
+
+// Increases the timestamp by daily increments
+export function increaseTime(timestamp: string, increase: number): string {
+  return (parseInt(timestamp) + increase * 86400).toString();
+}
```
rudolfs, line 48: This way we don't need the comment:
```diff
export function increaseTime(timestamp: string, days: number): string {
  return (parseInt(timestamp) + days * 86400).toString();
}
```

tests/support/cobs/issue.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+import type { RadiclePeer } from "@tests/support/peerManager.js";
+
+export async function createIssue(
```
rudolfs, line 4: We could also call the methods `create`, `changeState` and then import the module with:
```
import * as issue from "./issue.ts"

issue.create
issue.changeState
```

tests/support/cobs/issue.ts
```diff
@@ -0,0 +1,39 @@
...
+  ];
+  const { stdout } = await peer.rad(issueOptions, { cwd: projectFolder });
+  const match = stdout.match(/Issue {3}([a-zA-Z0-9]*)/);
+  if (!match) {
+    throw new Error("Not able to parse issue id");
+  }
+  return match[1];
+}
+
+export async function stateIssue(
```
rudolfs, line 29: I think `changeIssueState` would be clearer.

src/views/projects/Cob/Revision.svelte
```diff
@@ -208,6 +208,9 @@
       {/if}
       {#if previousRevOid}
         <ProjectLink
+          title="Compare {utils.formatObjectId(
```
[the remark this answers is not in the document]
sebastinez: 🙇

src/views/projects/Patch.svelte
```diff
@@ -347,6 +347,7 @@
                       resource: "patch",
                       params: { patch: patch.id, revision: item.id },
                     },
+                    search: `tab=${currentTab}`,
```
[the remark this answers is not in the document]
sebastinez: 🙇

tests/support/support.ts
```diff
@@ -4,3 +4,7 @@ import * as Crypto from "node:crypto";
 export function randomTag(): string {
   return Crypto.randomBytes(8).toString("hex");
 }
+
+export const unreachable = (value: never): never => {
```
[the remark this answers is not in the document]
sebastinez: Removed it, since not needed anymore by extracting functions

tests/support/cobs/shared.ts
sebastinez: Inlined this function in the end in the `createOptions` fn to create the `execa` options object

tests/support/cobs/issue.ts
sebastinez: Done

sebastinez:
> One thought: I think we would benefit from splitting up the create issue/patch function actions into separate functions. 

Done, I don't feel it's that nice, but as we talked about, this will eventually allow us to do a abstraction that makes sense for us down the line.

tests/support/cobs/patch.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
rudolfs, line 1: Don't need this one either.

src/views/projects/Issue.svelte
```diff
@@ -18,7 +18,7 @@
   import Markdown from "@app/components/Markdown.svelte";
   import TagInput from "./Cob/TagInput.svelte";
   import Textarea from "@app/components/Textarea.svelte";
-  import Thread from "@app/components/Thread.svelte";
+  import ThreadComponent from "@app/components/Thread.svelte";
```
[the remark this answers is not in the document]
sebastinez: It's just for consistency, we use the same naming in `/src/views/projects/Cob/Revision.svelte`.
So I thought either we change both to `Thread` or both to `ThreadComponent`

tests/support/cobs/patch.ts
sebastinez: bc58cd8

tests/support/cobs/issue.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
sebastinez: bc58cd8

src/views/projects/Issue.svelte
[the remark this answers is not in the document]
rudolfs: Ah, I think we used it like this in one of the places because we were importing a type called `Thread` and there was a clash. We should just use `Thread` in both places, since there isn't any clash.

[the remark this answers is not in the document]
sebastinez: aae5ec7

tests/support/cobs/patch.ts
```diff
@@ -17,15 +17,14 @@ export async function create(
...
-    ],
-    options,
-  );
+  const cmd = [
+    "push",
+    ...messages.map(msg => ["-o", `patch.message=${msg}`]).flat(),
+    "rad",
+    "HEAD:refs/patches",
+  ];
+  console.log(cmd);
```
rudolfs, line 26: 👀

sebastinez: ouh thanks 😁

sebastinez: done

PR #825, opened 2023-06-12
Update dependencies

rudolfs:
There's a warning about TypeScript and eslint version mismatch when running `scripts/check`. Should be harmless and will probably go away with the next eslint upgrade.

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.1.0

YOUR TYPESCRIPT VERSION: 5.1.3

Please only submit bug reports when using the officially supported version.

=============
```


sebastinez (approved):
Yeah seems to work fine, LGTM 👍

sebastinez (approved):
Nice, seems like the `fake-timers` package has reverted the mocking Node `timers`

PR #838, opened 2023-06-14
Add more tests to revision selector

### LINKED ISSUE #837: Test revision selector in Patch commits and files tabs
We bumped into a [regression](https://github.com/radicle-dev/radicle-interface/pull/835) while refactoring project data loading. It would be nice to have this functionality covered by e2e tests.

Screenshot 2023-06-13 at 17 10 00

sebastinez:
This PR adds more expects and cases to the revision selector.

- Making sure the visibility of the RevisionSelector is correct
- Validating the displayed content changes to the correct one before and after a revision change.

Closes #837

### DIFF
--- a/tests/e2e/project/patches.spec.ts
+++ b/tests/e2e/project/patches.spec.ts
@@ -36,12 +36,36 @@ test("navigate patch details", async ({ page }) => {
 
 test("use revision selector", async ({ page }) => {
   await page.goto(`${cobUrl}/patches/0f3697fed2743549e3bf531e9fa81284a6de1466`);
-  await page.locator("role=link[name='Files']").click();
-  await page.locator("text='Revision febcbbd'").click();
-  await page.locator("role=link[name='Revision febcbbd']").click();
+  await page.getByRole("link", { name: "Files" }).click();
+
+  // Validating the latest revision state
+  await expect(
+    page.getByRole("cell", { name: "Had to push a new revision" }),
+  ).toBeVisible();
+  await page.getByRole("link", { name: "Commits" }).click();
+  await expect(page.locator(".commit-list .teaser")).toHaveCount(2);
+  await expect(
+    page.locator(".commit-list .teaser .markdown").first(),
+  ).toHaveText("Add more text");
+
+  // Switching to the initial revision
+  await page.getByText("Revision 5140fb2").click();
+  await expect(page.locator(".dropdown")).toBeVisible();
+  await page.getByRole("link", { name: "Revision 0f3697f" }).click();
+  await expect(page.locator(".dropdown")).toBeHidden();
+
+  // Validating the initial revision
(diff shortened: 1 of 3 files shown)


tests/e2e/project/patches.spec.ts
```diff
@@ -37,11 +37,35 @@ test("navigate patch details", async ({ page }) => {
 test("use revision selector", async ({ page }) => {
   await page.goto(`${cobUrl}/patches/0f3697fed2743549e3bf531e9fa81284a6de1466`);
   await page.locator("role=link[name='Files']").click();
-  await page.locator("text='Revision febcbbd'").click();
-  await page.locator("role=link[name='Revision febcbbd']").click();
+
+  // Validating the latest revision state
+  await expect(
+    page.locator("role=cell[name='Had to push a new revision']"),
```
geigerzaehler, line 43: You can also write this as follows:
```suggestion
    page.getByRole("cell", { name: "Had to push a new revision" }),
```

rudolfs (approved):
Thanks! LGTM.

tests/e2e/project/patches.spec.ts
sebastinez: Thanks @geigerzaehler for the feedback!
I followed the style of the codebase and kept using the `locator("role=...)`

Do you know if there is any benefit of using one over the other?
I also like `getByRole` better since that is what the `Pick locator` in debug mode outputs.

sebastinez: Ok nvm, from what I found in the playwright library it's a wrapper over the bare `locator` function

geigerzaehler: > Do you know if there is any benefit of using one over the other?

For me there are two main benefits: It’s easier to read since the information is not encoded with a custom syntax in a string but just regular JavaScript. And then it is also easier to write correctly because I get completion support and the code is type checked.

On top of that the `getByXXX()` functions also offer more [advanced features](https://playwright.dev/docs/api/class-framelocator#frame-locator-get-by-role-option-checked) for more expressive locators.

tests/e2e/project/patches.spec.ts
```diff
@@ -37,11 +37,35 @@ test("navigate patch details", async ({ page }) => {
...
+  ).toBeVisible();
+  await page.locator("role=link[name='Commits']").click();
+  await expect(page.locator(".commit-list .teaser")).toHaveCount(2);
+  await expect(
+    page.locator(".commit-list .teaser .markdown").first(),
+  ).toHaveText("Add more text");
+
+  // Switching to the initial revision
+  await page.locator("text='Revision 5140fb2'").click();
```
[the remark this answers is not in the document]
sebastinez: Done

tests/e2e/project/patches.spec.ts
sebastinez: ✅

Issue #711, opened 2023-04-26
Improve ErrorMessage component


rudolfs:
It would be nice to at least show a snippet of the actual error along with the human readable message.

<img width="500" alt="Screenshot 2023-04-26 at 11 28 20" src="https://user-images.githubusercontent.com/158411/234533507-c2cc3b0d-252c-411c-b686-b35324875de2.png">

Something along these lines:

<img width="628" alt="Screenshot 2023-06-19 at 09 16 30" src="https://github.com/radicle-dev/radicle-interface/assets/158411/abf7da98-61d4-4acb-9cae-81ed7746f3ac">

[Check the `ErrorModal` before we threw out Ethereum](https://github.com/radicle-dev/radicle-interface/blob/3d018dae1830f4ac0bd8473b4c58455833c1064c/src/components/ErrorModal.svelte).

rudolfs:
@geigerzaehler also [had some good points](https://github.com/radicle-dev/radicle-interface/pull/727#discussion_r1184890515) on improving error handling within the app:

> Here are some thoughts around error handling. Nothing that we need to change now but we should keep an eye on this and explore different solutions.
>
> - We need to show error.message as well
> - We should probably unfold [error cause chains](https://github.com/tc39/proposal-error-cause#chaining-errors) in the error view to provide more detailed information
> - We should make use of error cause chains to provide more detailed information

rudolfs:
Superseded by https://github.com/radicle-dev/radicle-interface/issues/986.

PR #841, opened 2023-06-15
Add error handling for issue creation and commenting

### LINKED ISSUE #793: Add error handling to issue editing and commenting
We make API requests and only handle the happy path. We should inform the user if a API request fails.

sebastinez:
Catches errors that can occur during issue editing, commenting and replying.
Also adds a `authenticatedPeer` fixture, which authenticates a `RadiclePeer` with `radicle-httpd`

Shows a display that tells the user something went wrong, and a `ErrorMessage` component to copy the message and stackTrace
Bildschirmfoto 2023-06-15 um 13 29 36

Also when there is an issue while refreshing the issue with a GET request after the PATCH update action, we show a different error

Bildschirmfoto 2023-06-15 um 13 32 51

Closes #793

### DIFF
--- a/src/views/projects/Issue.svelte
+++ b/src/views/projects/Issue.svelte
@@ -1,19 +1,23 @@
 <script lang="ts">
   import type { BaseUrl, Issue, IssueState } from "@httpd-client";
+  import type { IssueUpdateAction } from "@httpd-client/lib/project/issue";
+  import type { Session } from "@app/lib/httpd";
 
   import { isEqual } from "lodash";
 
   import * as router from "@app/lib/router";
+  import * as modal from "@app/lib/modal";
   import * as utils from "@app/lib/utils";
   import { HttpdClient } from "@httpd-client";
   import { httpdStore } from "@app/lib/httpd";
 
-  import AssigneeInput from "./Cob/AssigneeInput.svelte";
+  import AssigneeInput from "@app/views/projects/Cob/AssigneeInput.svelte";
   import Authorship from "@app/components/Authorship.svelte";
   import Badge from "@app/components/Badge.svelte";
   import Button from "@app/components/Button.svelte";
   import CobHeader from "@app/views/projects/Cob/CobHeader.svelte";
   import CobStateButton from "@app/views/projects/Cob/CobStateButton.svelte";
+  import ErrorModal from "@app/views/projects/Cob/ErrorModal.svelte";
   import Icon from "@app/components/Icon.svelte";
   import Markdown from "@app/components/Markdown.svelte";
   import TagInput from "./Cob/TagInput.svelte";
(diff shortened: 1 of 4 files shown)


sebastinez:
Also this implementation will benefit from closing #711, which I'm looking into.

tests/e2e/project/issues.spec.ts
```diff
@@ -17,3 +18,41 @@ test("navigate single issue", async ({ page }) => {
...
+  );
+
+  await page.goto(
+    `/seeds/127.0.0.1:8070/${rid}/issues/d316f7a90a40dacbfb8728044bad50c9f71d44ba`,
+  );
+
+  await page.getByPlaceholder("Leave your comment").fill("This is a comment");
+  await page.getByRole("button", { name: "Comment" }).click();
+  await expect(page.getByText("Issue editing failed")).toBeVisible();
+  await expect(page.locator(".modal")).toBeVisible();
```
rudolfs, line 57: Checking for the modal is probably redundant, because we already check the text content in the previous expect.

src/lib/issues.ts
```diff
@@ -0,0 +1,60 @@
+import type { HttpdClient, Issue } from "@httpd-client";
```
rudolfs, line 1: I think we can move this into `Issue.svelte` as it is only used there.

src/views/projects/Cob/CrudErrorModal.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
```
rudolfs, line 1: Can we call this just `ErrorModal` or `CobErrorModal`? As CRUD stands for Create, Read, Update and Delete. We don't really have errors for read and delete.

src/views/projects/Issue.svelte
```diff
@@ -115,39 +119,31 @@
...
-      await api.project.updateIssue(
+      await issues.updateIssue(
         projectId,
         issue.id,
         { type: "lifecycle", state },
-        $httpdStore.session.id,
+        $httpdStore.session,
+        api,
       );
-      void router.push({
```
rudolfs, line 140: Removing the push breaks issue counter updates in the header when changing the issue state. We can maybe leave the push as is, since if the request fails the user will still land on an error page.

tests/e2e/project/issues.spec.ts
sebastinez: 3af3c74

src/lib/issues.ts
sebastinez: 3af3c74

src/views/projects/Cob/CrudErrorModal.svelte
sebastinez: 3af3c74

rudolfs:
- The title edit button shows an error the first time around when the backend is gone, but not subsequent times.


https://github.com/radicle-dev/radicle-interface/assets/158411/89a681fc-4dc7-4ebf-b20b-e0d84f77f34b

- There's a modal on top of an error when closing an issue and the backend is gone. Maybe this isn't such a big deal, not sure. Or maybe in the error case we just don't try to reload the issue?


https://github.com/radicle-dev/radicle-interface/assets/158411/8fd51f9d-64d2-47c7-ad58-77c15b912380

rudolfs:
> hmm good catches, will look into them.

I'm still seeing the same behaviour.

sebastinez:
> I'm still seeing the same behaviour.

yeah just did a rebase for now, but I'm working on

src/views/projects/Issue.svelte
sebastinez: @rudolfs added issue #897 to implement the header counter update test

rudolfs (approved):

Issue #831, opened 2023-06-13
Description text not rendering <RID> and <OID>


xphoniex:
Actual text:

![Screenshot from 2023-06-13 15-19-42](https://github.com/radicle-dev/radicle-interface/assets/14107758/22149c8c-a5d7-4adf-90a2-e503bd0b5eae)

Getting:

![Screenshot from 2023-06-13 15-18-59](https://github.com/radicle-dev/radicle-interface/assets/14107758/8e509244-1ad9-4855-a68a-e40895bb4ccb)

[Link](https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/536a72783b7b0d2d3fe6e627ce0d3ba73923ba72?tab=activity)

rudolfs:
This is resolved now.

<img width="724" alt="Screenshot 2023-06-26 at 13 20 22" src="https://github.com/radicle-dev/radicle-interface/assets/158411/1a9207da-754b-4c3c-88d7-744442e391f8">

sebastinez:
So just wanted to follow up with this issue, since we reverted back with #888 and the text is not rendering again.
The way to do this so it works as a patch description is to use markdown code ticks in the commit message

---
This improves response time of:

- GET `/api/v1/projects/<RID>/tree/<OID>/from` 205.71ms down to 376.4µs
- GET `/api/v1/projects/<RID>/tree/<OID>/.nix` from 111.06ms down to 351.70µs
Also, CPU spike goes from 13.8% and 7.8% respectively down to 0.0%.

Signed-off-by: xphoniex [dj.2dixx@gmail.com](mailto:dj.2dixx@gmail.com)

PR #846, opened 2023-06-16
Custom HTML tags in markdown not rendering

sebastinez:
Ok this PR should address and close issue #831, by escaping all tokens besides some trusted ones.


src/lib/markdown.ts
```diff
@@ -118,6 +120,21 @@ const anchorMarkedExtension = {
   },
 };
 
+export const walkTokens = (token: marked.Tokens.Generic) => {
+  if (token.type !== "code" && token.type !== "codespan") {
+    if (
+      "text" in token &&
+      !token.text.match(new RegExp(`${trustedHtmlTags.join("|")}`, "g"))
```
cloudhead, line 127: What exactly is `text`? Is it `<foo>` and `</foo>`, for example? If so, why not just replace the contents in one go, instead of matching first? Also, I think closing tags are not handled by this code?

src/lib/markdown.ts
```diff
@@ -118,6 +120,21 @@ const anchorMarkedExtension = {
...
+    ) {
+      token.text = token.text.replace(
+        /<(.*)>/g,
+        (_match: RegExpMatchArray, tagContent: string) =>
+          `&lt;${tagContent}&gt;`,
```
cloudhead, line 132: Would escaping work? ie. `\<` and `\>`?

tests/support/fixtures.ts
```diff
@@ -365,7 +365,12 @@ export async function createCobsFixture(peer: RadiclePeer) {
     "Add README\n\nThis commit adds more information to the README",
     "feature/add-readme",
     () => Fs.writeFile(Path.join(projectFolder, "README.md"), "# Cobs Repo"),
-    ["Let's add a README", "This repo needed a README"],
+    [
+      "Let's add a README",
+      "This repo needed a README",
+      "Let's also check that <RID> and <OID> get rendered correctly",
+      "Outside of a code block and `inside <RID> and <OID> of a codeblock`",
```
cloudhead, line 372: We should test closing tags, and trusted tags as well, though maybe not in the fixture. We should have a dedicated test for the markdown rendering of tags.

src/lib/markdown.ts
sebastinez: > What exactly is text? 
`token.text` is what the markdown tokenizer categorizes as `text` and it depends on each token, here you can see some definitions and which one have and which one don't have the text property.
https://github.com/markedjs/marked/blob/450da06c7c8466958002e5c1744c91d02b300e7a/src/Tokenizer.js#L67
In our case e.g. we don't want to match against `<hr/`> or a whitespace, that's why I check for `token.text`

>  If so, why not just replace the contents in one go, instead of matching first.
I matched against a regex to skip the tokens we don't want to escape, but we could probably also avoid regex and check like `!trustedHtmlTags.includes(token.text)`.

>  closing tags are not handled by this code
Closing tags are handled since we match against the inner text and don't include a closing slash e.g. `<code> === </code>` for us 😁

src/lib/markdown.ts
sebastinez: I tried this first, but it ended up rendering the backslashes instead of escaping the `<` and hiding the backslash

tests/support/fixtures.ts
sebastinez: yeah we can do that

src/lib/markdown.ts
sebastinez: Ok just tried the 

> I matched against a regex to skip the tokens we don't want to escape, but we could probably also avoid regex and check like !trustedHtmlTags.includes(token.text).

And I think it's easier to match the token out of the html tag and handle the closing tags instead of trying to match with string or array methods..

tests/support/fixtures.ts
sebastinez: Added a new test in the markdown repository to the visual tests that verifies the closing tags and trusted tags.

Bildschirmfoto 2023-06-19 um 11 47 48

src/lib/markdown.ts
sebastinez: Okay found this solution that I think is nice and we can avoid another regex expression and matching
https://github.com/radicle-dev/radicle-interface/pull/846/commits/50c24f592963ef98508d0245b6682acc70b9e4a4

tests/visual/markdown.spec.ts
```diff
@@ -173,3 +173,16 @@ test("relative image not able to being loaded", async ({ page }) => {
   });
   await expect(page).toHaveScreenshot({ fullPage: true });
 });
+
+test.describe(async () => {
+  test.use({ viewport: { width: 1280, height: 625 } });
```
rudolfs, line 178: Can we just make this full page, since we're not splitting up anything?

rudolfs (approved):

PR #847, opened 2023-06-16
Fix padding issues in source browser on mobile

### LINKED ISSUE #843: Regression on mobile between sourceHeader and CobHeader
It seems like the project browsing refactor has added some margins and paddings that broke it on mobile

IMG_D7787F272154-1

sebastinez:
Current:
Bildschirmfoto 2023-06-16 um 13 43 42

New:
Bildschirmfoto 2023-06-16 um 13 43 42
Closes #843

### DIFF
--- a/public/index.css
+++ b/public/index.css
@@ -10,7 +10,7 @@
   --border-radius-round: 10rem;
 
   --content-max-width: 1920px;
-  --content-min-width: 480px;
+  --content-min-width: 400px;
 
   --scrollbar-width: 0.5rem;
 
--- a/src/views/projects/Header.svelte
+++ b/src/views/projects/Header.svelte
@@ -38,9 +38,6 @@
     .header {
       padding-left: 2rem;
     }
-    .header {
-      margin-bottom: 1.5rem;
-    }
   }
 </style>
 
--- a/src/views/projects/View.svelte
+++ b/src/views/projects/View.svelte
@@ -45,12 +45,6 @@
     min-width: var(--content-min-width);
     padding-bottom: 4rem;
   }
-  @media (max-width: 960px) {
-    main {
-      padding-top: 2rem;
-      min-width: 0;
-    }
-  }
 </style>
 
 <div class="header">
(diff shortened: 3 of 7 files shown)


tests/visual/project.spec.ts
```diff
@@ -5,9 +5,16 @@ import {
...
 
-test("source tree page", async ({ page }) => {
-  await page.goto(sourceBrowsingUrl, { waitUntil: "networkidle" });
-  await expect(page).toHaveScreenshot();
+test.describe(async () => {
+  test("source tree page", async ({ page }) => {
+    await page.goto(sourceBrowsingUrl, { waitUntil: "networkidle" });
+    await expect(page).toHaveScreenshot();
+  });
+  test.use({ viewport: { width: 414, height: 896 } });
```
rudolfs, line 13: I think it's best to have a separate folder in `tests/visual/mobile` and copy the project spec there and run all the tests with a custom viewport.

We can do this for all the other tests as well, but that can be a follow up (see https://github.com/radicle-dev/radicle-interface/issues/850).

It seems like `test.use` applies to the whole test otherwise:

Screenshot 2023-06-19 at 14 46 22

rudolfs: Which phone is the custom size `{ width: 414, height: 896 }`? Maybe we can extract that into a variable and use it for the other mobile tests.

sebastinez: thats an iPhoneXR according to chrome, can extract the size into a variable

sebastinez: Ok moved the mobile test into a separate folder.
Would prefer to move the other tests in a separate PR.

rudolfs: I think the other alternative is to wrap all tests in a file in a top level `describe` scope and define the screen size there.  No need to change anything.

```
describe "mobile" {
  use mobile-screen size

  test-1
  test-2
  ...
}
describe "desktop" {
  use mobile-screen size

  test-1
  test-2
  ...
}
```

rudolfs (approved):

tests/visual/project.spec.ts
sebastinez: hmm ok yeah makes sense 👍

PR #849, opened 2023-06-16
Separate commit listing from revision timeline

sebastinez:
Moving the commit listing out of the revision timeline, improves the code somewhat, since the revision object isn't really a revision event like merge, review, etc.

image

### DIFF
--- a/src/components/Comment.svelte
+++ b/src/components/Comment.svelte
@@ -26,7 +26,7 @@
   .card {
     display: flex;
     flex-direction: column;
-    border-radius: var(--border-radius);
+    border-radius: inherit;
     background-color: inherit;
   }
   .card-header {
@@ -38,7 +38,7 @@
   }
   .card-body {
     font-size: var(--font-size-small);
-    padding: 0 1rem 0.5rem 1rem;
+    padding: 0 1rem 0.7rem 1rem;
   }
   .actions {
     display: flex;
--- a/src/components/Link.svelte
+++ b/src/components/Link.svelte
@@ -5,6 +5,7 @@
   import { push, routeToPath, useDefaultNavigation } from "@app/lib/router";
 
   export let route: Route;
+  export let title: string | undefined = undefined;
 
   const dispatch = createEventDispatcher<{
     afterNavigate: never;
@@ -21,6 +22,6 @@
   }
 </script>
 
-<a on:click={navigateToRoute} href={routeToPath(route)}>
+<a on:click={navigateToRoute} href={routeToPath(route)} {title}>
   <slot />
 </a>
--- a/src/components/Thread.svelte
+++ b/src/components/Thread.svelte
@@ -57,10 +57,10 @@
   }
   .comment {
     background-color: var(--color-foreground-1);
-    border-radius: var(--border-radius);
+    border-radius: var(--border-radius-small);
   }
   .reply {
-    margin-left: 3rem;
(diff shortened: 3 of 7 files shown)


rudolfs (commented):
Could we add some visual specs for the various elements?
  - comments
  - statuses (accepted/merged/archived)

Is this just outdated schema, or is there something wrong with this comment? It seems like it doesn't have a body text.

http://localhost:3000/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/cb6e04ae92779022fb740ea44643ee6c55d01496

Screenshot 2023-06-19 at 13 53 02

rudolfs:
> You mean to have more test specs that check the content? Or just visually?

Visual specs are okay. I want to see multiple comments and the PR status elements in visual specs.

cloudhead:
Would be good to see it with a revision description too.

sebastinez:
Accepted and merged patch, with multiple comments and replies.
P.S. I removed the blackground that made out of each comment and is replies a big grey area, seemed off.
patch-page-3-visual-darwin

A big revision description and a rejected patch
patch-page-4-visual-darwin

For draft and archival, we don't have yet status elements, since they aren't part of the patch timeline.

cloudhead:
Hmm I'm almost thinking we should pull the comments/activity out of the box now.. The nested boxes are a bit much after this change.

image


wdyt?

sebastinez:
> Hmm I'm almost thinking we should pull the comments/activity out of the box now.. The nested boxes are a bit much after this change.
> 
> image
> wdyt?

hmm interesting, doesn't sound bad, I think we could improve it by indenting the timeline? So the timeline events aren't on the same level as the revision component? wdyt? @cloudhead

cloudhead:
I don't think that's even necessary, since the revision box is styled differently already and is clearly a "header" of sorts. Problem with indentation is always that we lose content-width, so I try to avoid it unless it's really needed (like comment replies).

cloudhead:
We *could* try a half indent though, ie. half the width of the comment reply indentation.. I'd have to see it to decide :)

cloudhead:
Looking at it more, the only issue is we'd probably want to align the `open patch` activity with the comments. So we'd have to unindent that as well in the revision box. Probably worth playing around with some options.

sebastinez:
Okay added some concepts on putting everything inside their own box, and a half indent to show the relationship of the timeline events with the revision head.
Didn't want to remove the merge, review colors, since I think they really help visually.

wdyt @cloudhead ?

Bildschirmfoto 2023-06-28 um 14 20 45
Bildschirmfoto 2023-06-28 um 14 19 46

cloudhead:
Ok, a few comments:

* I think it looks good
* I'd go back with the borderless with bg-color style for comments, to differentiate it from the revision header
* The indent works, though I'd keep it small, also for comment replies
* It might make sense for the indent to be the same everwhere
* The padding under the comment body looks too small

sebastinez:
Thanks @cloudhead, did apply some changes and reverting others.
wdyt about the latest one?

> The padding under the comment body looks too small

Did normalize the padding on the Comment component to 1rem.

The indentation is now 1.5rem for the first level and another 1.5rem for replies.

Bildschirmfoto 2023-07-03 um 08 46 02

rudolfs (approved):

PR #853, opened 2023-06-20
Simplify project view

rudolfs:
Move everything that is sub-view related into the respective views.

### DIFF
--- a/src/views/projects/Issue/New.svelte
+++ b/src/views/projects/Issue/New.svelte
@@ -1,10 +1,9 @@
-<script lang="ts" strictEvents>
+<script lang="ts">
   import type { BaseUrl } from "@httpd-client";
-  import type { Session } from "@app/lib/httpd";
 
-  import { createEventDispatcher } from "svelte";
-
+  import * as httpd from "@app/lib/httpd";
   import * as modal from "@app/lib/modal";
+  import * as router from "@app/lib/router";
   import * as utils from "@app/lib/utils";
   import { HttpdClient } from "@httpd-client";
   import { httpdStore } from "@app/lib/httpd";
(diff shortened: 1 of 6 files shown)


src/views/projects/Patch.svelte
```diff
@@ -136,7 +142,8 @@
   }));
 
   $: currentRevision =
-    patch.revisions.find(r => r.id === revision) || patch.revisions[0];
+    patch.revisions.find(r => r.id === (revision ?? latestRevision)) ||
+    patch.revisions[0];
```
geigerzaehler, line 146: How about this:
```suggestion
    (revision && patch.revisions.find(r => r.id === revision) || patch.revisions[patch.revisions.length - 1]
```
Then you can get rid of `latestRevision` and the intention becomes much clearer.

src/views/projects/Issue/New.svelte
```diff
@@ -51,9 +49,20 @@
           assignees: utils.stripDidPrefix(assignees),
           tags: tags,
         },
-        session.id,
+        sessionId,
       );
-      dispatch("create", result.id);
+
+      void router.push({
```
geigerzaehler, line 55: I’d move this call after the try/catch block. If the call fails, it’s not an authentication error and the issue actually has been created so the error message would confuse the user.

geigerzaehler (commented):
Looks great! I have two suggestions

src/views/projects/Patch.svelte
rudolfs: 069d0a16ce2fc8355c363dbbb80477a365240ffb

geigerzaehler (approved):

PR #854, opened 2023-06-20
Handle app URL translation into seed BaseUrl one place

rudolfs:
To make the code easier to reason about we handle all app URL encoding and decoding and the related seed BaseUrl extraction in a single place.


geigerzaehler (approved):

geigerzaehler (commented):
I think you need to fix the tests in a couple of places

PR #855, opened 2023-06-20
Syntax highlighting moving to tree-sitter exploration

sebastinez:
This PR explores the migration from the @wooorm/starry-light syntax highlighter to use tree-sitter.

**Known issues**
- [x] If a file is deleted in an issue we can't fetch it on the new commit, but need to have the parent commit ready, needs change on `radicle-httpd` level


sebastinez:
With help of returning the blobs from `radicle-httpd` with the diff, I'm looking into a way to keep the current highlighter and add only the diff highlighting which was the initial missing feature

sebastinez:
We'll be moving part of the highlighting back to `radicle-httpd` so this PR is outdated

PR #865, opened 2023-06-22
tests: Bring FakeTimers into the present

### LINKED ISSUE #851: Seed project listing visual spec flaky
I often see that the project listing visual diff contains changes. It's either due to the creation order of the fixtures or the order in the httpd backend response is not stable. Either way, we should fix it.

seed-page-1-diff

sebastinez:
Okay I'm not sure if this fixes it, but this PR is still needed, we were faking a time that was before any commits, patches or issues would have been created..
So let's go back to the future 😅 

- Forwards the fakeTimer to January 21 2023 12:00:00

Refs #851

### DIFF
--- a/tests/e2e/project/commit.spec.ts
+++ b/tests/e2e/project/commit.spec.ts
@@ -25,7 +25,7 @@ test("relative timestamps", async ({ page }) => {
   await page.addInitScript(() => {
     window.initializeTestStubs = () => {
       window.e2eTestStubs.FakeTimers.install({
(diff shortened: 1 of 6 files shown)


rudolfs:
I'm not yet convinced that this actually fixes anything. Looking at the commit timestamps in the `source-browsing` fixture and the `GIT_AUTHOR_DATE` `GIT_COMMITTER_DATE` env variables this change doesn't seem necessary. But I need to look at it more deeply.

sebastinez:
I think I can maybe just move some cob creation points more into the past.

PR #866, opened 2023-06-22
Add ability to expand a commit message

### LINKED ISSUE #813: Expand commit message
It's often useful to see the full commit messages in the context of a patch, but it's a bit annoying to have to navigate to the commit pages.

It would be nice if the commit description could be expanded, like in some places on github, via a little `[...]` button next to the summary:

image

The full message would be in a small monospace font.

I'd also consider showing the commits expanded by default in that view, since there's enough room to do so.

sebastinez:
Add ability, if a commit description is available to click a button next to the commit message in a `CommitTeaser`

See https://github.com/radicle-dev/radicle-interface/assets/7912302/1f9bd228-5cf2-4182-b5db-e356f21e0d31

Screenshot:
Bildschirmfoto 2023-06-27 um 11 02 42

Video:
https://github.com/radicle-dev/radicle-interface/assets/7912302/53e8c4c1-776b-4771-9ea6-5ce9eafa4415

Closes #813

### DIFF
--- a/src/views/projects/Commit/CommitTeaser.svelte
+++ b/src/views/projects/Commit/CommitTeaser.svelte
@@ -9,6 +9,8 @@
   import ProjectLink from "@app/components/ProjectLink.svelte";
 
   export let commit: CommitHeader;
+
+  let expandCommitMessage = false;
 </script>
 
 <style>
@@ -29,6 +31,27 @@
   .left {
     padding-left: 1rem;
   }
+  .message {
+    align-items: center;
+    display: flex;
+    flex-direction: row;
+    gap: 0.5rem;
+    margin-bottom: 0.25rem;
+  }
+  .expand-toggle {
+    background-color: var(--color-foreground-2);
+    border: 1px solid var(--color-foreground-5);
+    border-radius: var(--border-radius-tiny);
+    color: var(--color-foreground);
+    cursor: pointer;
+    font-weight: var(--font-weight-medium);
+    height: 12px;
+    line-height: 6px;
+    padding: 0 5px 5px;
+  }
+  .expand-toggle:hover {
+    background-color: var(--color-foreground-5);
+  }
   .right {
     display: flex;
     align-items: center;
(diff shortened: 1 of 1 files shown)


cloudhead:
Nice, can we make the icon smaller? See the github one for eg.
image

sebastinez:
> Nice, can we make the icon smaller? See the github one

Yeah sure, did mimick the GitHub one, see 5d9d368

rudolfs:
Here's some inconsistencies I've spotted. But to avoid double work, maybe @cloudhead can first comment on this before changing anything.

- There's something off with newlines [here](http://localhost:3000/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/9625ee6747ca00f510e32f7b8539de8b968a57ae?tab=commits). Also wonder if the description be rendered as markdown or maybe just render the inline markdown styles as we do in titles?
  Screenshot 2023-06-27 at 11 35 18

- This might be a design/taste thing, but we don't use highlights on button active states anywhere else in the app. Also the hover color is blue, normally we just use a different shade of the button color, like lighter gray. And finally, all buttons in the app have rounded corners, maybe here we could also do 2px rounding? 

  https://github.com/radicle-dev/radicle-interface/assets/158411/f2ab5736-2d3e-4901-866f-3936f3f67b16

cloudhead:
> There's something off with newlines here.

Yeah, probably needs to be in a `<pre>`?

> Also wonder if the description shouldn't be rendered as markdown or maybe just render the inline markdown styles as we do in titles?

I think just the inline styles _might_ make sense, ie. `<code>`, `<em>` etc. Though I would keep it in monospace font. I'm not 100% sold though, GitHub doesn't do it either, and it might be unexpected to see markdown in there. So we could just leave it out for now.

> This might be a design/taste thing..

I think to keep things consistent there should be some rounding and indeed the hover color should be a different grey.

sebastinez:
@rudolfs  wdyt about something like this?
https://github.com/radicle-dev/radicle-interface/assets/7912302/1c4b711b-791e-4ca9-952a-061051234b4e

We have a kind of special case here, that is a button on a lighter background, so if we make it equal to the project header buttons for example, the user doesn't recognize the 3 dots as a button IMO.
So we can show a kind of outline, and by hovering the button takes the same background as the border

rudolfs (approved):

PR #884, opened 2023-06-28
Add `<details>` and `<summary>` and comments to trusted tags

### LINKED ISSUE #883: Markdown comments are not parsed
See this page:

https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/tree/master/radicle-cli/examples/rad-clone-all.md

The `<!-- TODO ... -->` comment should not show up in the doc.

sebastinez:
This PR fixes #883 also handles codespans and code blocks inside of trusted html tags.

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -190,6 +190,10 @@
     margin-bottom: 0;
   }
 
+  .markdown :global(details) {
+    margin-bottom: 0.625rem;
+  }
+
   .markdown :global(blockquote) {
     color: var(--color-foreground-6);
     border-left: 0.3rem solid var(--color-foreground-4);
--- a/src/lib/markdown.ts
+++ b/src/lib/markdown.ts
@@ -3,7 +3,16 @@ import katex from "katex";
 import { marked } from "marked";
 import { isUrl } from "@app/lib/utils";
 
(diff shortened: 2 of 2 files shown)


rudolfs:
Hm, the visual spec changes look suspicious.

sebastinez:
@rudolfs I agree with you on getting the parsing right and opening up to eventual injection vulnerabilities.
I think it makes more sense to let the sanitizer and marked do their thing and just escape some tags we want to display like `<RID>` and `<OID>`, this keeps the workload on the previously mentioned packages, and we can escape a few edge cases like the rid and oid tags.

sebastinez:
Also following the `CommonMark` spec which allows HTML to be rendered as raw HTML which won't be escaped nor parsed by marked https://spec.commonmark.org/0.30/#html-blocks.
So the correct way IMO would be to also use the corresponding HTML tags inside.

So instead of
```
<details>
<summary>🚧 `rad node start` is under construction 🚧</summary>
If the node is not running we can start it by using the `rad node
start` command:
</details>
```

we should advocate for

```
<details>
<summary>🚧 <code>rad node start</code> is under construction 🚧</summary>
If the node is not running we can start it by using the <code>rad node
start</code> command:
</details>
```

PR #887, opened 2023-06-29
Clean up project view props

rudolfs:
We pass down props only where it's needed.

TBD remove the following props from the top-level `ProjectsParams` and `ProjectLoadedParams` types:

```
  hash?: string;
  path?: string;
  peer?: string;
  revision?: string;
  route?: string;
  search?: string;
```


geigerzaehler (approved):
I think we should merge this already. It works and is an improvement. No need to delay this longer to try to fix more stuff. We can tackle the other props in a follow-up PR.

PR #891, opened 2023-07-03
Teardown hanging processes in Vitest

sebastinez:
After setting up the global env for the httpd:api tests, there were tests hanging until the timeout of 10 seconds killed them.
By returning an async function that works as a globalTeardown fn from the globalSetup we can kill them preemptively.


sebastinez:
Okay I'm not really seeing much improvement on the CI, seems like it was more a local env issue.

tests/support/globalSetup.ts
```diff
@@ -64,6 +66,14 @@ export default async function globalSetup(_config: FullConfig): Promise<void> {
   } else {
     await startPalmHttpd();
   }
+
+  return async () => {
+    if (teardownHappened) {
+      throw new Error("GlobalTeardown called twice");
+    }
+    teardownHappened = true;
+    killAllProcesses();
```
geigerzaehler, line 75: `killAllProcesses` is idempotent. (In fact it is actually called twice: once here and in the exit hook). So no need to guard this.

```suggestion
    killAllProcesses();
```

geigerzaehler (approved):
This does fix things for me locally as well. But I can’t explain why CI still hangs for the 10 seconds. Maybe you could try to add more logging when you’re running it on CI?

sebastinez:
So I added also a way to stop a `radicle-node` and a `radicle-httpd` process from a `RadiclePeer`
This seems to allow on one side for more isolation, since we don't share data with the bob and alice nodes once source-browsing is created.
And since the `radicle-node` processes are the ones that seem to make Vitest hang, this should help

rudolfs:
> I'm also thinking we could probably skip the markdown and cobs fixture in the httpd-api tests for now

Yeah, we don't need to create these two fixtures for the httpd-api tests.

tests/support/globalSetup.ts
```diff
@@ -37,14 +36,11 @@ export default async function globalSetup(_config: FullConfig): Promise<void> {
...
 
-    // Workaround for fixing MaxListenersExceededWarning.
-    // Since every prefixOutput stream adds stream listeners that don't autoClose.
-    // TODO: We still seem to have some descriptors left open when running vitest, which we should handle.
     const peerManager = await createPeerManager({
       dataDir: Path.resolve(Path.join(tmpDir, "peers")),
       outputLog: FsSync.createWriteStream(
         Path.join(tmpDir, "peerManager.log"),
-      ).setMaxListeners(20),
+      ).setMaxListeners(16),
```
rudolfs, line 43: Do we still need this at all?

sebastinez: yes, sadly we still need it, not for the reasons in the comment, so I removed it but we still get the warning that we have more than 15 max listeners added.
This is a default by node

> By default EventEmitters will print a warning if more than 10 listeners are added for a particular event. This is a useful default that helps finding memory leaks. The emitter.setMaxListeners() method allows the limit to be modified for this specific EventEmitter instance. The value can be set to Infinity (or 0) to indicate an unlimited number of listeners.
source: https://nodejs.org/api/events.html#emittersetmaxlistenersn

sebastinez:
> > I'm also thinking we could probably skip the markdown and cobs fixture in the httpd-api tests for now
> 
> Yeah, we don't need to create these two fixtures for the httpd-api tests.

Ok this got it even more down, since we don't have to wait for the markdown and cob fixture to be created.

geigerzaehler (approved):
Great detective work on this :detective:! I’m still curious why killing `radicle-node` with SIGKILL did not lead to a proper shutdown. Do you know?

sebastinez:
> Great detective work on this 🕵️! I’m still curious why killing `radicle-node` with SIGKILL did not lead to a proper shutdown. Do you know?

So I'm not 100% sure and I'm not able now to dive into vitest and playwright codebase, but it seems that while Playwright after running the test suites, calls the globalTeardown quickly and exits all the pending processes.
Vitest seems to be more patient and waits that the child processes close themselves before running the globalTeardown function.

PR #893, opened 2023-07-03
Get rid of project-specific routing

rudolfs:
We don't want to have a custom way to do routing within project views.


sebastinez (approved):
Nice! LGTM

PR #899, opened 2023-07-05
Clean up Project Header and SourceBrowsing header props

rudolfs:
This simplifies the Project top level view.

### DIFF
--- a/src/views/projects/Browser.svelte
+++ b/src/views/projects/Browser.svelte
@@ -1,23 +1,31 @@
 <script lang="ts">
-  import type { BaseUrl, Project, Tree } from "@httpd-client";
+  import type { BaseUrl, Project, Remote, Tree } from "@httpd-client";
   import type { BlobResult } from "./router";
+  import type { LoadedSourceBrowsingView } from "@app/views/projects/router";
 
   import * as utils from "@app/lib/utils";
   import { HttpdClient } from "@httpd-client";
 
   import Button from "@app/components/Button.svelte";
   import Placeholder from "@app/components/Placeholder.svelte";
+  import SourceBrowsingHeader from "./SourceBrowsingHeader.svelte";
 
   import BlobComponent from "./Blob.svelte";
   import TreeComponent from "./Tree.svelte";
 
-  export let path: string;
-  export let hash: string | undefined = undefined;
-  export let project: Project;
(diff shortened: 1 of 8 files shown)


src/views/projects/Browser.svelte
```diff
@@ -1,23 +1,32 @@
...
-  export let tree: Tree;
-  export let revision: string | undefined;
+  export let branches: Record<string, string>;
   export let commit: string;
+  export let commitCount: number;
+  export let contributorCount: number;
+  export let defaultBranch: string;
+  export let hash: string | undefined = undefined;
+  export let path: string;
+  export let peer: string | undefined = undefined;
```
geigerzaehler, line 24: Could you remove the default value here (and, if possible, for `hash`)? If we remove the default value, the type checker notifies us when `projects/View` stops providing the value. If that’s the case, we should also remove the property.

As a general rule of thumb you should only provide default values for properties if the component is used in multiple places and for convenience you want to omit some of the properties at some call sites.

src/views/projects/Browser.svelte
```diff
@@ -156,7 +177,7 @@
             {hash}
             blob={blobResult.blob}
             highlighted={blobResult.highlighted}
-            rawPath={utils.getRawBasePath(project.id, baseUrl, commit)} />
+            rawPath={utils.getRawBasePath(projectId, baseUrl, commit)} />
```
geigerzaehler, line 180: I’m not sure whether it’s a good idea to replace the `project` property with the two properties `defaultBranch` and `projectId`. This change shifts the responsibility of unpacking the project object provided by the router from the `Browser` component to the project `View` component. But one of the goals of the routing refactoring is to decouple `View` from the individual view components so that we can change an individual view by only changing the data loading in the router and the component file. In the end we want a simple `View` component that looks much like `App`.

We can tackle this fully at a later stage, but I think it would be good to not move into the opposite direction in the meantime.

geigerzaehler (commented):
Really good change!

src/views/projects/Browser.svelte
rudolfs: https://github.com/radicle-dev/radicle-interface/pull/899/commits/df135ea9872da2e237a0e756f8f1c20d52688e3e

geigerzaehler (approved):

PR #900, opened 2023-07-05
Use global Link component instead of ProjectLink

rudolfs:
Eliminate remaining `ProjectLink` instances. We want to use the generic `Link` component everywhere to keep the router generic.


src/components/Markdown.svelte
```diff
@@ -40,11 +43,15 @@
   function navigateToMarkdownLink(event: any) {
     if (event.target.matches(".file-link")) {
       event.preventDefault();
-      if ($activeRouteStore.resource === "projects") {
-        void updateProjectRoute({
+      void router.push({
+        resource: "projects",
```
geigerzaehler, line 47: I believe we need to add the `peer` and `revision` properties here. The following test fails

1. Go to the awesome-radicle project.
2. Select the peer “dave” and one of the branches starting with “patches/”
3. Click on `macos-launch-agents`
4. Expected to see the the file `launchers/macos/​README.md` for the peer “dave” and the selected branch

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -170,14 +173,19 @@
...
-          revision,
+      <Link
+        route={{
+          resource: "projects",
+          params: {
+            id: projectId,
+            baseUrl,
+            view: { resource: "tree" },
+            path: file.path,
+            revision,
```
geigerzaehler, line 184: We’re missing the peer here

src/views/projects/SourceBrowser/FileLocationChange.svelte
```diff
@@ -49,14 +51,19 @@
...
-          revision,
+      <Link
+        route={{
+          resource: "projects",
+          params: {
+            id: projectId,
+            baseUrl,
+            view: { resource: "tree" },
+            path: file.newPath,
+            revision,
```
geigerzaehler, line 62: We‘re missing the `peer` here

sebastinez (commented):
Really nice idea 👍 

Maybe we could also add the extensions and deprecated options we define inside the `Renderer`?

Also since we are touching marked, we could maybe also look into using a `Marked` instance instead of having different `marked` imports, and maybe have a getter for that instance, on which we call `parse()`?

```
/src/components/InlineMarkdown.svelte
  11,24:     dompurify.sanitize(marked.parseInline(content));

/src/components/Markdown.svelte
  55,7:       marked.parse(content, {

/src/lib/markdown.ts
  96,10:     }. ${marked.parseInline(
```

See: https://github.com/markedjs/marked/pull/2831

rudolfs:
> Really nice idea 👍
> 
> Maybe we could also add the extensions and deprecated options we define inside the `Renderer`?
> 
> Also since we are touching marked, we could maybe also look into using a `Marked` instance instead of having different `marked` imports, and maybe have a getter for that instance, on which we call `parse()`?
> 
> ```
> /src/components/InlineMarkdown.svelte
>   11,24:     dompurify.sanitize(marked.parseInline(content));
> 
> /src/components/Markdown.svelte
>   55,7:       marked.parse(content, {
> 
> /src/lib/markdown.ts
>   96,10:     }. ${marked.parseInline(
> ```
> 
> See: [markedjs/marked#2831](https://github.com/markedjs/marked/pull/2831)

Thanks for the review @sebastinez!

I'll merge this as is for now and leave the feedback for @geigerzaehler as a follow up: https://github.com/radicle-dev/radicle-interface/issues/919.

PR #901, opened 2023-07-05
Remove unnecessary logic in BranchSelector

geigerzaehler:
We remove logic used in the `BranchSelector` component. Instead we restructure the component properties and rely on the `Browser`, `History` and `Commit` components to pass values that result in the same behavior.

The PR also includes a follow-up commit that simplifies the branch loading:

* Don’t load branches in `commits` view. We the branch selector is not enabled there.
* Don’t silently swallow errors when fetching branches. Let the route loading mechanism handle errors.
* Don’t use the default branch as the fallback for `branches`. They are allowed to be empty and the empty case is handled properly now.


rudolfs (approved):
Nice 🧹

PR #909, opened 2023-07-06
Don’t replace path on internal navigation

geigerzaehler:
We don’t replace the path when we navigate internally. This means we keep the hash part of the URL even if it is not part of the route and we can remove the workaround for that.

This also means that we keep the original URL if the parsed route would result in a different URL. This change is visible through a slight adjustments in the tests.


rudolfs (approved):
🧹

PR #914, opened 2023-07-07
Allow httpd API ports to be generated random

### LINKED ISSUE #913: Creating more than one `authenticatedPeer` in the e2e test, messes with e2e tests
I hardcoded the port 8070 in the `authenticatedPeer` fixture, so when we have multiple workers creating `authenticatedPeers` now, this starts to mess with tests..

We probably want to get a random port with `getPort` and assign that to the current `RadiclePeer` which then navigates on that isolated httpd

sebastinez:
To avoid port collisions between tests that can eventually run in parallel, we need to be able to generate random port numbers.

This commit generates random httpd port numbers for inline tests, while keeping the current palm seed with the default port number 8080.

Closes #913

### DIFF
--- a/tests/e2e/project/commits.spec.ts
+++ b/tests/e2e/project/commits.spec.ts
@@ -11,12 +11,12 @@ import { createProject } from "@tests/support/project";
 
 test("peer and branch switching", async ({ page }) => {
   await page.goto(sourceBrowsingUrl);
-  await page.locator('role=link[name="6 commits"]').click();
+  await page.getByRole("link", { name: "6 commits" }).click();
 
   // Alice's peer.
   {
     await page.getByTitle("Change peer").click();
-    await page.locator(`text=${aliceRemote}`).click();
+    await page.getByText(aliceRemote).click();
     await expect(page.getByTitle("Change peer")).toHaveText(
       `  did:key:${aliceRemote
         .substring(8)
@@ -37,7 +37,7 @@ test("peer and branch switching", async ({ page }) => {
     await expect(earliestCommit).toContainText("36d5bbe");
 
     await page.getByTitle("Change branch").click();
-    await page.locator("text=feature/branch").click();
+    await page.getByText("feature/branch").click();
 
     await expect(page.getByTitle("Current branch")).toContainText(
       "feature/branch 1aded56",
@@ -46,7 +46,7 @@ test("peer and branch switching", async ({ page }) => {
     await expect(page.locator(".history .teaser")).toHaveCount(9);
 
     await page.getByTitle("Change branch").click();
-    await page.locator("text=orphaned-branch").click();
+    await page.getByText("orphaned-branch").click();
 
     await expect(page.getByTitle("Current branch")).toContainText(
       "orphaned-branch af3641c",
@@ -58,7 +58,7 @@ test("peer and branch switching", async ({ page }) => {
   // Bob's peer.
   {
     await page.getByTitle("Change peer").click();
-    await page.locator(`text=${bobRemote}`).click();
+    await page.getByText(bobRemote).click();
     await expect(page.getByTitle("Change peer")).toContainText(
       ` did:key:${bobRemote.substring(8).substring(0, 6)}…${bobRemote.slice(
         -6,
@@ -89,7 +89,7 @@ test("peer and branch switching", async ({ page }) => {
 
 test("expand commit message", async ({ page }) => {
   await page.goto(sourceBrowsingUrl);
-  await page.locator('role=link[name="6 commits"]').click();
+  await page.getByRole("link", { name: "6 commits" }).click();
   const commitToggle = page
     .locator("div")
     .filter({ hasText: /^Add a C source file and its binary …$/ })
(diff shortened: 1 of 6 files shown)


tests/support/peerManager.ts
```diff
@@ -184,25 +186,44 @@ export class RadiclePeer {
     });
   }
 
-  public async startHttpd(port: number) {
+  public async startHttpd(port?: number) {
+    if (!port) {
+      port = await getPort();
+    }
+    this.#httpdBaseUrl = {
+      hostname: "0.0.0.0",
```
geigerzaehler, line 194: Shouldn’t this be `127.0.0.1`?

tests/support/peerManager.ts
```diff
@@ -312,10 +333,23 @@ export class RadiclePeer {
   }
 
   public uiUrl(): string {
-    return `/seeds/127.0.0.1:8080`;
+    if (!this.#httpdBaseUrl) {
+      throw new Error("No httpd service running");
+    }
+
+    return `/seeds/${this.#httpdBaseUrl.hostname}`;
```
geigerzaehler, line 340: Shouldn’t this be 
```suggestion
    return `/seeds/${this.#httpdBaseUrl.hostname}:${this.#httpdBaseUrl.port}`;
```

tests/e2e/project/commits.spec.ts
```diff
@@ -138,12 +138,13 @@ test("pushing changes while viewing history", async ({ page, peerManager }) => {
     name: "alice",
     gitOptions: gitOptions["alice"],
   });
-  await alice.startHttpd(8090);
+  await alice.startHttpd();
+  const aliceHttpdUrl = `${alice.httpdBaseUrl.hostname}:${alice.httpdBaseUrl.port}`;
   await alice.startNode();
   const { rid, projectFolder } = await createProject(alice, "alice-project");
-  await page.goto(`/seeds/127.0.0.1:8090/${rid}`);
+  await page.goto(`/seeds/${aliceHttpdUrl}/${rid}`);
```
geigerzaehler, line 145: Couldn’t you do the following
```suggestion
  await page.goto(`${alicel.uiUrl()}/${rid}`);
```

tests/support/peerManager.ts
```diff
@@ -184,25 +186,44 @@ export class RadiclePeer {
...
 
     await waitOn({
-      resources: [`tcp:127.0.0.1:${port}`],
+      resources: [
+        `tcp:${this.#httpdBaseUrl.hostname}:${this.#httpdBaseUrl.port}`,
+      ],
       timeout: 7000,
     });
+
+    return this.#httpdBaseUrl;
```
geigerzaehler, line 210: I don’t think this return value is used anywhere. You can remove it

tests/e2e/project/commits.spec.ts
sebastinez: Yeah that's a good one 👍

tests/support/peerManager.ts
sebastinez: Yeah probably better since we eventually use random ports

tests/support/peerManager.ts
sebastinez: AFAIK we usually define the UI with localhost or 127.0.0.1 but the httpd we usually bind to 0.0.0.0
We also use 0.0.0.0 as default in heartwood, so probably better to use 0.0.0.0 here.
https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/tree/master/radicle-httpd/src/main.rs#L76

tests/support/peerManager.ts
sebastinez: yeah is a leftover probably doesn't need it

tests/e2e/project/commits.spec.ts
sebastinez: Done ✅

tests/support/peerManager.ts
sebastinez: Done ✅

tests/support/peerManager.ts
geigerzaehler: > We also use 0.0.0.0 as default in heartwood, so probably better to use 0.0.0.0 here.
> https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/tree/master/radicle-httpd/src/main.rs#L76

In general, the IP address 0.0.0.0 is [only used as a source address](https://en.wikipedia.org/wiki/0.0.0.0).  Here, you’re using it as a destination address. While this works with the browser network stack and also a lot of other tools it’s very unusual to see this and not standard.

In the code you linked to, the address 0.0.0.0 tells the networking stack that it should bind to all network interfaces on the host. This includes the IP address 127.0.0.1 so it is safe (and the standard) to use that as the destination address.

tests/e2e/project/issues.spec.ts
```diff
@@ -48,7 +48,7 @@ test("test issue editing failing", async ({ page, authenticatedPeer }) => {
   );
 
   await page.goto(
-    `/seeds/127.0.0.1:8070/${rid}/issues/d316f7a90a40dacbfb8728044bad50c9f71d44ba`,
+    `/seeds/${authenticatedPeer.httpdBaseUrl.hostname}:${authenticatedPeer.httpdBaseUrl.port}/${rid}/issues/d316f7a90a40dacbfb8728044bad50c9f71d44ba`,
```
geigerzaehler, line 51: I think you can used `authenticatedPeer.uiUrl()` here as well.

tests/support/peerManager.ts
sebastinez: Thanks @geigerzaehler for the explainer, will change it!

tests/e2e/project/issues.spec.ts
sebastinez: nice, changed it 👍

tests/support/peerManager.ts
sebastinez: done

rudolfs (approved):
👏

PR #920, opened 2023-07-10
Update patch review schema

sebastinez:
Heartwood received changes how to handle patch reviews, so we have to update our zod schemas to not break patch listing

Here are the changes introduced to heartwood
https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/commits/45e87a63b94fb2fbc5daba6be68877f8f9b535ee

I removed the `codeLocation` and the `codeComment` schema and interface since it's not needed currently, when code review lands in heartwood we'll probably need to add the final implementation

- [x] This PR is in draft until seed.radicle.xyz is updated to at least commit 45e87a6


rudolfs (approved):
🐛 🧹

PR #921, opened 2023-07-10
Stop handling URL hash through routing

geigerzaehler:
We stop handling the URL hash through the routing system. Hashes are local to the current view and can be handled by the shown components.


rudolfs (approved):
🧹

PR #922, opened 2023-07-10
Use ESlint warnings and Prettier config

geigerzaehler:
* We remove all styling related lints and use `eslint-config-prettier`. There’s no need to bother developers with styling lints because Prettier already takes care of that. (This is the [recommended integration with Prettier][1])

* We use the “warning” level for lints that are code style related. This allows developers to distinguish issues that need to be fixed now (like type errors) and issues that can be fixed later (like style errors). We do enforce all lints on CI by settings `--max-warnings=0`.

[1]: https://prettier.io/docs/en/integrating-with-linters.html


rudolfs (approved):
💅

PR #926, opened 2023-07-11
Remove redundant project route props

geigerzaehler:
This PR consists of two commits that remove redundant properties from the project route types.

We remove the requirement that `LoadedSourceBrowsingView` needs to be assignable to `ProjectParams["view"]`. This decouples `ProjectRoute` and `ProjectLoadedRoute` and lets us modify them independently.

I did not use a common `routeParamsView` function because I prefer the flexibility of being able to modify them independently while this area of code changes a lot.


rudolfs (approved):
🧹

PR #930, opened 2023-07-11
Set page title in the router

rudolfs:
This improves UX making it easier to find the tab that I'm looking for when I have multiple tabs open.

We can come up with a more elaborate structure, but for now this might be sufficient.

Screenshot 2023-07-11 at 17 30 59

### DIFF
--- a/index.html
+++ b/index.html
@@ -93,7 +93,7 @@
       import { Buffer } from "buffer";
       window.Buffer = Buffer;
     </script>
-    <title>Radicle Interface</title>
+    <title>Radicle</title>
   </head>
   <body>
     <script type="module" src="/src/index.ts"></script>
--- a/src/lib/router.ts
+++ b/src/lib/router.ts
@@ -8,15 +8,13 @@ import * as utils from "@app/lib/utils";
 import { config } from "@app/lib/config";
 import {
   projectRouteToPath,
+  projectTitle,
   resolveProjectRoute,
 } from "@app/views/projects/router";
 import { loadRoute } from "@app/lib/router/definitions";
 import { seedPath } from "@app/views/seeds/router";
 
 export { type Route };
-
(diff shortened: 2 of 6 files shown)


index.html
```diff
@@ -93,7 +93,6 @@
       import { Buffer } from "buffer";
       window.Buffer = Buffer;
     </script>
-    <title>Radicle Interface</title>
```
geigerzaehler, line 96: I’d keep this and set it to “Radicle”. This way you’ll have a title until all the Javascript is loaded and the app has booted.

src/lib/router.ts
```diff
@@ -112,10 +109,61 @@ async function navigate(
...
+      title.push("issues");
+    } else if (loadedRoute.params.view.resource === "patches") {
+      title.push(loadedRoute.params.project.name);
+      title.push("patches");
+    } else if (loadedRoute.params.view.resource === "patch") {
+      title.push(loadedRoute.params.view.params.loadedPatch.title);
+      title.push("patch");
+    } else {
+      utils.unreachable(loadedRoute.params.view);
+    }
```
geigerzaehler, line 159: To keep things modular I would move this code into a function in `app/views/project/router`. This way we keep the size of this code in check and we only have to change one file if we want to adjust the project routing.

index.html
rudolfs: 2b7afe4917f870515910b624295a725559da4e48

src/lib/router.ts
rudolfs: 3b1495657fa10d51157d6d2bcc76d8709490fee9

cloudhead:
For the root (`/`) I'd just use "Radicle" as the title. For the rest I think it's good enough for now :+1:

geigerzaehler (approved):

PR #933, opened 2023-07-12
Don’t intercept navigation to external URLs in Markdown

### LINKED ISSUE #928: Links to external pages don't work in markdown
Reproduction of issue

1. Go to https://app.radicle.xyz/seeds/seed.rhizoma.dev/rad:z3yQUb9HDAC7TQrUDGkQsXDsYFj9G
2. Try to click on `Radicle jetbrains plugin`
3. See error in console 
```
Uncaught (in promise) Error: Cannot navigate to other origin
    at M_ (index-e6b5f19a.js:17:79299)
    at HTMLDivElement.d (index-e6b5f19a.js:299:26802)
```

This seems to originate in https://github.com/radicle-dev/radicle-interface/blob/50dd25a6b892b76e64e50816e5dd13f730c15661/src/lib/router.ts#L51

Seems weird that we don't allow users to navigate from our Markdown to another page 🤔

geigerzaehler:
Fixes #928

Also includes a commit that improves Playwright route mocking:

We determine URLs to mock with a URL matcher instead of mocking every URL and determining the behavior in the route handler. This should be slightly faster and should result in better debugging experience because we don’t come across the `route.continue()` event that often.

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -45,8 +45,10 @@
       return;
     }
 
-    event.preventDefault();
-    void router.navigateToUrl("push", url);
+    if (url.origin === window.origin) {
+      event.preventDefault();
+      void router.navigateToUrl("push", url);
+    }
   }
 
   function render(content: string): string {
--- a/tests/e2e/project.spec.ts
+++ b/tests/e2e/project.spec.ts
@@ -438,14 +438,11 @@ test.describe("browser error handling", () => {
 });
 
 test("external markdown link", async ({ page }) => {
-  await page.route("http://github.github.com/**", route => {
+  await page.route("https://example.com/**", route => {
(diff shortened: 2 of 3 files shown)


rudolfs:
In this case we actually want to see the black square image in the visual spec, because that is how the spec and repository are set up. We might need to add an exception in the mocking for this test or find another workaround.

This is what we see when we run the fixture locally outside the test environment:

Screenshot 2023-07-12 at 16 34 13

geigerzaehler:
> In this case we actually want to see the black square image in the visual spec, because that is how the spec and repository are set up.

Fixed that

> Now that I think about it, we might want to replace the mock with an image that is more obvious that it is a mock, so that it is more explicit and easy to detect.

Sounds like a good idea. I’ll follow up with this in a separate PR

rudolfs (approved):

PR #934, opened 2023-07-12
Cleanup shared peer and revision props for project route

geigerzaehler:
We move the `peer` and `revision` properties from the shared `ProjectLoadedParams` interface to the `tree` and `history` subroutes where they are needed.


rudolfs:
There are some unexpected visual changes. Please have a look.

rudolfs (approved):

PR #939, opened 2023-07-12
Switch to flat project route type

geigerzaehler:
We switch to a project route type that is just a union. This has the following advantages

* It’s much easier to construct project routes. Fewer properties are needed and there are no nested objects.
* We eliminate the implicitly shared properties `path`, `peer`, `revision` and `route` which only belonged to the tree route.
* The structure of the project route is more regular and thus easier to understand.

### DIFF
--- a/src/lib/router.ts
+++ b/src/lib/router.ts
@@ -187,19 +187,13 @@ function pathToRoute(url: URL): Route | null {
         const baseUrl = extractBaseUrl(hostAndPort);
         const id = segments.shift();
         if (id) {
-          const params = resolveProjectRoute(baseUrl, id, segments, url.search);
-          if (params) {
-            return {
-              resource: "projects",
-              params: params,
-            };
-          }
-          return null;
+          return resolveProjectRoute(baseUrl, id, segments, url.search);
+        } else {
+          return {
+            resource: "seeds",
+            params: { baseUrl, projectPageIndex: 0 },
+          };
         }
-        return {
-          resource: "seeds",
-          params: { baseUrl, projectPageIndex: 0 },
-        };
       }
       return null;
     }
@@ -235,8 +229,17 @@ export function routeToPath(route: Route): string {
     return seedPath(route.params.baseUrl);
   } else if (route.resource === "loadError") {
     return "";
-  } else if (route.resource === "projects") {
-    return projectRouteToPath(route.params);
+  } else if (
+    route.resource === "project.tree" ||
+    route.resource === "project.history" ||
+    route.resource === "project.commit" ||
+    route.resource === "project.issues" ||
+    route.resource === "project.newIssue" ||
+    route.resource === "project.issue" ||
+    route.resource === "project.patches" ||
+    route.resource === "project.patch"
+  ) {
+    return projectRouteToPath(route);
   } else if (route.resource === "booting") {
     return "";
   } else if (route.resource === "notFound") {
--- a/src/lib/router/definitions.ts
+++ b/src/lib/router/definitions.ts
@@ -55,8 +55,17 @@ export async function loadRoute(route: Route): Promise<LoadedRoute> {
     return await loadSeedRoute(route.params);
   } else if (route.resource === "home") {
     return await loadHomeRoute();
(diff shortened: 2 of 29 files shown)


src/lib/router.ts
```diff
@@ -235,8 +229,8 @@ export function routeToPath(route: Route): string {
     return seedPath(route.params.baseUrl);
   } else if (route.resource === "loadError") {
     return "";
-  } else if (route.resource === "projects") {
-    return projectRouteToPath(route.params);
+  } else if ("project" in route) {
```
rudolfs, line 232: Maybe add a comment explaining how this works.

src/lib/router/definitions.ts
```diff
@@ -55,8 +55,8 @@ export async function loadRoute(route: Route): Promise<LoadedRoute> {
     return await loadSeedRoute(route.params);
   } else if (route.resource === "home") {
     return await loadHomeRoute();
-  } else if (route.resource === "projects") {
-    return await loadProjectRoute(route.params);
+  } else if ("project" in route) {
```
rudolfs, line 58: Here too.

src/views/home/Index.svelte
```diff
@@ -70,14 +70,9 @@
...
-              resource: "projects",
-              params: {
-                view: { resource: "tree" },
-                id: project.id,
-                baseUrl,
-                peer: undefined,
-                revision: undefined,
-              },
+              resource: "project.tree",
+              project: project.id,
```
rudolfs, line 74: Wonder if this property shouldn't be `projectId`, so it's clear what it is?

src/views/home/Index.svelte
```diff
@@ -70,14 +70,9 @@
...
+              seed: baseUrl,
```
rudolfs, line 75: This could also be your own node which might not be a public seed. Could maybe use `node`?

rudolfs: We should maybe discuss this, because on a high level we use peer/node/seed interchangeably. But in the code there is a distinction. In the context of browsing a project we're interested in a node's httpd base url. The node can be either our own (private) or a public seed, but it's always a node. And when selecting a peer in the peer selector, we're talking about the nodeId of the Radicle network peer.

src/views/home/Index.svelte
geigerzaehler: I was thinking about that as well. In the end I went for `project` because of the following reasons:
* It’s typed as a string so you can’t assign a `Project` inadvertently.
* This property is part of a `Route` object where we always deal with identifiers.
* We use short forms for other identifiers as well: `commit` instead of `commitId`, `peer` instead of `peerId`, etc.

src/lib/router.ts
geigerzaehler: Instead of explaining I chose the more straightforward but more verbose implementation: d751e7c09a0d8513126d7b8c6cdc4bd9309dcf79

src/views/home/Index.svelte
geigerzaehler: Good point d751e7c09a0d8513126d7b8c6cdc4bd9309dcf79

src/views/projects/router.ts
```diff
@@ -101,6 +104,12 @@ export type LoadedSourceBrowsingView =
       totalCommitCount: number;
     };
 
+interface LoadedSourceBrowsingParams {
+  loadedBranches: Record<string, string> | undefined;
+  loadedPeers: Remote[];
+  loadedTree: Tree;
```
rudolfs, line 110: We probably should rename these as a follow-up.

geigerzaehler: Yes

src/views/projects/router.ts
```diff
@@ -462,11 +473,11 @@ function sanitizeQueryString(queryString: string): string {
 }
 
 export function resolveProjectRoute(
-  baseUrl: BaseUrl,
-  id: string,
+  seed: BaseUrl,
```
rudolfs, line 476: Here.

rudolfs (approved):
🧹 🚀

PR #942, opened 2023-07-13
Make diff line selectable and shareable

xphoniex:
Right now I'm using screenshots to reference a specific part of a diff. A better option would be a link, like GitHub.

Screenshot from 2023-07-13 21-52-53

For example the link for the picture would be:

`/seeds/radicle.local/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/e6d27cade05288b5bfa79f0e3c47bb07fdc3c2a0?diff=1ddfa4fe94adca5096a82316fbaf7b147e3b5d9a..5242a14b5e9fefd07db040940223b7f37ba72988#radicle-cli/src/commands/patch/edit.rsH1R9`

0. Right now, the link wouldn't scroll to the line, as we load things in 2 stages, but it shouldn't matter as we're eventually moving to a single stage. Scrolls on GitHub doesn't work that well either anyway, highlighting is the most important part.

1. GitHub uses a different format, `#diff-86922c6cf991e9927a1a37a653efc6e9c1db5dfacb4a99a772798d30d4b0f3dbR69`, we can achieve a similar format by using a hash function.

2. I'm using a `resetHighlights` fn to undo the selection of other lines (if any), is there a better pattern for this?  also if I move the line `line.highlighted = true;` inside a function and call it with `fn(line)`, it wouldn't have the same effect, and I don't know why, any ideas?

3. Since in `Changeset.svelte`, each diff is a separate component, selecting in one file, doesn't undo select in another. Somehow a selected-file var has to be shared among `FileDiff`s, any thoughts how to do this? We can use the first part of  `window.location.hash` if we don't hash it, but something needs to re-trigger render of the components.

4. Do we have a way of adding hash to `window.location` right now?

### DIFF
--- a/src/views/projects/SourceBrowser/FileDiff.svelte
+++ b/src/views/projects/SourceBrowser/FileDiff.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { onDestroy, onMount } from "svelte";
   import type {
     BaseUrl,
     DiffAddedDeletedModifiedChangeset,
@@ -16,6 +17,58 @@
   export let projectId: string;
 
   let collapsed = false;
+  let selection: Selection | undefined = undefined;
+
+  onMount(() => {
+    window.addEventListener("click", deselectHandler);
+    window.addEventListener("hashchange", updateSelection);
+
+    updateSelection();
+
+    if (selection) {
+      document
+        .getElementById(
+          [
+            file.path,
+            "H" + selection.startHunk,
+            "L" + selection.startLine,
+          ].join("-"),
+        )
+        ?.scrollIntoView();
+    }
+  });
+
+  onDestroy(() => {
+    window.removeEventListener("click", deselectHandler);
+    window.removeEventListener("hashchange", updateSelection);
+  });
+
+  function deselectHandler(e: MouseEvent) {
+    if (
+      !(
+        e.target instanceof HTMLElement &&
+        (e.target.hasAttribute("data-line-type") ||
+          e.target.classList.contains("diff-line-content"))
+      )
+    ) {
+      updateHash("");
+    }
+  }
+
+  function updateSelection() {
+    const fragment = window.location.hash.substr(1);
+    const match = fragment.match(/(.+):H(\d+)L(\d+)(H(\d+)L(\d+))?/);
+    if (match && match[1] === file.path) {
+      selection = {
+        startHunk: parseInt(match[2]),
+        startLine: parseInt(match[3]),
+        endHunk: match[4] ? parseInt(match[5]) : undefined,
+        endLine: match[4] ? parseInt(match[6]) : undefined,
+      };
+    } else {
+      selection = undefined;
+    }
+  }
 
   function lineNumberR(line: HunkLine): string | number {
     switch (line.type) {
@@ -45,7 +98,18 @@
     }
   }
 
-  function lineSign(line: HunkLine): string {
+  function lineSign(line: HunkLine, hashSelected = false): string {
+    if (hashSelected) {
+      switch (line.type) {
+        case "addition": {
+          return "@+";
+        }
+        case "deletion": {
+          return "@-";
+        }
+      }
+      return "@";
+    }
     switch (line.type) {
       case "addition": {
         return "+";
@@ -57,6 +121,101 @@
         return "-";
       }
     }
+  }
+
+  function isLineSelected(
+    selection: Selection | undefined,
+    hunkIdx: number,
+    lineIdx: number,
+  ): boolean {
+    if (!selection) {
+      return false;
+    }
+
+    if (selection.endHunk !== undefined && selection.endLine !== undefined) {
+      return (
+        hunkIdx >= selection.startHunk &&
+        hunkIdx <= selection.endHunk &&
+        (hunkIdx === selection.startHunk
+          ? lineIdx >= selection.startLine
+          : true) &&
+        (hunkIdx === selection.endHunk ? lineIdx <= selection.endLine : true)
+      );
+    } else {
+      return hunkIdx === selection.startHunk && lineIdx === selection.startLine;
+    }
+  }
+
+  function hashFromSelection(
+    hunkIdx: number,
+    lineIdx: number,
+    event: MouseEvent,
+  ): string {
+    const path = file.path;
+    // single line selection
+    if (!event.shiftKey) {
+      return path + ":H" + hunkIdx + "L" + lineIdx;
+    }
+
+    if (!selection) {
+      return "";
+    }
+
+    // range selection
+    if (hunkIdx === selection.startHunk) {
+      if (lineIdx >= selection.startLine) {
+        return `${path}:H${hunkIdx}L${selection.startLine}H${hunkIdx}L${lineIdx}`;
+      } else {
+        return `${path}:H${hunkIdx}L${lineIdx}H${hunkIdx}L${selection.startLine}`;
+      }
+    } else if (hunkIdx < selection.startHunk) {
+      return `${path}:H${hunkIdx}L${lineIdx}H${selection.startHunk}L${selection.startLine}`;
+    } else {
+      return `${path}:H${selection.startHunk}L${selection.startLine}H${hunkIdx}L${lineIdx}`;
+    }
+  }
+
+  function selectLine(hunkIdx: number, lineIdx: number, event: MouseEvent) {
+    updateHash(hashFromSelection(hunkIdx, lineIdx, event));
+  }
+
+  function updateHash(newHash: string) {
+    if (newHash !== "") {
+      window.location.hash = newHash;
+    } else {
+      window.history.replaceState(
+        window.history.state,
+        "",
+        window.location.pathname + window.location.search,
+      );
+      selection = undefined;
+    }
+  }
+
+  function headerSign(
+    selection: Selection | undefined,
+    hunkIdx: number,
+  ): string {
+    if (!selection) {
+      return "";
+    }
+
+    if (
+      selection.endHunk !== undefined &&
+      hunkIdx > selection.startHunk &&
+      hunkIdx <= selection.endHunk
+    ) {
+      return "@@@";
+    }
+
+    return "";
+  }
+
+  interface Selection {
+    startHunk: number;
+    startLine: number;
+    endHunk: number | undefined;
+    endLine: number | undefined;
   }
 </script>
 
@@ -112,19 +271,45 @@
   .diff-line {
     vertical-align: top;
   }
-  .diff-line[data-type="+"] > * {
+  .diff-line[data-line-type="+"] > * {
     color: var(--color-positive-6);
     background-color: var(--color-positive-2);
   }
-  .diff-line[data-type="-"] > * {
+  .diff-line[data-line-type="-"] > * {
     color: var(--color-negative-6);
     background-color: var(--color-negative-2);
+  }
+  .diff-line[data-line-type="@"] > * {
+    color: var(--color-foreground-6);
+    background-color: var(--color-foreground-4);
+  }
+  .diff-line[data-line-type="@+"] > * {
+    color: var(--color-positive-6);
+    background-color: var(--color-positive-4);
+  }
+  .diff-line[data-line-type="@-"] > * {
+    color: var(--color-negative-6);
+    background-color: var(--color-negative-4);
+  }
+  .diff-line[data-line-type="@"] {
+    border-left: 4px solid var(--color-primary);
+  }
+  .diff-line[data-line-type="@+"] {
+    border-left: 4px solid var(--color-primary);
+  }
+  .diff-line[data-line-type="@-"] {
+    border-left: 4px solid var(--color-primary);
+  }
+  .diff-line[data-line-type="@@@"] {
+    border-left: 4px solid var(--color-primary);
+    background-color: var(--color-foreground-4);
   }
   .diff-line-number {
     text-align: right;
     user-select: none;
     line-height: 1.5rem;
     min-width: 3rem;
+    cursor: pointer;
   }
   .diff-line-number.left {
     padding: 0 0.5rem 0 0.75rem;
@@ -190,22 +375,37 @@
       {#if file.diff.type === "plain"}
         {#if file.diff.hunks.length > 0}
           <table class="diff">
-            {#each file.diff.hunks as hunk}
-              <tr class="diff-line">
+            {#each file.diff.hunks as hunk, hunkIdx}
+              <tr
+                class="diff-line"
+                data-line-type={headerSign(selection, hunkIdx)}>
                 <td colspan={2} />
                 <td colspan={6} class="diff-expand-header">
                   {hunk.header}
                 </td>
               </tr>
-              {#each hunk.lines as line}
-                <tr class="diff-line" data-expanded data-type={lineSign(line)}>
-                  <td class="diff-line-number left" data-type={lineSign(line)}>
+              {#each hunk.lines as line, lineIdx}
+                <tr
+                  class="diff-line"
+                  data-expanded
+                  data-line-type={lineSign(
+                    line,
+                    isLineSelected(selection, hunkIdx, lineIdx),
+                  )}>
+                  <td
+                    id={[file.path, "H" + hunkIdx, "L" + lineIdx].join("-")}
+                    class="diff-line-number left"
+                    on:click={e => selectLine(hunkIdx, lineIdx, e)}
+                    data-line-type={lineSign(line)}>
                     {lineNumberL(line)}
                   </td>
-                  <td class="diff-line-number right" data-type={lineSign(line)}>
+                  <td
+                    class="diff-line-number right"
+                    on:click={e => selectLine(hunkIdx, lineIdx, e)}
+                    data-line-type={lineSign(line)}>
                     {lineNumberR(line)}
                   </td>
-                  <td class="diff-line-type" data-type={line.type}>
+                  <td class="diff-line-type" data-line-type={line.type}>
                     {lineSign(line)}
                   </td>
                   <td class="diff-line-content">{line.line}</td>
--- a/tests/e2e/project.spec.ts
+++ b/tests/e2e/project.spec.ts
@@ -6,6 +6,7 @@ import {
   bobHead,
   bobRemote,
   expect,
+  cobUrl,
   markdownUrl,
   sourceBrowsingRid,
   sourceBrowsingUrl,
@@ -448,4 +449,14 @@ test("external markdown link", async ({ page }) => {
   await page.getByRole("link", { name: "https://example.com" }).click();
   await page.pause();
   await expect(page).toHaveURL("https://example.com");
+});
+
+test("diff selection de-select", async ({ page }) => {
+  await page.goto(
+    `${cobUrl}/patches/013f8b2734df1840b2e33d52ff5632c8d66b199a?tab=files#README.md:H0L0H0L3`,
+  );
+  await page.getByText("Add subtitle to README").click();
+  await expect(page).toHaveURL(
+    `${cobUrl}/patches/013f8b2734df1840b2e33d52ff5632c8d66b199a?tab=files`,
+  );
 });
--- a/tests/visual/project.spec.ts
+++ b/tests/visual/project.spec.ts
@@ -1,6 +1,7 @@
 import {
   test,
   expect,
+  cobUrl,
   sourceBrowsingUrl,
   aliceRemote,
 } from "@tests/support/fixtures.js";
@@ -48,5 +49,22 @@ test("commit page", async ({ page }) => {
     )}/commits/1aded56c3ad55299df9f06c326af50b802a05949`,
   );
   await expect(page.getByText("subconscious.txt added")).toBeVisible();
+  await expect(page).toHaveScreenshot({ fullPage: true });
+});
+
+test("diff selection", async ({ page }) => {
+  await page.addInitScript(() => {
+    window.initializeTestStubs = () => {
+      window.e2eTestStubs.FakeTimers.install({
+        now: new Date("November 24 2022 12:00:00").valueOf(),
+        shouldClearNativeTimers: true,
+        shouldAdvanceTime: false,
+      });
+    };
+  });
+
+  await page.goto(
+    `${cobUrl}/patches/013f8b2734df1840b2e33d52ff5632c8d66b199a?tab=files#README.md:H0L0H0L3`,
+  );
   await expect(page).toHaveScreenshot({ fullPage: true });
 });


geigerzaehler:
Hey @xphoniex! What you’re trying to implement looks quite similar to what we already have for the source browser: URLs with line fragments like https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z4V1sjrXqjvFdnCUbxPFqd5p4DtH5/tree/master/src/views/projects/Blob.svelte#L32.

Take a look at how the `Blob` component implements highlighting and links. It uses anchor elements (https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z4V1sjrXqjvFdnCUbxPFqd5p4DtH5/tree/master/src/lib/syntax.ts#L138) and let’s the browser do much of the work. A similar implementation for shareable links to hunk lines should resolve the problems that you ran into.

xphoniex:
Thanks, I'd already looked into `Blob.svelte` before implementing this and thought the logic is implemented on [this](https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z4V1sjrXqjvFdnCUbxPFqd5p4DtH5/tree/master/src/views/projects/Blob.svelte#L259) line, which is imported from `hast-util-to-html`.

Can you briefly explain how `createLine` is used? I can't find any line using it in `Blob.svelte`, and grep doesn't return much either.

geigerzaehler:
> Can you briefly explain how `createLine` is used? I can't find any line using it in `Blob.svelte`, and grep doesn't return much either.

Sure! The overall process for rendering source code goes like this: Retrieve the source code as a string and [run the highlighter](https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z4V1sjrXqjvFdnCUbxPFqd5p4DtH5/tree/master/src/views/projects/router.ts#L229). The highlighter gives you a [Hypertext Abstract Syntax Tree](https://github.com/syntax-tree/hast) (HAST), so basically HTML. Then we put the HTML representation of the highlighted code into [`lineNumbersGutter`](https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z4V1sjrXqjvFdnCUbxPFqd5p4DtH5/tree/master/src/views/projects/Blob.svelte#L27). This function transforms the HTML so that each line of code is prefixed with the line number, which also is an anchor element. The transformation happens by manipulating the HAST representation, specifically in `createLine`. Finally we render the transformed HAST representation to an HTML string that we can pass to Svelte.

All of that is not really relevant for your use case though, since we don’t do any source highlighting and don’t use HAST in `FileDiff`. What is important is that in the source code view we have an anchor that points to the line:

image

So what you can do is to also wrap the line numbers in `FileDiff` in anchor elements and use the corresponding element IDs. For the highlighting, take a look at `Blob` to see how that is achieved.

xphoniex:
There are two major issues with anchor element approach:
- It can only select a single line number, as opposed to a range, which github offers:
Screenshot from 2023-07-14 15-39-35

- It forces scroll on click, which is annoying, won't work with ranges, and will hide the context (any code above selection). Again, referencing github, you can freely select any line you want and it won't initiate a scroll.

So, I'm doing it manually here, and made some progress. Please review & let me know if any part of the code is not idiomatic.

src/views/projects/SourceBrowser/Changeset.svelte
```diff
@@ -78,14 +106,26 @@
...
       {revision}
       headerBadgeCaption="deleted" />
   {/each}
   {#each diff.modified as file}
-    <FileDiff {projectId} {baseUrl} {file} {revision} />
+    <FileDiff
+      isSelected={file.path == selectedFilename}
+      selectedHunkLine={file.path == selectedFilename
+        ? selectedHunkLine
+        : undefined}
```
xphoniex, line 124: is it possible to reference own property? in that case we can re-write:

```
 selectedHunkLine={isSelected ? selectedHunkLine : undefined}
```

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -193,19 +216,62 @@
...
-                <tr class="diff-line" data-expanded data-type={lineSign(line)}>
-                  <td class="diff-line-number left" data-type={lineSign(line)}>
+              {#each hunk.lines as line, lineIdx}
+                {@const hunkLineSelected =
+                  selectedHunkLine == "H" + hunkIdx + "L" + lineIdx}
+                {#if hunkLineSelected}
+                  {#if !initialScroll}
+                    {(setTimeout(function () {
+                      const target = document.getElementById(
+                        file.path + "::H" + hunkIdx + "L" + lineIdx,
```
xphoniex, line 233: - scroll now works, regardless of two stage loading, as it's initiated here manually 
- notice, target id is actually `::H...` whereas hash would be `:H`, that's because I couldn't `e.preventDefault()` or `e.stopPropagation()` inside `updateSelectedLineId`, so now initial scroll would work which is desired behavior, but any further click on lines would force scroll, which is undesired

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -193,19 +216,62 @@
...
+                      );
+                      target.scrollIntoView();
+                      initialScroll = true;
+                    }, 200),
+                    "")}
+                  {/if}
+                {:else}
+                  {((initialScroll = true), "")}
```
xphoniex, line 241: is code like this acceptable?

xphoniex:
`Blob.svelte` color palette doesn't look very good:

Screenshot from 2023-07-14 20-17-51

Keeping it white/black for now:

white

xphoniex:
- added support for multi-line (range) selection, click on start, shift+click on end of range.

- updated color palette since it was not very readable:

Screenshot from 2023-07-15 10-06-48

Options:

1. `bg=positive|negative,color=foreground-4`

bg=positive|negative,color=foreground-4

2. `bg=positive|negative,color=positive-6|negative-6`

bg=positive|negative,color=positive-6|negative-6

3. `bg=positive-6|negative-6,color=positive|negative` (chosen)

bg=positive-6|negative-6,color=positive|negative

Here's a live example: [https://radicle-interface-3r3mwyfc7-radicle.vercel.app/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/4ad455bf0e4527c7f6852059f691656cd2bade92?diff=d5bc7386211318ac7c7652d0bdd53f048f0598b1..eb701be033610992ed0a1a845e4ab9098dbd53bd#radicle-node/src/tests.rs:H1L1L9](https://radicle-interface-3r3mwyfc7-radicle.vercel.app/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/4ad455bf0e4527c7f6852059f691656cd2bade92?diff=d5bc7386211318ac7c7652d0bdd53f048f0598b1..eb701be033610992ed0a1a845e4ab9098dbd53bd#radicle-node/src/tests.rs:H1L1L9)

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -58,6 +72,101 @@
...
+
+    return "";
+  }
+
+  function updateHash(newHash: string) {
+    if (newHash !== "") {
+      window.location.hash = newHash;
+    } else {
+      // eslint-disable-next-line no-restricted-globals
+      history.pushState(
```
xphoniex, line 162: `npm run check` forced me to add `// eslint...` otherwise:

```
error  Unexpected use of 'history'  no-restricted-globals
```

--

had to `history` here because, if we just set the hash to `""`, it'll still trigger a scroll to `link#`, which is the top of the page, both annoying and undesired. with `history` we can remove the hash, with forcing a reload of the page.

src/views/projects/SourceBrowser/Changeset.svelte
geigerzaehler: You can always define a variable `isSelected = file.path == selectedFilename`.

src/views/projects/SourceBrowser/Changeset.svelte
```diff
@@ -39,6 +40,27 @@
...
+      [selectedFilename, selectedHunkLine] = fragment.split(":");
+    }
+  }
+  onMount(async () => {
+    window.addEventListener("hashchange", updateSelectedLine);
+  });
+
+  onDestroy(() => {
+    window.removeEventListener("hashchange", updateSelectedLine);
+  });
```
geigerzaehler, line 63: Is there a reason why we can’t move this code to the `FileDiff` component? For one, we should try to minimize the API surface (i.e. the properties) for components. And we should also try to keep code that implements a single feature close together.

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -193,19 +315,56 @@
...
+                        file.path + "::H" + hunkIdx + "L" + lineIdx,
+                      );
+                      if (target) {
+                        target.scrollIntoView();
+                      }
+                    }, 200),
+                    "")}
+                  {/if}
+                {:else}
+                  {((initialScroll = true), "")}
```
geigerzaehler, line 340: I don’t see a reason why this code cannot be moved to the `<script>` part of the component. Embedding imperative code in the declarative template makes it very hard to understand what is going on, especially in this case where the codes sits inside four `if` statements and a `each` iteration.

src/views/projects/SourceBrowser/FileDiff.svelte
geigerzaehler: You can use `window.history`. I’m curious why you need this code

> had to use `history` here because, if we just set the hash to `""`, it'll still trigger a scroll to `<link>#`,

Can you explain which actions the user needs to take to trigger this  code path? I wasn’t able to do it because `newHash` is always set when I try this.

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -58,6 +72,101 @@
...
+    if (match) {
+      // make sure we're on the same hunk
+      if (match[1] === hunkIdx.toString()) {
+        if (lineIdx >= parseInt(match[2])) {
+          return path + ":H" + hunkIdx + "L" + match[2] + "L" + lineIdx;
+        } else {
+          return path + ":H" + hunkIdx + "L" + lineIdx + "L" + match[2];
+        }
+      }
+    }
```
geigerzaehler, line 152: There are three places where we parse a hunk selector. There should be only one place  where parsing happens and it should happen as early as possible.

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -58,6 +72,101 @@
...
+
+    // range selection
+    const match = selectedHunkLine.match(/H(\d*)L(\d*)L?(\d*)/);
+    if (match) {
+      // make sure we're on the same hunk
+      if (match[1] === hunkIdx.toString()) {
+        if (lineIdx >= parseInt(match[2])) {
+          return path + ":H" + hunkIdx + "L" + match[2] + "L" + lineIdx;
+        } else {
+          return path + ":H" + hunkIdx + "L" + lineIdx + "L" + match[2];
```
geigerzaehler, line 149: There are also multiple places where we build the fragment identifier from a selection. Again, there should be a one place, a function, that does this.

Taking this together with the comment above, we should probably deal with a structured representation of the selection and introduce the following type:

```typescript
interface Selection {
  path: string
  hunkIdx: number
  fromLine: number
  toLine: number | undefined
}
```

geigerzaehler (commented):
I’ve made a couple of suggestions to simplify the implementation and make it more robust and understandable.

I’ve also noticed that it is not possible to make a selection across hunks. This is quite irritating for me because there is no strong visual delineation between hunks. Is that something you want to address?

src/views/projects/SourceBrowser/Changeset.svelte
xphoniex: it's because of the event listener, here we only add one on `Changeset`, but if we mount them on `FileDiff`, it'll be linear and depend on number of files involved in diff.

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: If you start a selection and then cross hunk boundary, on click it'd return `""` and hit the code path.

xphoniex:
> I’ve also noticed that it is not possible to make a selection across hunks. This is quite irritating for me because there is no strong visual delineation between hunks. Is that something you want to address?

Correct, right now we don't support cross hunk selection.

I initially thought that would not be necessary since most of the time, we'll be linking to a highlight/selection that was changed, so most of the time that'd fall into a hunk.

Is the issue more the visual part or the usage part, e.g. you'd like to link big chunks?

geigerzaehler:
> Is the issue more the visual part or the usage part, e.g. you'd like to link big chunks?

It’s more the usage part. Sometimes you want to point to a range that spans multiple related hunks.

src/views/projects/SourceBrowser/Changeset.svelte
geigerzaehler: > it's because of the event listener, here we only add one on `Changeset`, but if we mount them on `FileDiff`, it'll be linear and depend on number of files involved in diff.

I see. But I wouldn’t be to concerned about the performance of adding event listeners. Browsers are pretty good at handling a ton of listeners. For example your register two click listeners for every line your render and this is a lot more than  one for every file.

So splitting the code between the components has almost no effect on performance but does make the code a lot more complex and brittle.

src/views/projects/SourceBrowser/FileDiff.svelte
geigerzaehler: Ah, yes.

I’d suggest the following adjustment to retain the state and also pass an empty string as the second argument. This argument is ignored by browsers and passing `document.title` is a legacy thing.

```typescript
      window.history.replaceState(
        window.history.state,
        "",
        window.location.pathname + window.location.search,
      );
```

xphoniex:
commit https://github.com/radicle-dev/radicle-interface/pull/942/commits/f2b80aa654c530b9ce68543904a38bcfa134341b adds support for selecting across hunks, sample:

Screenshot from 2023-07-17 17-24-31

let me know what you think.

src/views/projects/SourceBrowser/Changeset.svelte
xphoniex: updated in https://github.com/radicle-dev/radicle-interface/pull/942/commits/1f80f35a0967e226ffea940f640d819e35da456e

xphoniex:
can do something like:

Screenshot from 2023-07-17 18-19-26

sebastinez:
Really nice @xphoniex, kudos!

Found a minor nit, on selecting multiple files but not persisting.
Maybe we should deselect one file, when selecting a new one?

Bildschirmfoto 2023-07-17 um 17 42 22

xphoniex:
can you add the link @sebastinez so I can investigate? that's a bug and should not be happening

xphoniex:
should be fixed in https://github.com/radicle-dev/radicle-interface/pull/942/commits/7e89a9c830e00c9ce8f0c67eac8cc2463f87d808

xphoniex:
so, right now there's no way to cancel a selection, unless you manually modify url and load again, on GitHub, clicking on the side on the code un-selects the selection, so 

- I tried adding an `on:click` on `<main>` inside **src/views/projects/View.svelte**, but it'd also intercept any click on childs, including on line numbers, so selection would stop working.
- tried adding it to hotkeys, on `Esc`, but neither `window.history.replaceState` nor `.pushState` trigger the `hashchange` event.
- tried doing `window.location.hash = ""` on `Esc`, it works, but scroll all the way to the top of the page.

any ideas?

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -58,6 +114,90 @@
...
+    } else {
+      return `${path}:H${selection.startHunk}L${selection.startLine}H${hunkIdx}L${lineIdx}`;
+    }
+  }
+
+  function updateHash(newHash: string) {
+    if (newHash !== "") {
+      window.location.hash = newHash;
+    } else {
+      window.history.replaceState(
```
sebastinez, line 185: With the implementation of the custom page router, we started to avoid leaking `window.history` into individual components.
Could we maybe move this function to `src/lib/router.ts` or solve this without having `window.history` in this component?

sebastinez (commented):
Just two minor things on architecture.

@rudolfs could we maybe let our designer look over the color scheme for the diff highlighting and the selection of those lines? I think there is room for improvement

xphoniex:
This is the latest color scheme btw (haven't pushed yet):

Screenshot from 2023-07-18 09-52-06

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: I can remove `window.history` if you have a solution for this: https://github.com/radicle-dev/radicle-interface/pull/942#issuecomment-1638536224

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -16,6 +17,48 @@
...
 
   let collapsed = false;
+  let selection: Selection | undefined = undefined;
+
+  onMount(async () => {
+    window.addEventListener("hashchange", updateSelection);
+  });
+
+  onDestroy(() => {
+    window.removeEventListener("hashchange", updateSelection);
```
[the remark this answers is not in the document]
xphoniex: not sure I understand, you want me to use only `onMount`? I only borrowed this piece of code from **src/views/projects/Blob.svelte**.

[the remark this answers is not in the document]
hecatia-elegua: @sebastinez no, since the onMount is async

src/views/projects/SourceBrowser/FileDiff.svelte
hecatia-elegua: @xphoniex try the event modifier self: `on:click|self`

xphoniex: @hecatia-elegua that's not the only issue, `window.history.replaceState | .pushState` doesn't cause `hashchange` to be emitted, so while the url changes, selection is still visible and `FileDiff` doesn't update.

src/views/projects/SourceBrowser/FileDiff.svelte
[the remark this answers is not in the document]
sebastinez: @hecatia-elegua you are right with the async part, but I don't think we need the `onMount` function to be async, since we don't await anything.

@xphoniex I leave it to you if you want to return from `onMount` or use the additional `onDestroy`

src/views/projects/SourceBrowser/FileDiff.svelte
hecatia-elegua: Maybe you can `updateSelection` after `replaceState, or `updateSelection` in the `on:click|self` as well, I'm not really sure. I've tried around a bunch.

Also you probably don't want `<main>`, since it would only trigger at the screen edges, but probably the .metadata `<div>` or some other way to do this

xphoniex: works, added in https://github.com/radicle-dev/radicle-interface/pull/942/commits/d1c5916da5581f26fe02e54c67ae17b80fe7ab4b

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -120,11 +278,37 @@
...
+  }
+  .diff-line[data-type="@-"] > * {
+    color: var(--color-negative);
+    background-color: var(--color-negative-6);
+  }
+  .diff-line[data-type="@"] {
+    border-left: 4px solid gold;
+  }
+  .diff-line[data-type="@+"] {
+    border-left: 4px solid gold;
```
xphoniex, line 297: had to use `gold` color, cause `--color-caution` in not what I want.

geigerzaehler:
> so, right now there's no way to cancel a selection, unless you manually modify url and load again, on GitHub, clicking on the side on the code un-selects the selection, so
> any ideas?

You can do something like the following:
```typescript
window.addEventListener("click", e => {
  if (
    !(e.target instanceof HTMLElement && e.target.hasAttribute("data-type"))
  ) {
    updateHash("");
  }
});
```

If you implement this, please change the attribute name to something that is less ambiguous and also make sure that you remove the event listener again on unmounting

src/views/projects/SourceBrowser/FileDiff.svelte
geigerzaehler: > With the implementation of the custom page router, we started to avoid leaking `window.history` into individual components.

I share your concern, @sebastinez. But I do think the hack is fine for the moment.

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -16,6 +17,53 @@
...
+              file.path + "::H" + hunkIdx + "L" + lineIdx,
+            );
+            if (target) {
+              target.scrollIntoView();
+            }
+          }, 200);
+        }
+      }
+    }
+  }
```
geigerzaehler, line 50: You can replace all of this code with the following:

```suggestion
  onMount(() => {
    updateSelection();

    if (selection) {
      document
        .getElementById(
          file.path + "::H" + selection.startHunk + "L" + selection.startLine,
        )
        ?.scrollIntoView();
    }
  });
```

There’s no need for the the timeout because we are running after mounting.

I would also suggest to simplify the ID format and not rely on a subtle difference to the hash format to prevent scrolling. Something like `[file.path, selection.startHunk, selection.startLine].join("-")`

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -193,19 +377,40 @@
...
+                    data-type={lineSign(line)}>
                     {lineNumberL(line)}
                   </td>
-                  <td class="diff-line-number right" data-type={lineSign(line)}>
+                  <td
+                    class="diff-line-number right"
+                    on:click={e => {
+                      updateHash(
+                        calcNewHash(file.path, hunkIdx, lineIdx, e.shiftKey),
+                      );
```
geigerzaehler, line 411: You can merge these to function into one. I’d also suggest passing the whole event instead of `e.shiftKey` and have the function choose how to handle modifier keys.

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -16,6 +17,53 @@
...
+      }
+    }
+  }
+
+  function updateSelection() {
+    const fragment = window.location.hash.substr(1);
+    const match = fragment.match(/(.+):H(\d+)L(\d+)(H(\d+)L(\d+))?/);
+    if (match && match[1] === file.path) {
+      selection = {
+        file: match[1],
```
geigerzaehler, line 57: This property is never used.

geigerzaehler (commented):
I think we’re almost there. It’s working pretty well for me.

geigerzaehler:
> want me to keep both Esc and click?

I personally don’t think handling Esc is necessary but I’d leave it up to you to make that call.

> btw, is the border animation annoying? see this gif:

Yeah, I think this should be fixed :magic_wand:

xphoniex:
> 
> Yeah, I think this should be fixed magic_wand

I thought adding a simple `box-sizing: border-box;` would fix this but it's not working. alternative is to keep border for all lines, and just change the color, but that wouldn't be ideal. any ideal why `box-sizing` is not working?

geigerzaehler:
> I thought adding a simple `box-sizing: border-box;` would fix this but it's not working. alternative is to keep border for all lines, and just change the color, but that wouldn't be ideal. any ideal why `box-sizing` is not working?

I’m not sure why box-sizing is not working but it probably has to do with the fact that this is a table. I guess your best bet is to add a fixed transparent border to the first cell in the table and then color it gold when the line is selected.

xphoniex:
I don't like this at all:

Screenshot from 2023-07-18 17-48-47

The bottom margin was already enough triggering as it was, now we have the left margin too  :smile: 

I tried matching the color, but somehow I can't exactly match it, and don't know why either:

Screenshot from 2023-07-18 18-03-52

maybe best to leave it as-is for now?

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: can do this, worth the squeeze?
- need it to be two functions cause of `return`s, otherwise would be unreadable `else if`s
- also notice `e: any`

```diff
diff --git a/src/views/projects/SourceBrowser/FileDiff.svelte b/src/views/projects/SourceBrowser/FileDiff.svelte
index a133f29..4617dd9 100644
--- a/src/views/projects/SourceBrowser/FileDiff.svelte
+++ b/src/views/projects/SourceBrowser/FileDiff.svelte
@@ -22,7 +22,7 @@
   onMount(() => {
     window.addEventListener("keydown", e => {
       if (e.key === "Escape") {
-        updateHash("");
+        _updateHash("");
       }
     });
     window.addEventListener("hashchange", updateSelection);
@@ -152,36 +152,39 @@
     );
   }

-  function calcNewHash(
-    path: string,
-    hunkIdx: number,
-    lineIdx: number,
-    shift: boolean,
-  ): string {
+  function updateHash(path: string, hunkIdx: number, lineIdx: number, e: any) {
     // single line selection
-    if (!shift) {
-      return path + ":H" + hunkIdx + "L" + lineIdx;
+    if (!e.shiftKey) {
+      return _updateHash(path + ":H" + hunkIdx + "L" + lineIdx);
     }

     if (!selection) {
-      return "";
+      return _updateHash("");
     }

     // range selection
     if (hunkIdx === selection.startHunk) {
       if (lineIdx >= selection.startLine) {
-        return `${path}:H${hunkIdx}L${selection.startLine}H${hunkIdx}L${lineIdx}`;
+        return _updateHash(
+          `${path}:H${hunkIdx}L${selection.startLine}H${hunkIdx}L${lineIdx}`,
+        );
       } else {
-        return `${path}:H${hunkIdx}L${lineIdx}H${hunkIdx}L${selection.startLine}`;
+        return _updateHash(
+          `${path}:H${hunkIdx}L${lineIdx}H${hunkIdx}L${selection.startLine}`,
+        );
       }
     } else if (hunkIdx < selection.startHunk) {
-      return `${path}:H${hunkIdx}L${lineIdx}H${selection.startHunk}L${selection.startLine}`;
+      return _updateHash(
+        `${path}:H${hunkIdx}L${lineIdx}H${selection.startHunk}L${selection.startLine}`,
+      );
     } else {
-      return `${path}:H${selection.startHunk}L${selection.startLine}H${hunkIdx}L${lineIdx}`;
+      return _updateHash(
+        `${path}:H${selection.startHunk}L${selection.startLine}H${hunkIdx}L${lineIdx}`,
+      );
     }
   }

-  function updateHash(newHash: string) {
+  function _updateHash(newHash: string) {
     if (newHash !== "") {
       window.location.hash = newHash;
     } else {
@@ -396,9 +399,7 @@
                     id={file.path + "::H" + hunkIdx + "L" + lineIdx}
                     class="diff-line-number left"
                     on:click={e => {
-                      updateHash(
-                        calcNewHash(file.path, hunkIdx, lineIdx, e.shiftKey),
-                      );
+                      updateHash(file.path, hunkIdx, lineIdx, e);
                     }}
                     data-type={lineSign(line)}>
                     {lineNumberL(line)}
@@ -406,9 +407,7 @@
                   <td
                     class="diff-line-number right"
                     on:click={e => {
-                      updateHash(
-                        calcNewHash(file.path, hunkIdx, lineIdx, e.shiftKey),
-                      );
+                      updateHash(file.path, hunkIdx, lineIdx, e);
                     }}
                     data-type={lineSign(line)}>
                     {lineNumberR(line)}
```

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: https://github.com/radicle-dev/radicle-interface/pull/942/commits/8fcea773f93446c084dc40273d8db4f6f0c7bcc5

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: https://github.com/radicle-dev/radicle-interface/pull/942/commits/1b19c1999d0a7f99a2cf91bbef7a076d7be2594e

xphoniex:
@sebastinez I opted for click instead of Escape key, and had to re-introduce old `onMount` and `onDestroy` in commit https://github.com/radicle-dev/radicle-interface/pull/942/commits/4f2245d4ac5d755ae1dd1e67f97c7e98c5ade98c due to this comment:

> If you implement this, please change the attribute name to something that is less ambiguous and also make sure that you remove the event listener again on unmounting

please let me know if this works.

src/views/projects/SourceBrowser/FileDiff.svelte
geigerzaehler: > * need it to be two functions cause of `return`s, otherwise would be unreadable `else if`s

Yes, you’re right. But no need to overthink this. How about this?

```diff
diff --git a/src/views/projects/SourceBrowser/FileDiff.svelte b/src/views/projects/SourceBrowser/FileDiff.svelte
index eba16a8..24f8daa 100644
--- a/src/views/projects/SourceBrowser/FileDiff.svelte
+++ b/src/views/projects/SourceBrowser/FileDiff.svelte
@@ -144,14 +144,14 @@
     }
   }
 
-  function calcNewHash(
-    path: string,
+  function hashFromSelection(
     hunkIdx: number,
     lineIdx: number,
-    shift: boolean,
+    event: MouseEvent,
   ): string {
+    const path = file.path;
     // single line selection
-    if (!shift) {
+    if (!event.shiftKey) {
       return path + ":H" + hunkIdx + "L" + lineIdx;
     }
 
@@ -173,6 +173,10 @@
     }
   }
 
+  function selectLine(hunkIdx: number, lineIdx: number, event: MouseEvent) {
+    updateHash(hashFromSelection(hunkIdx, lineIdx, event));
+  }
+
   function updateHash(newHash: string) {
     if (newHash !== "") {
       window.location.hash = newHash;
@@ -392,21 +396,13 @@
                   <td
                     id={[file.path, "H" + hunkIdx, "L" + lineIdx].join("-")}
                     class="diff-line-number left"
-                    on:click={e => {
-                      updateHash(
-                        calcNewHash(file.path, hunkIdx, lineIdx, e.shiftKey),
-                      );
-                    }}
+                    on:click={e => selectLine(hunkIdx, lineIdx, e)}
                     data-line-type={lineSign(line)}>
                     {lineNumberL(line)}
                   </td>
                   <td
                     class="diff-line-number right"
-                    on:click={e => {
-                      updateHash(
-                        calcNewHash(file.path, hunkIdx, lineIdx, e.shiftKey),
-                      );
-                    }}
+                    on:click={e => selectLine(hunkIdx, lineIdx, e)}
                     data-line-type={lineSign(line)}>
                     {lineNumberR(line)}
                   </td>
```

Not a huge deal though. Feel free to leave it like it is.

geigerzaehler:
> maybe best to leave it as-is for now? css rules are getting more complex than necessary.

I think that’s a good idea. We can fix this later.

geigerzaehler:
> If you implement this, please change the attribute name to something that is less ambiguous and also make sure that you remove the event listener again on unmounting

Once you address that this PR should be ready to go :checkered_flag:

sebastinez:
> @sebastinez I opted for click instead of Escape key, and had to re-introduce old `onMount` and `onDestroy` in commit [4f2245d](https://github.com/radicle-dev/radicle-interface/commit/4f2245d4ac5d755ae1dd1e67f97c7e98c5ade98c) due to this comment:
> 
> > If you implement this, please change the attribute name to something that is less ambiguous and also make sure that you remove the event listener again on unmounting
> 
> please let me know if this works.

Works for me.
I don't think you had to re-introduce `onDestroy` for removing the `hashChange` event listener, you could've just returned the `window.removeEventListener` function from the `onMount function.
More importantly, I think you should remove the `click` eventListener in the `onDestroy` hook

rudolfs:
> Just two minor things on architecture.
> 
> @rudolfs could we maybe let our designer look over the color scheme for the diff highlighting and the selection of those lines? I think there is room for improvement

I'd try to get this PR in a "good enough" state to merge and then iterate on it later with polish. Daniel's currently focused on the overall color system and design, I don't want to take away his attention from that right now.

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: https://github.com/radicle-dev/radicle-interface/pull/942/commits/55a71d1c651b38c6715fd67b97e23b5da97d6d6a

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -48,7 +48,8 @@
     if (
       !(
         e.target instanceof HTMLElement &&
-        e.target.hasAttribute("data-line-type")
+        (e.target.hasAttribute("data-line-type") ||
+          e.target.classList.contains("diff-line-content"))
```
xphoniex, line 52: clicking on anything on the page, causes the highlighted part to be de-selected, so I make an exception for `diff-line-content` class here.

it's still not 100% accurate, and I guess it has do to with tables cell, but it's good enough.

xphoniex:
I also like to add a visual test for this, since it might break once we add expander logic, yay or nay?

if yay, should I just create another file `tests/visual/diffSelection.spec.ts` with:

```ts
test("diff selection page", async ({ page }) => {
  await page.addInitScript(() => {
    window.initializeTestStubs = () => {
      window.e2eTestStubs.FakeTimers.install({
        now: new Date("November 24 2022 12:00:00").valueOf(),
        shouldClearNativeTimers: true,
        shouldAdvanceTime: false,
      });
    };
  });

  await page.goto("seeds/radicle.local/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/e6d27cade05288b5bfa79f0e3c47bb07fdc3c2a0?diff=1ddfa4fe94adca5096a82316fbaf7b147e3b5d9a..5242a14b5e9fefd07db040940223b7f37ba72988#radicle-cli/src/commands/issue.rs:H1L116H2L7", { waitUntil: "networkidle" });
});
```

rudolfs:
> I also like to add a visual test for this, since it might break once we add expander logic, yay or nay?
> 
> if yay, should I just create another file `tests/visual/diffSelection.spec.ts` with:
> 
> ```ts
> test("diff selection page", async ({ page }) => {
>   await page.addInitScript(() => {
>     window.initializeTestStubs = () => {
>       window.e2eTestStubs.FakeTimers.install({
>         now: new Date("November 24 2022 12:00:00").valueOf(),
>         shouldClearNativeTimers: true,
>         shouldAdvanceTime: false,
>       });
>     };
>   });
> 
>   await page.goto("seeds/radicle.local/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/e6d27cade05288b5bfa79f0e3c47bb07fdc3c2a0?diff=1ddfa4fe94adca5096a82316fbaf7b147e3b5d9a..5242a14b5e9fefd07db040940223b7f37ba72988#radicle-cli/src/commands/issue.rs:H1L116H2L7", { waitUntil: "networkidle" });
>   await expect(page).toHaveScreenshot();
> });
> ```

Sounds good. I'd also suggest adding a couple of playwright tests covering:
  - selection of hunks
  - URL changes on selection
  - deselection

src/views/projects/SourceBrowser/FileDiff.svelte
rudolfs: We should use the primary color for now to avoid introducing colors that aren't part of the theme. We use the primary color for text selection at the moment anyway.


Screenshot 2023-07-19 at 16 45 34



```diff
diff --git a/src/views/projects/SourceBrowser/FileDiff.svelte b/src/views/projects/SourceBrowser/FileDiff.svelte
index 6aca391..ab40256 100644
--- a/src/views/projects/SourceBrowser/FileDiff.svelte
+++ b/src/views/projects/SourceBrowser/FileDiff.svelte
@@ -293,16 +293,16 @@
     background-color: var(--color-negative-6);
   }
   .diff-line[data-line-type="@"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
   }
   .diff-line[data-line-type="@+"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
   }
   .diff-line[data-line-type="@-"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
   }
   .diff-line[data-line-type="@@@"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
     background-color: var(--color-foreground-4);
   }
   .diff-line-number {
```

xphoniex:
> Sounds good. I'd also suggest adding a couple of playwright tests covering:
> 
> * selection of hunks
> * URL changes on selection
> * deselection

https://github.com/radicle-dev/radicle-interface/pull/942/commits/a5f69b011f779535e61baf514f51ff1b385faa06 should cover 2nd & 3rd.

for hunks, our README.md is too small and has no hunks atm:

Screenshot from 2023-07-19 19-56-10

tried adding another md file in initial patch but was getting `git` error:
```
Creating cobs fixture
Error: Command failed with exit code 1: git switch main
error: Your local changes to the following files would be overwritten by checkout:
	BIG_README.md
	README.md
Please commit your changes or stash them before you switch branches.
Aborting
```

src/views/projects/SourceBrowser/FileDiff.svelte
rudolfs: Actually, let's do this, it's a bit more in line with the current overall design:

```diff
diff --git a/src/views/projects/SourceBrowser/FileDiff.svelte b/src/views/projects/SourceBrowser/FileDiff.svelte
index 6aca391..8e5fcae 100644
--- a/src/views/projects/SourceBrowser/FileDiff.svelte
+++ b/src/views/projects/SourceBrowser/FileDiff.svelte
@@ -281,28 +281,28 @@
     background-color: var(--color-negative-2);
   }
   .diff-line[data-line-type="@"] > * {
-    color: var(--color-foreground-5);
-    background-color: var(--color-foreground);
+    color: var(--color-foreground-6);
+    background-color: var(--color-foreground-4);
   }
   .diff-line[data-line-type="@+"] > * {
-    color: var(--color-positive);
-    background-color: var(--color-positive-6);
+    color: var(--color-positive-6);
+    background-color: var(--color-positive-4);
   }
   .diff-line[data-line-type="@-"] > * {
-    color: var(--color-negative);
-    background-color: var(--color-negative-6);
+    color: var(--color-negative-6);
+    background-color: var(--color-negative-4);
   }
   .diff-line[data-line-type="@"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
   }
   .diff-line[data-line-type="@+"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
   }
   .diff-line[data-line-type="@-"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
   }
   .diff-line[data-line-type="@@@"] {
-    border-left: 4px solid gold;
+    border-left: 4px solid var(--color-primary);
     background-color: var(--color-foreground-4);
   }
   .diff-line-number {
```

Screenshot 2023-07-20 at 10 06 39
Screenshot 2023-07-20 at 10 07 03

src/views/projects/SourceBrowser/FileDiff.svelte
```diff
@@ -16,6 +17,59 @@
...
+    window.addEventListener("click", deselectHandler);
+    window.addEventListener("hashchange", updateSelection);
+  });
+
+  onDestroy(() => {
+    window.removeEventListener("click", deselectHandler);
+    window.removeEventListener("hashchange", updateSelection);
+  });
+
+  onMount(() => {
```
rudolfs, line 32: Can we combine these into one `onMount`?

xphoniex: we can, but I find separate `onMount` and `onDestroy` more readable, your call.

src/views/projects/SourceBrowser/FileDiff.svelte
xphoniex: https://github.com/radicle-dev/radicle-interface/pull/942/commits/a9d75a3b1fa093f215d536d106aa863f0bacb883

src/views/projects/SourceBrowser/FileDiff.svelte
rudolfs: Yeah, let's do one `onMount` to be consistent with the rest of the codebase.

rudolfs: I don't follow, why would we want to change something there?

xphoniex: gotcha, https://github.com/radicle-dev/radicle-interface/pull/942/commits/5ad8e2ef11a983ed3d22bd9e23c5093ed8f6820a

rudolfs (approved):
🚀 ✨

PR #945, opened 2023-07-17
Replace marked namespaces with Marked instance

### LINKED ISSUE #919: Address comments on `Renderer`
Add the extensions and deprecated options we define inside the `Renderer`.

Also since we are touching marked, we could maybe also look into using a `Marked` instance instead of having different `marked` imports, and maybe have a getter for that instance, on which we call `parse()`?

```
/src/components/InlineMarkdown.svelte
  11,24:     dompurify.sanitize(marked.parseInline(content));

/src/components/Markdown.svelte
  55,7:       marked.parse(content, {

/src/lib/markdown.ts
  96,10:     }. ${marked.parseInline(
```

See: https://github.com/markedjs/marked/pull/2831

sebastinez:
Based on a new addition to marked, this commits replaces the imported marked namespaces with a Marked class, that doesn't change when `marked.use` introduces new options or extensions.

Also changes the naming slighty to use `markdown` as module, so it sounds more generic.

Reference: https://github.com/markedjs/marked/pull/2831

Closes #919

### DIFF
--- a/src/components/Markdown.svelte
+++ b/src/components/Markdown.svelte
@@ -2,13 +2,13 @@
   import dompurify from "dompurify";
   import matter from "@radicle/gray-matter";
   import { afterUpdate } from "svelte";
-  import { marked } from "marked";
   import { toDom } from "hast-util-to-dom";
 
   import * as router from "@app/lib/router";
+  import markdown from "@app/lib/markdown";
+  import { Renderer } from "@app/lib/markdown";
   import { highlight } from "@app/lib/syntax";
   import { isUrl, twemoji, scrollIntoView, canonicalize } from "@app/lib/utils";
-  import { Renderer } from "@app/lib/markdown";
 
   export let content: string;
   // If present, resolve all relative links with respect to this URL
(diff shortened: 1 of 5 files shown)


src/components/Markdown.svelte
```diff
@@ -52,14 +52,10 @@
...
-        renderer: new Renderer(linkBaseUrl),
-        // TODO: Disables deprecated options, remove once removed from marked
-        mangle: false,
-        headerIds: false,
-      }),
-    );
+    markdown.setOptions({
+      renderer: new Renderer(linkBaseUrl),
+    });
+    return dompurify.sanitize(markdown.parse(content));
```
geigerzaehler, line 58: I believe this modifies the options of the shared `Markdown` instance. This may mess with other places where we use `markdown`. I would suggest the following:

```suggestion
    return dompurify.sanitize(markdown.parse(content, {
      renderer: new Renderer(linkBaseUrl)
    }));
```

sebastinez: Nice, yeah that seems better 👍

sebastinez: Added

geigerzaehler (approved):

PR #956, opened 2023-07-19
Update dependencies

sebastinez:
Gets prettier to v3, which starts to write doctype in lowercase

See https://prettier.io/blog/2023/07/05/3.0.0.html#print-html5-doctype-in-lowercase-7391httpsgithubcomprettierprettierpull7391-by-fiskerhttpsgithubcomfisker


rudolfs (approved):
✨

PR #957, opened 2023-07-20
Fix movement of diff text when lines are selected

geigerzaehler:


geigerzaehler:
> one thing I'd change is the [gray](https://user-images.githubusercontent.com/158411/254817333-74df56cc-e1a1-4c8f-8189-1c14d21b2694.png) background, can we have a lighter color @rudolfs ? also the selected bg color for header is the same as context lines, can we revert that?

Since I haven’t changed any of the colors in this PR I’d prefer it if somebody else tackles it in a separate PR, maybe with some input from a designer.

xphoniex:
hey, this is not directly related to this PR but this [line](https://github.com/radicle-dev/radicle-interface/commit/c44cd5771bfc4698533602b6b470f978a4c21707#diff-86922c6cf991e9927a1a37a653efc6e9c1db5dfacb4a99a772798d30d4b0f3dbR260) is causing the color to be the same for context and header. you can fix it by using:

```css
.diff-line.selected.type-context > * 
```

instead.

this is my last comment, so cheerio boys!

Issue #938, opened 2023-07-12
Behavior of peer and branch selector in single commit view


geigerzaehler:
At the moment, when you navigate to the single commit view, we show the peer selector with no peer selected the commit ID in place of the branch selector.

The current behavior raises a couple of UX questions and issues

* What should happen when a peer is selected from the peer selector? (Right now nothing happens, except for a URL change)
* Should we show a branch selector but with no branch selected? What should happen when a branch is selected?
* Should we show the peer selector at all?

![image](https://github.com/radicle-dev/radicle-interface/assets/3919579/1bbfbc76-8f29-4397-831d-91c70e160dbe)

rudolfs:
Yeah, I think hiding the peer selector makes the most sense.

teramfo:
In response to the second question, it's not necessary to have a branch selector as each commit can only belong to one branch. Still, it would be useful to display the branch that the commit belongs to, and this is consistent with what we do on other pages.

Something like this:
![image](https://github.com/radicle-dev/radicle-interface/assets/139389093/02c1d1ce-da10-4b81-8aeb-419a018255b8)
 
 @cloudhead to review this comment.

rudolfs:
> In response to the second question, it's not necessary to have a branch selector as each commit can only belong to one branch. Still, it would be useful to display the branch that the commit belongs to, and this is consistent with what we do on other pages.

Actually any commit can belong to any number of branches. So in this case it might be useful to show in which branches the commit is present, but it wouldn't make sense to switch between these branches, because the commit is always the same. So we might want to show it in a different way than a dropdown selector to distinguish it from the existing branch selector.

PR #958, opened 2023-07-20
Remove source browsing header from commit view

geigerzaehler:
For a single commit view the source browsing header (peer selector, branch selector, commits, contributor count) does not contain meaningful information. We remove the header and simply show the commit.

See #938.

Before:
image

After:
image


rudolfs (approved):
🧹

PR #976, opened 2023-07-26
Fix navigating to history when a file is selected

rudolfs:
This fixes not being able to navigate to the commit history while a file is selected on the default peer/branch.


sebastinez (approved):
Nice catch, works as advertised 👍

PR #994, opened 2023-08-08
Extract project menu into a layout

rudolfs:
This also flattens the `ProjectLoadedRoute` type and moves all project components to `App.svelte`.


geigerzaehler (approved):
Amazing work :muscle:! It looks so much cleaner now.

Issue #894, opened 2023-07-03
`Back` button not working


cloudhead:
https://app.radicle.xyz/seeds/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/cb61560b6f7a2ab39be13a45899445ce0b921f59

Try clicking on `Back`. Probably because it does `history.back()` and if you don't have history it won't work. We should change this to go to the patch listing if there's no history maybe. We should look at this globally though, not just this page, but all error pages like this.

Issue #986, opened 2023-08-01
Routing error handling


geigerzaehler:
We want to implement handling of errors that are thrown while loading a view in the following stages:

- [ ] Catch all errors thrown while loading a route and show a generic error view with error message and stacktrace ([design](https://www.figma.com/file/zKSJOOeO6qg4GwaFSM8poL/error-UX?type=design&node-id=2-266))
- [x] Add error handling “not found” with views that don’t require us to show navigation context. The errors are: seed not found, project, issue, patch, commit not found. ([design](https://www.figma.com/file/zKSJOOeO6qg4GwaFSM8poL/error-UX?type=design&node-id=1-78))
  - [x] adjust backend to distinguish not found (404) errors from actual server errors (500)
- [x] Add error handling for “file not found” in tree browser ([design](https://www.figma.com/file/zKSJOOeO6qg4GwaFSM8poL/error-UX?type=design&node-id=0-1))

rudolfs:
Moved to Radicle

https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z4V1sjrXqjvFdnCUbxPFqd5p4DtH5/issues/c1562cad02138454ee7a64203ab79561dbc13c40

PR #996, opened 2023-08-09
Streamline error handling

rudolfs:
Streamline our 404 not found pages by showing the name of the resource that wasn't found, e.g. `Page`, `Issue`, `Patch`, `Commit` and `Seed`.

We remove the `Back` button which simplifies the router logic, because we don't have to keep track of the browser history manually. Users can use the browser back button.

Finally we add tests for the 404 and loading error pages.

In a follow-up we'll implement a global unhandled exception page, which could possibly replace the `loadError` route.

Refs: https://github.com/radicle-dev/radicle-interface/issues/986.
Closes: https://github.com/radicle-dev/radicle-interface/issues/894.


tests/support/peerManager.ts
```diff
@@ -287,24 +293,23 @@ export class RadiclePeer {
...
-                const result: NodeEvent = JSON.parse(data);
-                this.#eventRecords.push(result);
-              } catch (e) {
-                console.log("Error parsing event", data);
-              }
-            }
-          });
-      },
-    );
+    const stdout = this.rad(["node", "events"], { cwd: this.#radHome }).stdout;
```
sebastinez, line 296: Really nice improvement! thanks 🙇

sebastinez (approved):
Okay tested the code paths and agree with the simplicity of the NotFound view 👍

tests/visual/landingPage.spec.ts
```diff
@@ -19,3 +19,14 @@ test("landing page", async ({ page }) => {
   await page.goto("/", { waitUntil: "networkidle" });
   await expect(page).toHaveScreenshot();
 });
+
+test("load error", async ({ page }) => {
```
sebastinez, line 23: I think we should add the FakeTimers to this test, else the activity is going to be shifting every week or so.

sebastinez (commented):
Just some minor things I saw

tests/visual/landingPage.spec.ts
rudolfs: Good catch. Fixed.

sebastinez (approved):
Nice, LGTM

PR #1001, opened 2023-08-16
Fix project sorting on node page when there is no activity data

rudolfs:
TBD: add a testcase.


sebastinez (approved):
Does what it says it does. 👍

PR #998, opened 2023-08-14
Enable adding and removing reactions

sebastinez:
Add the reaction selector back and be able to react to comments

Missing:

- [x] Adding tests
- [x] [Patch to add ability to remove reactions on `radicle-httpd`](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/6aab1d036354378edfaaaa9347b5f40d2c852302)
- [x] Get UI to an acceptable state
- [ ] Remove a reaction by clicking on it in the reaction selector
- [ ] Add tests for multiple peers reaction on the same comment..

PR #1008, opened 2023-08-22
Enable adding and removing reactions

sebastinez:
Add the reaction selector back and be able to react to comments

Missing:

- [x] Adding tests
- [x] [Patch to add ability to remove reactions on `radicle-httpd`](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/6aab1d036354378edfaaaa9347b5f40d2c852302)
- [x] Get UI to an acceptable state
- [x] Remove a reaction by clicking on it in the reaction selector
- [x] Add tests for multiple peers reaction on the same comment..~
  - Did test it manually and worked, will open an issue to add multiple peers doing changes simultaneously.

Successor of #998


rudolfs:
Screenshot 2023-08-23 at 09 11 36

Can we make all reactions look the same? I prefer the look of the top (issue) reactions which are slightly smaller. The text is also smaller and blue.

src/components/Comment.svelte
```diff
@@ -99,17 +124,39 @@
     {:else}
       <Markdown {rawPath} content={body} />
     {/if}
-    {#if groupedReactions.size > 0}
-      <div class="reactions">
-        {#each groupedReactions as [reaction, nids], key}
-          <Chip {key}>
-            <div class="reaction">
+    <div class="reactions">
+      {#if $httpdStore.state === "authenticated"}
```
rudolfs, line 128: Can we move this check one element above so it wraps the reaction container?

rudolfs:
When not logged in, can we show the reactions just read-only? Like we do for the issue reactions:

Screenshot 2023-08-23 at 09 39 16

src/components/Comment.svelte
sebastinez: I think we should keep it like this, so we can show the reactions to a comment, even if a user isn't authenticated.

sebastinez:
> Screenshot 2023-08-23 at 09 11 36
> Can we make all reactions look the same? I prefer the look of the top (issue) reactions which are slightly smaller. The text is also smaller and blue. The reactions could also benefit from a `user-select: none` css attribute, so that clicking them doesn't select any text.

Yeah no problem, done

sebastinez:
> When not logged in, can we show the reactions just read-only? Like we do for the issue reactions:
> 
> Screenshot 2023-08-23 at 09 39 16

Yeah works for me

rudolfs:
There's a weird margin now showing up in the visual tests:

Screenshot 2023-08-23 at 14 57 52

sebastinez:
Visual changes, still show the issue icon change

sebastinez:
> There's a weird margin now showing up in the visual tests:
> 
> Screenshot 2023-08-23 at 14 57 52

Is fixed with [3fdd989](https://github.com/radicle-dev/radicle-interface/pull/1008/commits/3fdd98994d1ae99dc3ac413aa19c686084668568)

rudolfs:
There seems to be a margin missing between the reactions: 
Screenshot 2023-08-23 at 15 39 30

rudolfs (approved):

PR #1014, opened 2023-08-24
Add patch comment and state modifiers

sebastinez:
This PR adds the ability to add root level comments to a patch revision
And to change the state of a patch from draft to open and open to archived.


rudolfs:
The patch numbers don't update when I change the state.

sebastinez:
> The patch numbers don't update when I change the state.

Ok this has been fixed I adapted the way patches update to the way we implemented it in issues.

rudolfs (approved):
Could we also add a test for changing patch states?

PR #1015, opened 2023-08-24
Remove apple-touch-icon from index.html

rudolfs:
Firefox uses the `apple-touch-icon` as favicon for some inexplicable reason. This icon has a black background and is only intended for apple home screen. The black background is important as otherwise when you minimise the app in iOS it briefly flickers to transparent and then black. By having a black background we avoid this flickering on iOS.

[The solution](https://github.com/go-gitea/gitea/issues/10394#issuecomment-589690317) seems to be removing `apple-touch-icon` from the index file and renaming the file itself, so that apple devices find it.

Closes: https://github.com/radicle-dev/radicle-interface/issues/783.

Screenshot 2023-08-24 at 18 12 13

It's also very tricky to clear the favicon cache of Firefox, I had to delete: `rm Library/Application\ Support/Firefox/Profiles/1q9ptpg1.default-release-1692887325961/favicons.sqlite`.


sebastinez (approved):
One day platforms and browers will unite and do these things the same way.. 😅

PR #1016, opened 2023-08-25
Refactor some of the reaction components

sebastinez:
To adapt to different form factors, this PR refactors some of the reactions elements.

Some images:
Bildschirmfoto 2023-08-25 um 15 22 44

Bildschirmfoto 2023-08-25 um 14 36 36

### DIFF
--- a/src/components/Comment.svelte
+++ b/src/components/Comment.svelte
@@ -5,9 +5,10 @@
   import { httpdStore } from "@app/lib/httpd";
 
   import Authorship from "@app/components/Authorship.svelte";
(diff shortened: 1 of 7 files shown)


rudolfs:
Wonder if we should flip around the hover states? Normal: gray, hover: white.

sebastinez:
> Wonder if we should flip around the hover states? Normal: gray, hover: white.

Yeah thought so too at one point, but the gray on a clickable icon seems to me too disabled.

rudolfs (approved):

PR #1019, opened 2023-08-26
Try clipboard

dnlklmn:
trying out colors/behavior for the copy to clipboard field


rudolfs:
Pulled your commits into my PR

PR #1021, opened 2023-08-28
Improve layout by moving its concerns into a component

rudolfs:


rudolfs:
Moved this into https://github.com/radicle-dev/radicle-interface/pull/1018

PR #1022, opened 2023-08-28
Add local projects to landing page, if available

sebastinez:
This PR shows either the projects found on a local node to the landing page, or fallback to the pinned ones in `config.projects.pinned`

### DIFF
--- a/src/App.svelte
+++ b/src/App.svelte
@@ -28,8 +28,7 @@
 
   const activeRouteStore = router.activeRouteStore;
 
-  void router.loadFromLocation();
-  httpd.initialize();
+  void httpd.initialize().finally(() => void router.loadFromLocation());
 
   if (!window.VITEST && !window.PLAYWRIGHT && import.meta.env.PROD) {
     const plausible = Plausible({
--- a/src/views/home/Index.svelte
+++ b/src/views/home/Index.svelte
@@ -1,12 +1,15 @@
 <script lang="ts">
   import type { ProjectBaseUrlActivity } from "./router";
 
+  import { api } from "@app/lib/httpd";
   import { twemoji } from "@app/lib/utils";
 
   import Link from "@app/components/Link.svelte";
   import ProjectCard from "@app/components/ProjectCard.svelte";
 
   export let projects: ProjectBaseUrlActivity[];
+
+  $: localProjects = projects[0]?.baseUrl === api.baseUrl;
 </script>
 
 <style>
@@ -61,8 +64,13 @@
 
   {#if projects.length > 0}
     <div class="heading">
-      Explore <span class="txt-bold">projects</span>
-      on the Radicle network.
+      {#if localProjects}
+        Explore <span class="txt-bold">projects</span>
+        on your local node.
+      {:else}
+        Explore <span class="txt-bold">projects</span>
+        on the Radicle network.
+      {/if}
     </div>
 
     <div class="projects">
--- a/src/lib/httpd.ts
+++ b/src/lib/httpd.ts
@@ -147,7 +147,7 @@ function pollSession() {
   pollHttpdStateHandle = window.setInterval(() => checkState(), 10_000);
 }
 
-export function initialize() {
+export async function initialize() {
   // Sync session state changes with other open tabs and windows.
   addEventListener("storage", event => {
     if (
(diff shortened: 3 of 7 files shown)


sebastinez:
> One of the visual tests fails now, the one where we test what happens if a project doesn't load.

Yeah that is because we load all the available projects from the local node so we aren't able to make one fail. Will see how we can keep the test working.

> I guess the "Browse" button now is redundant? Screenshot 2023-08-30 at 12 59 50

I think it still makes sense to get to the local projects from anywhere without going through the landing page.

> And when I start `radicle-httpd` the `radicle.local` button updates, but the home screen doesn't reload automatically.

Yeah the landing page isn't reactive to the httpd state currently, will see if I can improve that

src/views/home/Index.svelte
```diff
@@ -5,8 +5,52 @@
...
+    fetchLocalProjects()
+      .then(localProjects => {
+        switchProjects(localHeading, localProjects);
+      })
+      .catch(() => {
+        switchProjects(networkHeading, projects);
+      });
+  } else {
+    switchProjects(networkHeading, projects);
+  }
```
geigerzaehler, line 28: This code introduces a couple problems:
1. We now have two places where project data is fetched: here and in `home/router.ts`.
2. The is susceptible to a race condition: If `$httpdStore` switches quickly, two `fetchLocalProjects()` tasks may run concurrently and call `switchProjects` out of order.
3. If there is a local node running, the user sees the pinned projects first and then the view updates to the local projects.

An alternative implementation would be to move the fetching of local projects into `loadHomeRoute()` in `home/router.ts`. What we loose then is some reactivity. When you change the port the page is not automatically updated. But I think that’s actually good because there is no loading indicator anyways and you want to avoid a situation where part of the page is updated without the  user knowing about it.

src/App.svelte
```diff
@@ -28,8 +28,11 @@
 
   const activeRouteStore = router.activeRouteStore;
 
-  void router.loadFromLocation();
-  httpd.initialize();
+  // This allows us to get the correct values for the httpd state and port, before navigation.
+  httpd
+    .initialize()
+    .then(() => void router.loadFromLocation())
+    .catch(() => void router.loadFromLocation());
```
geigerzaehler, line 35: You can use
```suggestion
    .finally(() => router.loadFromLocation());
```

sebastinez: hmm yeah that seems nicer, the thing is that eslint doesn't like the missing reject or catch clause
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md

We could probably just ignore eslint on it in this case

src/views/home/Index.svelte
sebastinez: Ok yeah the loading of the projects in the router was my first version, maybe once we have some notification system we can show something that says "We detected a running local node, do you want to reload to see your local projects?"

src/App.svelte
sebastinez: Addressed in f98c22b

src/views/home/Index.svelte
sebastinez: Addressed in 1f9e2a4

src/views/home/Index.svelte
```diff
@@ -59,14 +103,13 @@
     </p>
   </div>
 
-  {#if projects.length > 0}
+  {#if displayedProjects.length > 0}
     <div class="heading">
-      Explore <span class="txt-bold">projects</span>
-      on the Radicle network.
+      {@html heading}
```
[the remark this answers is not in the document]
sebastinez: Addressed in 1f9e2a4

src/App.svelte
```diff
@@ -28,8 +28,9 @@
 
   const activeRouteStore = router.activeRouteStore;
 
-  void router.loadFromLocation();
-  httpd.initialize();
+  // This allows us to get the correct values for the httpd state and port, before navigation.
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  httpd.initialize().finally(() => void router.loadFromLocation());
```
geigerzaehler, line 33: The prevalent convention in the code base seems to be using `void` to disable the no-floating-promises warning:
```suggestion
  void httpd.initialize().finally(() => router.loadFromLocation())
```

geigerzaehler (approved):

PR #1026, opened 2023-09-01
Tabs and icons

dnlklmn:


rudolfs:
I pulled these in https://github.com/radicle-dev/radicle-interface/pull/1018.

PR #1028, opened 2023-09-01
aligned icons on a pixel grid, should be crisp as f*ck

dnlklmn:


rudolfs:
Integrated into https://github.com/radicle-dev/radicle-interface/pull/1018.

PR #1029, opened 2023-09-05
Add embeds

sebastinez:
This PR adds embeds to the issue comments and replies.

- Allows dragging and dropping a file into the comment text area
- Files get encoded into base64 and added to the payload of a comment
- Issues and comments with embeds display their body with the dropped embeds for those `/raw/<rid>/blobs/<oid>?mime=<content-type>` is queried

Requirements:
- [x] [Heartwood patch 87e88e7](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/87e88e7c25dd4c18f2144ab34b85756755482d14)


web-httpd-implementation.md
```diff
@@ -0,0 +1,10 @@
+# Talking points from protocol.berg
```
rudolfs, line 1: Does this belong here?

rudolfs (commented):
Awesome work!
A couple of small nits:

- do we need the `tests/e2e/assets/radicle-228x228.png` file?
- i think it's good to avoid the prefix `get` for functions, e.g. `getBase64String` -> `base64String` and `getEmbed` -> `embed`.

rudolfs:
Issues and patches show the `x` icon even when not editing.

Screenshot 2023-09-27 at 10 26 36
Screenshot 2023-09-27 at 10 26 40

web-httpd-implementation.md
sebastinez: Uuh no it does not 😅

rudolfs:
Probably a regression unrelated to this PR, but patch label's can't be removed anymore:

Screenshot 2023-09-27 at 10 30 36

```
{
  "errorMessage": "Response error",
  "stackTrace": "Error: Response error\n    at Fetcher.fetchOk (http://localhost:3000/httpd-client/lib/fetcher.ts:46:13)\n    at async updatePatch (http://localhost:3000/src/views/projects/Patch.svelte:3034:4)\n    at async LabelInput.saveLabels (http://localhost:3000/src/views/projects/Patch.svelte:3024:19)"
}
```

rudolfs:
It's a bit confusing that I can drop an image into the issue comment text box, but it doesn't work for patches.

sebastinez:
> * do we need the `tests/e2e/assets/radicle-228x228.png` file?

The file is nice to simulate the drop in of a file into the textarea, if you want we can use a smaller image like a small black square instead of the radicle logo, but I wouldn't strip it.

> * i think it's good to avoid the prefix `get` for functions, e.g. `getBase64String` -> `base64String` and `getEmbed` -> `embed`.

Ok cool 👍

sebastinez:
> It's a bit confusing that I can drop an image into the issue comment text box, but it doesn't work for patches.

Yeah patches doesn't work yet, since patches doesn't support embeds yet.
I would suggest to start with issues and when patches supports embeds we can just copy the behavior over to patches on the web client

rudolfs:
> The file is nice to simulate the drop in of a file into the textarea, if you want we can use a smaller image like a small black square instead of the radicle logo, but I wouldn't strip it.

But are we using it anywhere in the code? I thought you used the image from `./public/images/radicle-228x228.png`, no?

sebastinez:
> > The file is nice to simulate the drop in of a file into the textarea, if you want we can use a smaller image like a small black square instead of the radicle logo, but I wouldn't strip it.
> 
> But are we using it anywhere in the code? I thought you used the image from `./public/images/radicle-228x228.png`, no?

Yeah you're right, will remove it 👍

rudolfs:
Could issue comment replies support embeds? Maybe we can add a hint next to the markdown hint about the possibility to add attachments?

rudolfs:
Screenshot 2023-09-27 at 10 37 03

I tried adding a bigger image ~1,6M and got this error. I assume it's due to the size. Would it maybe be possible to warn the user before submitting that the file is too large, or improve the error message?

sebastinez:
> Attachemnts sidebar shows filenames, but is otherwise useless. Could we maybe link them to the files in the raw endpoint? Or make them copy the oid or something? Alternatively we could also remove it.
> 
> Screenshot 2023-09-27 at 10 28 57

I kept them for now, so even if a inlined embed is removed from the comment, one can find it there.
Will see if I can bring the copy functionality back.

web-httpd-implementation.md
sebastinez: ea4f9ec4fdba3b2af5665ba840db7db3ecbb1daa

sebastinez:
> Issues and patches show the `x` icon even when not editing.
> 
> Screenshot 2023-09-27 at 10 26 36 Screenshot 2023-09-27 at 10 26 40



> Awesome work! A couple of small nits:
> 
> * do we need the `tests/e2e/assets/radicle-228x228.png` file?
> * i think it's good to avoid the prefix `get` for functions, e.g. `getBase64String` -> `base64String` and `getEmbed` -> `embed`.

Fixed by ea4f9ec4fdba3b2af5665ba840db7db3ecbb1daa and 43de95467f39cbed98a21afedbecbba0e675f8b2

sebastinez:
> Probably a regression unrelated to this PR, but patch label's can't be removed anymore:
> 
> Screenshot 2023-09-27 at 10 30 36
> ```
> {
>   "errorMessage": "Response error",
>   "stackTrace": "Error: Response error\n    at Fetcher.fetchOk (http://localhost:3000/httpd-client/lib/fetcher.ts:46:13)\n    at async updatePatch (http://localhost:3000/src/views/projects/Patch.svelte:3034:4)\n    at async LabelInput.saveLabels (http://localhost:3000/src/views/projects/Patch.svelte:3024:19)"
> }
> ```

Hmm have you checked that your `radicle-httpd` is on the latest version? Could you check it works when running the test suite with the latest binaries?

sebastinez:
> Screenshot 2023-09-27 at 10 37 03
> ```
> {
>   "errorMessage": "Unexpected token 'F', \"Failed to \"... is not valid JSON",
>   "stackTrace": "SyntaxError: Unexpected token 'F', \"Failed to \"... is not valid JSON"
> }
> ```
> 
> I tried adding a bigger image ~1,6M and got this error. I assume it's due to the size. Would it maybe be possible to warn the user before submitting that the file is too large, or improve the error message?

Thanks for bringing this up. There are two issues here.

On one side `radicle-httpd` with `axum` limits bodys to 2 MB to avoid DDOS attacks.
https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html
Will provide a patch to increase this limit for the routes that we need to send embeds to.

I think we can increase this to 4 MB to match the `/raw` route limit of 4 MB.
Also the body we transmit is usually a bit bigger due to the encoding into base64.

On the other side I'm now showing a modal when the base64 string to be sent is bigger than the 4 MB limit, so we spare the user a not needed request we already know will fail.

rudolfs:
When I copy OIDs from the Attachments section and paste them into a comment, I get this:
Screenshot 2023-09-27 at 15 20 01

I expected to see the screenshot instead.

sebastinez:
> When I copy OIDs from the Attachments section and paste them into a comment, I get this: Screenshot 2023-09-27 at 15 20 01
> 
> I expected to see the screenshot instead.

f6055927d34af8ac7d5c11b7bfc202eb56af81fd fixes this.

sebastinez:
I'm also adding embeds for patch revision comments and reviews, so we can get this PR eventually por issues and patches.

rudolfs (approved):
LGTM, squash and merge! 🚀

sebastinez:
> LGTM, squash and merge! 🚀

Ok is squashed 👍

PR #1034, opened 2023-09-08
icon size and alignment fixes on filter dropdowns

dnlklmn:


public/colors.css
```diff
@@ -11,12 +11,21 @@
   --color-foreground-black: #000000;
   --color-foreground-primary: #ff55ff;
   --color-foreground-success: #5f930b;
+  --color-foreground-primary-fade: linear-gradient(
```
rudolfs, line 14: I'm not using this anywhere, so normally I leave it out from the file.

public/colors.css
```diff
@@ -11,12 +11,21 @@
...
   --color-foreground-red: #af1833;
+  --color-foreground-yellow: #e5c001;
+  --color-foreground-gray: #9494b8;
+  --color-foreground-disabled: #70718f;
   --color-border-hint: #eaeaf1;
   --color-border-focus: #7070ff;
   --color-border-contrast: #24252d;
   --color-border-default: #b2b2cc;
   --color-border-red: #af1833;
+  --boxShadow-lg: 0 0 64px 0 #0000001a;
```
rudolfs, line 28: This lives in `elevations.css`.

src/components/IconSmall.svelte
```diff
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { unreachable } from "@app/lib/utils";
-
+  export let size = 16;
```
rudolfs, line 3: Why do we export this if we're not setting it anywhere from the outside?

rudolfs (commented):
Added some comments to clarify some things and pulled in the commit into my branch. No need to do anything here, I fixed the things on my end.

PR #1043, opened 2023-09-25
Improve binary detection

### LINKED ISSUE #1039: Binary files show up as empty files in diffs
Screenshot 2023-09-19 at 10 55 17

sebastinez:
libgit2 optimizes around not loading the content, when there's no content callbacks configured.

So we aren't able to get a clear answer from radicle-surf if a file in a diff is a binary or not.

A simple way without doing additional API requests is to check if the file type is a executable and if there are no hunks (since we don't get any deltas in binaries), this is a strong
indicator for a binary.

References:
https://github.com/libgit2/libgit2/issues/6637#issuecomment-1733118686
https://github.com/rust-lang/git2-rs/issues/991

Closes #1039

### DIFF
--- a/httpd-client/index.ts
+++ b/httpd-client/index.ts
@@ -13,6 +13,7 @@ import type {
   CommitHeader,
   Diff,
   DiffContent,
+  DiffFile,
   HunkLine,
 } from "./lib/project/commit.js";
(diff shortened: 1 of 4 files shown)


rudolfs (commented):
Could we add a testcase for this?

rudolfs (approved):
🎉

